PR #29, opened 2017-10-09
New package evergreen-checkbox added

jeroenransijn:
There is no issue for this PR.

Implements a Checkbox component. I was hoping to have the Checkbox separate from CheckboxField (which would include label). However that would require JS tot take care of the states instead of doing a CSS `+` selector.

screen shot 2017-10-09 at 11 26 21 am

```jsx
<Checkbox label="Checkbox default" />
<Checkbox checked label="Checkbox checked" />
<Checkbox disabled label="Checkbox disabled" />
<Checkbox disabled checked label="Checkbox checked disabled" />
```


packages/evergreen-checkbox/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "keywords": ["evergreen", "segment", "ui", "react", "Checkbox"],
+  "author": "Segment",
+  "license": "MIT",
+  "dependencies": {
+    "ui-box": "^0.4.0",
+    "evergreen-shared-styles": "^2.0.0",
+    "evergreen-colors": "^2.0.0"
+  },
+  "peerDependencies": {
+    "prop-types": "^15.0.0",
```
Rowno: I don't think prop-types needs to be a peerDependency, it can just be a regular dependency because it's just a library of functions that throw exceptions.

packages/evergreen-checkbox/src/components/Checkbox.js
```diff
@@ -0,0 +1,95 @@
...
+  <svg width={10} height={7} viewBox="0 0 10 7" {...props}>
+    <path
+      fill={fill}
+      fillRule="evenodd"
+      d="M4 4.586L1.707 2.293A1 1 0 1 0 .293 3.707l3 3a.997.997 0 0 0 1.414 0l5-5A1 1 0 1 0 8.293.293L4 4.586z"
+    />
+  </svg>
+)
+
+export default class Checkbox extends PureComponent {
```
Rowno: I would make primitives like this a Component rather than a PureComponent because PureComponent has performance overhead. Higher level components like those in the app should be making the PureComponent optimisation.

packages/evergreen-checkbox/src/components/Checkbox.js
```diff
@@ -0,0 +1,95 @@
...
+    checked: PropTypes.bool,
+    id: PropTypes.string,
+    name: PropTypes.string,
+    onChange: PropTypes.func,
+    value: PropTypes.string,
+  }
+
+  static defaultProps = {
+    appearance: 'default',
+    marginY: 16,
```
Rowno: Do want a default margin? Might be annoying.

packages/evergreen-checkbox/src/components/Checkbox.js
```diff
@@ -0,0 +1,95 @@
...
+    onChange: () => {},
```
Rowno: I don't think you need a default `onChange`.

packages/evergreen-checkbox/src/components/Checkbox.js
```diff
@@ -0,0 +1,95 @@
...
+        display="flex"
+        {...props}
+      >
+        <Box
+          is="input"
+          id={id}
+          type="checkbox"
+          name={name}
+          value={value}
+          checked={checked}
```
Rowno: You should add a `defaultChecked` prop too.

PR #46, opened 2017-11-06
New package table

jeroenransijn:
# Table

`evergreen-table` exports building blocks to build tables with.

## Implementation details

* None of these components implement HTML table elements such as: `table`, `th`, `tr`
* Most components are basic `Pane` components combined with `Text`
* All components are presentational, no sorting build in

## Exported components

* `TableCell`
* `TextTableCell`
* `TableRow`
* `TableHeaderCell`
* `TextTableHeaderCell`
* `SearchTableHeaderCell`
* `TableBody`

## Design example

image


packages/evergreen-table/src/components/TableBody.js
```diff
@@ -0,0 +1,18 @@
+import React, { PureComponent } from 'react'
+import { Pane } from 'evergreen-layers'
+
+export default class TableBody extends PureComponent {
+  static propTypes = {
+    ...Pane.propTypes,
+  }
+
+  static defaultProps = {
+    overflowY: 'scroll',
```
dy: This long-going property forces tables look ugly in windows, when number of elements is less than the table height. Why not making it `auto`?

PR #93, opened 2017-12-07
Add text style between 500 and 600

jeroenransijn:
The sweet spot of headers are around 18 in our case, I added 18, and still maintain 20. I think having 16-18-20 all available makes a lot of sense.

This PR also sets lerna version: independent to true, which @Rowno pointed out to me 👍


jeroenransijn:
NOTE, I moved 600, not added something in between

PR #94, opened 2017-12-07
combobox: add support for custom item objects

Rowno:
Also add shortcut props for `selectedItem`, `defaultSelectedItem` and `itemToString` and fix the `propTypes`.


jeroenransijn:
Will open issue for standardizing `items` nomenclature into `options` for v3

PR #109, opened 2018-01-17
Add file-picker component

Rowno:
Adds a FilePicker component (with tests included!)

image

image

image


nettofarah:
cc @notfelineit, check this out!

packages/evergreen-file-picker/test/index.js
```diff
@@ -0,0 +1,95 @@
...
+  const component = shallow(<FilePicker accept="application/json" />)
+  t.is(
+    component.find(`.${CLASS_PREFIX}-file-input`).prop('accept'),
+    'application/json'
+  )
+})
+
+test('sets required', t => {
+  const component = shallow(<FilePicker required />)
+  t.is(component.find(`.${CLASS_PREFIX}-file-input`).prop('required'), true)
```
vadimdemedes: Use `t.true`/`t.false` instead of `t.is(something, true/false)`.

Rowno: Opps, good point.

PR #126, opened 2018-02-05
Add keyboard navigation to SelectMenu

vadimdemedes:
Up/down arrows do what everyone expects - navigate the list and move the selection. Enter closes the menu if an item has been selected.


jeroenransijn:
Looks good! Maybe down the line we can consolidate the option list in autocomplete and this component — and potentially use Downshift (or not) in that component :)

Issue #127, opened 2018-02-05
Dialog buttons proposal


jeroenransijn:
I am fixing #121 — and wanted to add an API for buttons/actions in the Dialog. This will lead to a more consistent interface for Dialogs.

This is currently my take on the API:

* I am not sure if cancelButton should implicitly be visible when primaryButton is passed
* I am not sure if primaryButton should accept button props.

```javascript
static propTypes = {
    /**
     * Composes the Overlay component as the base.
     */
    ...Overlay.propTypes,

    /**
     * Children can be a node or a function accepting { close }.
     */
    children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),

    /**
     * Props to be passed to the primary button.
     * You should pass `children` and `onClick``
     */
    primaryButton: PropTypes.object,

    /**
     * Label of the cancel button, shown when primaryButton is passed.
     */
    cancelLabel: PropTypes.node,

    /**
     * When
     */
    hideCancelButton: PropTypes.bool,

    /**
     * Title of the Dialog.
     */
    title: PropTypes.node,

    /**
     * Width of the Dialog.
     */
    width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),

    /**
     * Show the close icon.
     */
    hasCloseIcon: PropTypes.bool,

    /**
     * Props that are passed to the Dialog container.
     */
    containerProps: PropTypes.object,
  }

  static defaultProps = {
    width: 567,
    hasCloseIcon: true,
    cancelLabel: 'Cancel',
  }
```

## Stories

```javascript
import { storiesOf } from '@storybook/react' // eslint-disable-line import/no-extraneous-dependencies
import React, { PureComponent } from 'react'
import PropTypes from 'prop-types'
import Box from 'ui-box'
import { Paragraph } from 'evergreen-typography'
import { Button } from 'evergreen-buttons'
import { Dialog } from '../src/'

class DialogManager extends PureComponent {
  static propTypes = {
    children: PropTypes.func,
  }

  state = {
    isShown: false,
  }

  render() {
    return this.props.children({
      isShown: this.state.isShown,
      show: () =>
        this.setState({
          isShown: true,
        }),
      hide: () =>
        this.setState({
          isShown: false,
        }),
    })
  }
}

storiesOf('dialog', module).add('Dialog', () => (
  <Box padding={40}>
    {(() => {
      document.body.style.margin = '0'
      document.body.style.height = '100vh'
    })()}
    <DialogManager>
[...]

PR #134, opened 2018-02-15
Improved dialog and docs + button loading

jeroenransijn:
This PR is based on top of #127, #128. The other PR was still based on the old mono-repo, so I redid most what was in that PR and made some additional changes and tweaks. 

**This is a breaking change**

## This PR contains

* Dialog API improvements
* More Dialog stories
* Dialog documentation
* Button now has a `isLoading` prop
* Minor changes in content for Table and Buttons docs.

## Props and default props

```javascript
static propTypes = {
  /**
   * Children can be a node or a function accepting `({ close })`.
   * See an example to understand how this works.
   */
  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,

  /**
   * When true, the dialog is shown.
   */
  isShown: PropTypes.bool,

  /**
   * Title of the Dialog. Titles should use Title Case.
   */
  title: PropTypes.node,

  /**
   * When true, the header with the title and close icon button is shown.
   */
  hasHeader: PropTypes.bool,

  /**
   * When true, the footer with the cancel and confirm button is shown.
   */
  hasFooter: PropTypes.bool,

  /**
   * When true, the cancel button is shown.
   */
  hasCancel: PropTypes.bool,

  /**
   * Function that will be called when the exit transition is complete.
   */
  onCloseComplete: PropTypes.func,

  /**
   * Function that will be called when the enter transition is complete.
   */
  onOpenComplete: PropTypes.func,

  /**
   * Function that will be called when the confirm button is clicked.
   * This does not close the Dialog. A close function will be passed
   * as a paramater you can use to close the dialog.
   *
   * `onConfirm={(close) => close()}`
   */
  onConfirm: PropTypes.func,

  /**
   * Label of the confirm button.
   */
  confirmLabel: PropTypes.string,

  /**
   * The type of the message.
   */
  type: PropTypes.oneOf(['default', 'danger']),

  /**
   * When true, the confirm button is set to loading.
   */
  isConfirmLoading: PropTypes.bool,

  /**
   * When true, the confirm button is set to disabled.
   */
  isConfirmDisabled: PropTypes.bool,

  /**
   * Function that will be called when the cancel button is clicked.
   * This closes the Dialog by default.
   *
   * `onCancel={(close) => close()}`
   */
  onCancel: PropTypes.func,

  /**
   * Label of the cancel button.
   */
  cancelLabel: PropTypes.string,

  /**
   * Width of the Dialog.
   */
  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),

  /**
   * The space above the dialog.
   * This offset is also used at the bottom when there is not enough space
   * available on screen — and the dialog scrolls internally.
   */
  topOffset: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),

  /**
   * The min height of the body content.
   * Makes it less weird when only showing little content.
   */
  minHeightContent: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),

  /**
   * Props that are passed to the dialog container.
   */
  containerProps: PropTypes.object
}

static defaultProps = {
  isShown: false,
  hasHeader: true,
  hasFooter: true,
  hasCancel: true,
  type: 'default',
  width: 560,
  topOffset: '12vh',
  minHeightContent: 80,
  confirmLabel: 'Confirm',
  isConfirmLoading: false,
  isConfirmDisabled: false,
  cancelLabel: 'Cancel',
  onCancel: close => close(),
  onConfirm: close => close()
}
```

## Design Examples

These examples don‘t reflect the latest state of the component.

dialog
confirm loading
button isloading
dialog docs

### DIFF
--- a/docs/src/components/ComponentReadme.js
+++ b/docs/src/components/ComponentReadme.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-danger */
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import ComponentBlock from './ComponentBlock'
@@ -32,19 +33,17 @@ export default class ComponentReadme extends PureComponent {
       <article className="ComponentReadme" {...props}>
         <div className="Container ComponentReadme-inner">
           <header className="ComponentReadme-header">
-            <h1 className="ComponentReadme-title">{title}</h1>
-            <p className="ComponentReadme-subtitle">{subTitle}</p>
-            <dl>
-              <dt>Links</dt>
-              <dd>
-                <a
-                  href={`https://github.com/segmentio/evergreen/tree/master/src/${name}`}
-                  target="_blank"
-                >
-                  GitHub
-                </a>
-              </dd>
-            </dl>
+            <h1 className="ComponentReadme-title">{title} </h1>
+            <p className="ComponentReadme-subtitle">
+              {subTitle}{' '}
+              <a
+                className="ComponentReadme-githubLink"
+                href={`https://github.com/segmentio/evergreen/tree/master/src/${name}`}
+                target="_blank"
+              >
+                View on GitHub
+              </a>.
+            </p>
           </header>
           <div>
             {designGuidelines && (
@@ -99,8 +98,8 @@ export default class ComponentReadme extends PureComponent {
                 <div className="Content">
                   <h2 id="component-examples">Component Examples</h2>
                   <p>
-                    The <code>{name}</code> package exports the following
-                    documented components:
+                    The following components are exported by Evergreen for this
+                    UI pattern:
                   </p>
                   <ul>
                     {components.map(component => {
--- a/docs/src/components/ComponentReadme.js
+++ b/docs/src/components/ComponentReadme.js
@@ -1,4 +1,3 @@
-/* eslint-disable react/no-danger */
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import ComponentBlock from './ComponentBlock'
--- a/docs/src/css/component-readme.css
+++ b/docs/src/css/component-readme.css
@@ -23,32 +23,15 @@
 
 .ComponentReadme-header {
   border-bottom: 1px solid var(--neutral-20A);
-  /* padding-bottom: 16px; */
+  padding-bottom: 16px;
   margin-bottom: 24px;
+}
 
-  & dl {
-    font-size: 14px;
-    margin-top: 16px;
-    margin-bottom: px;
+.ComponentReadme-githubLink {
+  color: var(--blue-500);
 
-    & dt {
-      color: var(--neutral-200);
-      width: 72px;
-      display: inline-block;
-    }
-
-    & dd {
-      width: calc(100% - 72px);
-      margin-left: 0;
-      margin-bottom: 16px;
-      display: inline-block;
-
-      & a {
-        &:hover {
-          text-decoration: underline;
-        }
-      }
-    }
+  &:hover {
+    text-decoration: underline;
   }
 }
 
--- a/docs/src/css/component-readme.css
+++ b/docs/src/css/component-readme.css
@@ -33,6 +33,10 @@
   &:hover {
     text-decoration: underline;
   }
+
+  &:focus {
+    outline: 1px solid var(--blue-500);
+  }
 }
 
 /**
--- a/docs/src/css/main-layout.css
+++ b/docs/src/css/main-layout.css
@@ -9,7 +9,6 @@
 .MainLayout-main {
   flex: 1;
   display: flex;
-  /* background-color: var(--neutral-3); */
 
   @media (--range-palm), (--range-hand), (--range-lap) {
     flex-direction: column;
--- a/docs/src/css/main-layout.css
+++ b/docs/src/css/main-layout.css
@@ -9,7 +9,7 @@
 .MainLayout-main {
   flex: 1;
   display: flex;
-  background-color: var(--neutral-3);
+  /* background-color: var(--neutral-3); */
 
   @media (--range-palm), (--range-hand), (--range-lap) {
     flex-direction: column;
--- a/src/buttons/docs/LoadingManager.js
+++ b/src/buttons/docs/LoadingManager.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+export default class LoadingManager extends React.PureComponent {
+  static propTypes = {
+    children: PropTypes.func
+  }
+
+  state = {
+    isLoading: false
+  }
+
+  render() {
+    return this.props.children({
+      isLoading: this.state.isLoading,
+      setLoading: () => {
+        this.setState({
+          isLoading: true
+        })
+
+        window.setTimeout(() => {
+          this.setState({
+            isLoading: false
+          })
+        }, 2000)
+      }
+    })
+  }
+}
--- a/src/buttons/docs/LoadingManager.js
+++ b/src/buttons/docs/LoadingManager.js
@@ -10,20 +10,22 @@ export default class LoadingManager extends React.PureComponent {
     isLoading: false
   }
 
+  setLoading = () => {
+    this.setState({
+      isLoading: true
+    })
+
+    window.setTimeout(() => {
+      this.setState({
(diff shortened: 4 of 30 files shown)


docs/src/components/ComponentReadme.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-danger */
```
Rowno: Why add this? I don't see `dangerouslySetInnerHTML` anywhere?
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-danger.md

docs/src/css/component-readme.css
```diff
@@ -23,32 +23,15 @@
...
+  color: var(--blue-500);
 
-      & a {
-        &:hover {
-          text-decoration: underline;
-        }
-      }
-    }
+  &:hover {
+    text-decoration: underline;
```
Rowno: Might want a focus style too? 😉

docs/src/css/main-layout.css
```diff
@@ -9,7 +9,7 @@
 .MainLayout-main {
   flex: 1;
   display: flex;
-  background-color: var(--neutral-3);
+  /* background-color: var(--neutral-3); */
```
Rowno: Probably should just delete it.

src/buttons/docs/LoadingManager.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  state = {
+    isLoading: false
+  }
+
+  render() {
+    return this.props.children({
+      isLoading: this.state.isLoading,
+      setLoading: () => {
```
Rowno: You should really make this a method on the class and keep the JSX clean.

src/buttons/stories/index.stories.js
```diff
@@ -154,6 +154,56 @@ buttonsStory.add('Button + icons', () => (
...
+  }
+
+  state = {
+    isLoading: false
+  }
+
+  render() {
+    return this.props.children({
+      isLoading: this.state.isLoading,
+      setLoading: () => {
```
Rowno: Same here

src/dialog/docs/DialogManager.js
```diff
@@ -0,0 +1,40 @@
...
+  state = {
+    isShown: false,
+    isLoading: false
+  }
+
+  render() {
+    return this.props.children({
+      isShown: this.state.isShown,
+      isLoading: this.state.isLoading,
+      confirmLoading: () => {
```
Rowno: Move all these callback to class methods.

src/dialog/docs/examples/primary-button-confirmation.example
```diff
@@ -0,0 +1,23 @@
...
+  {({ isShown, isLoading, confirmLoading, show, hide }) => (
+    <Box>
+      <Dialog
+        isShown={isShown}
+        title="Dialog with Primary Confirmation"
+        onExited={hide}
+        primaryButton={{
+          children: isLoading ? 'Loading...' : 'Confirm Loading',
+          onClick: confirmLoading,
+          isLoading,
```
Rowno: I'm not sure how useful having `DialogManager` handle loading state is because in most real world cases the loading state will come down from the parent container.

src/dialog/docs/index.js
```diff
@@ -0,0 +1,183 @@
...
+      confirmation from the user before doing a lengthy or dangerous action.
+      This could be a deletion of some sorts or initiating a lengthy download.
+    </p>
+    <h3>Terminology</h3>
+    <p>
+      <a href="http://blueprintjs.com/docs/v2/#core/components/dialog">
+        BlueprintJS
+      </a>{' '}
+      pointed out in their documentation that &ldquo;modal&rdquo; is a misnomer
+      for &ldquo;dialog&rdquo;.
```
Rowno: Use “ and ” instead. Unicode is your friend. 😉

src/dialog/src/Dialog.js
```diff
@@ -50,28 +50,88 @@ const animationStyles = {
...
+    /**
+     * Composes the Overlay component as the base.
+     */
     ...Overlay.propTypes,
+
+    /**
+     * Children can be a node or a function accepting `({ close })`.
+     * See an example to understand how this works.
+     */
+    children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
Rowno: Might be worth making `children` required because a dialog without children is pretty useless. 😛

src/dialog/src/Dialog.js
```diff
@@ -50,28 +50,88 @@ const animationStyles = {
...
+     * Title of the Dialog. Titles should use Title Case.
+     */
     title: PropTypes.node,
+
+    /**
+     * Props passed through to the primary button.
+     * Passing this object will show the button and cancel button.
+     * You should pass `children` and `onClick`.
+     */
+    primaryButton: PropTypes.object,
```
Rowno: So it can't be boolean?

src/dialog/src/Dialog.js
```diff
@@ -50,28 +50,88 @@ const animationStyles = {
...
     width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-    height: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-    hasCloseIcon: PropTypes.bool,
+
+    /**
+     * The space above the Dialog.
+     * This offset is also used at the bottom when there is not enough space
+     * available on screen — and the Dialog scrolls internally.
+     */
+    topOffset: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
Rowno: This can't be a number because I don't think unitless values will work in the CSS calc below.

src/dialog/src/Dialog.js
```diff
@@ -81,40 +141,77 @@ class Dialog extends React.Component {
...
+                      marginLeft={8}
+                      appearance="green"
+                      {...primaryButton}
+                      onClick={() =>
+                        typeof primaryButton.onClick === 'function'
+                          ? primaryButton.onClick(close)
+                          : close()
+                      }
+                    />
+                    {hideCancelButton ? null : (
```
Rowno: Just do `{!hideCancelButton && (`.

docs/src/components/ComponentReadme.js
jeroenransijn: Oh yeah removed that later.

src/dialog/src/Dialog.js
jeroenransijn: Nope.

src/dialog/src/Dialog.js
jeroenransijn: Ah, we need to do add px then.

src/dialog/src/Dialog.js
jeroenransijn: Yes

docs/src/css/main-layout.css
jeroenransijn: Yes

src/buttons/docs/LoadingManager.js
jeroenransijn: Yes.

src/dialog/docs/DialogManager.js
jeroenransijn: Yes.

src/dialog/src/Dialog.js
Rowno: Most of the time you wouldn't really want to pass props to the primary button though?

Then you'll end up with lots of code like `primaryButton={{}}`.

src/dialog/src/Dialog.js
jeroenransijn: True

src/buttons/stories/index.stories.js
jeroenransijn: yes

docs/src/css/component-readme.css
jeroenransijn: I was too lazy.

src/dialog/src/Dialog.js
jeroenransijn: You have to pass the onClick handler

jeroenransijn: And you need to pass in children to give it a label. You always have to define something right?

PR #142, opened 2018-02-22
Add toasts (toaster)

jeroenransijn:
Implements the toaster functionality.

See the docs here: http://evergreen.surge.sh/components/toaster

toaster close all


IdealSystemsMCP:
@jeroenransijn is there any specific reason why you've chosen a toaster singleton instead of using the portal component to handle toast notifications?

PR #143, opened 2018-02-25
Making SearchInput width-aware

fforres:
Hallo @Rowno @jeroenransijn 
I came across a case where I needed to customize the width of the searchInput box. 

Passing a 'width' prop to `SearchInput` only set the container's width and not the input, so hopefully this pr fixes that.

> While It could've been done with CSS classes, I wanted to add the fix to evergreen.
> My other idea was to modify `splitBoxProps` to pass another parameter of whitelisted keys that should not be removed.

### Screenshot of storybook 
image

### DIFF
--- a/src/search-input/src/SearchInput.js
+++ b/src/search-input/src/SearchInput.js
@@ -20,9 +20,8 @@ export default class SearchInput extends PureComponent {
   render() {
     const { appearance, iconProps, disabled, height, ...props } = this.props
     const { matchedProps, remainingProps } = splitBoxProps(props)
-    const { width } = matchedProps
+    const { width = TextInput.defaultProps.width } = matchedProps
     const iconSize = getIconSizeForControlHeight({ height })
-
     return (
       <Box
         position="relative"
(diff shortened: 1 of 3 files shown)


src/search-input/src/SearchInput.js
```diff
@@ -45,6 +45,7 @@ export default class SearchInput extends PureComponent {
           paddingLeft={height}
           appearance={appearance}
           disable={disabled}
+          width={width}
```
Rowno: I think a better solution would be to add `display="block"` and `width={null}` to `TextInput` so it just stretches to fill the parent `Box`.

fforres: I tested it, and it works if the parent Box has an explicit width.
If the parent's `Box` width is not set up, the `TextInput` wont expand to fill up space.

image

jeroenransijn:
Thanks @fforres for taking a look at this and fixing this! This was totally an annoying bug in Evergreen ❤️

PR #145, opened 2018-02-28
Fforres/fix warnings

fforres:
On #143 I added the wrong proptype validations so this PR fixes it

Also, while using comboBox on our app, I'm getting a warning for
image
```
Warning: Failed prop type: The prop `appearance` is marked as required in `IconButton`, but its value is `undefined`.
    in IconButton
```

So this PR defines a defaultProp for `<IconButton/>` to be passed onto the `<Button/>` and allows overwriting it on `<ComboBox/>`


fforres:
:D Np

PR #146, opened 2018-03-01
Fforres/fix spellchheck typo

fforres:
- Small typo en `spellcheck` Prop on textArea
- Fixing height to receive a number on Segmented Control (I'm not sure on this one, should it be both as in `CornerDialog`? considering we can pass `50vh` as well as `22` for pixels ? cc @jeroenransijn  )


Rowno:
I think `height` needs to be a number here because it's used in calculations: https://github.com/segmentio/evergreen/blob/7f0a6bd3485ce529e70de20acff4e49803069846/src/segmented-control/src/SegmentedControlRadio.js#L75-L76

PR #149, opened 2018-03-03
Server Side Rendering (SSR) support (extractStyles)

jeroenransijn:
## Server Side Rendering

Evergreen bundles 2 CSS-in-JS solutions, from glamor and ui-box. To make it super
easy to do server side rendering and hydration, Evergreen exposes it's own function
that will do SSR for both at once.

### Next.js SSR example

Install a new [Next.js](https://github.com/zeit/next.js) app with [create-next-app](https://github.com/segmentio/create-next-app).
Create a `_document.js` inside of the `./pages` folder with the following content:

```javascript
// ./pages/_document.js
import Document, { Head, Main, NextScript } from 'next/document'
import { extractStyles } from 'evergreen-ui'

export default class MyDocument extends Document {
  static getInitialProps({ renderPage }) {
    const page = renderPage()
    // `css` is a string with css from both glamor and ui-box
    // no need to do glamor manually if you are using it elsewhere in your app
    //
    // `evergreenHydrateScript` is a script you should render on the server.
    // it will contain a stringified JSON of the cache of both glamor and ui-box.
    // Evergreen will look for that script on the client and automatically hydrate both glamor and ui-box
    const { css, evergreenHydrateScript } = extractStyles()
    return {
      ...page,
      css,
      evergreenHydrateScript
    }
  }

  constructor(props) {
    super(props)
    const { __NEXT_DATA__, ids, cache } = props
    if (ids) {
      __NEXT_DATA__.ids = this.props.ids
    }
  }

  render() {
    return (
      <html>
        <Head>
          <style dangerouslySetInnerHTML={{ __html: this.props.css }} />
          {this.props.evergreenHydrateScript}
        </Head>
        <body className="custom_class">
          <Main />
          <NextScript />
        </body>
      </html>
    )
  }
}
```


fforres:
🤘 🤘 🤘 🤘 🤘 🤘

README.md
```diff
@@ -131,6 +132,65 @@ The following file tree will be generated:
...
+    if (ids) {
+      __NEXT_DATA__.ids = this.props.ids
+    }
+  }
+
+  render() {
+    return (
+      <html>
+        <Head>
+          <style dangerouslySetInnerHTML={{ __html: this.props.css }} />
```
Rowno: Does it need to be `dangerouslySetInnerHTML`?

src/ssr/src/extractStyles.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const manualHydrateCache = {
+    uiBoxCache: cache,
+    glamorIds: ids
+  }
+
+  const scriptProps = {
+    type: 'application/json',
+    id: EVERGREEN_SCRIPT_ID,
+    dangerouslySetInnerHTML: { __html: JSON.stringify(manualHydrateCache) }
```
Rowno: Why do you need to use `dangerouslySetInnerHTML`?

src/ssr/src/extractStyles.js
```diff
@@ -0,0 +1,70 @@
...
+            'Evergreen automatic hydration object is invalid JSON',
+            err
+          )
+        }
+      }
+    }
+  }
+}
+
+autoHydrate()
```
Rowno: Unfortunately we'll need to remove `"sideEffects": false` from the `package.json` for this to work correctly in Webpack 4, which will result in tree shaking not working very well. Ideally the user should manually call `autoHydrate()` before rendering and we don't call it here.

src/ssr/test/index.js
```diff
@@ -0,0 +1,33 @@
...
+  shallow(<Box height={11} />)
+  const result = extractStyles()
+  t.snapshot(result)
+})
+
+test('extractStyles method returns css and scriptContents for a Button', t => {
+  shallow(<Button />)
+  const result = extractStyles()
+  t.snapshot(result)
+})
```
Rowno: These 2 unit tests do the same thing?

src/ssr/test/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+test('extractStyles should return a scriptProps', t => {
+  shallow(<Box height={16} />)
+  const result = extractStyles()
+  const wrapper = shallow(<script {...result.scriptProps} />)
```
Rowno: `scriptProps` isn't returned from `extractStyles()` so you're just snapshotting an empty script tag: https://github.com/segmentio/evergreen/pull/149/files#diff-d37f316bbb8eb20ed5fac6fb16270dc8R95

PR #153, opened 2018-03-07
Fixing some warnings

fforres:
I was receiving some warning on a couple of components we where using.
So this really small PR adds a default appearance for `ComboBox` & a default border PropType for `Pane`

### DIFF
--- a/src/layers/src/Pane.js
+++ b/src/layers/src/Pane.js
@@ -26,6 +26,7 @@ export default class Pane extends PureComponent {
 
     // Enable to set a boolean for a default border
     border: StringAndBoolPropType,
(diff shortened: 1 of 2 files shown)


src/layers/src/Pane.js
```diff
@@ -25,7 +25,7 @@ export default class Pane extends PureComponent {
     activeElevation: ElevationPropType,
 
     // Enable to set a boolean for a default border
-    borderTop: StringAndBoolPropType,
+    border: StringAndBoolPropType,
```
Rowno: Why was `borderTop` removed?

fforres: Shoot, i meant to add that. My bad, readding

PR #162, opened 2018-03-19
Adding some docs for evergreen's Autocomplete

fforres:
- Adding some docs for evergreen's Autocomplete
- I kept having some props error, so I changed `TextTableCell` prop `textProps` to be `PropTypes.shape(Text.propTypes)` instead of a `PropTypes.shape(PropTypes.object)`


src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,155 @@
...
+    return -1
+  }
+  if (nameA > nameB) {
+    return 1
+  }
+  return 0
+})
+
+const title = 'Autocomplete'
+const subTitle = 'A component to filter trhough a dataset'
```
jeroenransijn: A component to filter through a list of options

src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,155 @@
...
+
+const introduction = (
```
jeroenransijn: introduction => implementationDetails.

introductions should try to use accessible language to both designers and engineers on the long run.

src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,155 @@
...
+  Box,
+  Button,
+  Autocomplete,
+  items,
+  TextInput
+}
+
+const components = [
+  {
+    name: 'AutoComplete',
```
jeroenransijn: Autocomplete

src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,155 @@
...
+  <div>
+    <p>
+      The <code>autocomplete</code> component.
+    </p>
+  </div>
+)
+
+const description = (
+  <p>
+    The <code>AutoComplete</code> component.
```
jeroenransijn: The Autocomplete component.

src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,155 @@
...
+})
+
+const title = 'Autocomplete'
+const subTitle = 'A component to filter trhough a dataset'
+
+const introduction = (
+  <div>
+    <p>
+      The <code>Autocomplete</code> components implements a{' '}
+      <code>VirtualList</code> inside a React Portal, list and to be enable
```
jeroenransijn: The Autocomplete renders a filterable list of options in a Popover — it does not render the text input.  If you need a combination with a text input and trigger button, take a look at the Combobox component. This component is mainly used to create type aheads.

###  The Autocomplete component combines a set of external libraries

- Uses Downshift for autocomplete
- Uses react-tiny-virtual-list for performant list rendering
- Uses fuzzaldrin-plus for fuzzy filtering

src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,155 @@
...
+      filtering and handling big sets of data.
+    </p>
+  </div>
+)
+
+const designGuidelines = (
```
jeroenransijn: We can delete this if it's empty

src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,155 @@
...
+          <p>
+            Autocomplete with button that will trigger the opening of the
+            selectable options
+          </p>
+        ),
+        codeText: exampleWithButtonBasic,
+        scope
+      },
+      {
+        title:
```
jeroenransijn: title: 'Fully Featured Example'
description: 'Full width (using flex) example with options, title, onFocus, filtering, and a button to trigger the autocomplete''

src/autocomplete/src/Autocomplete.js
```diff
@@ -15,18 +15,61 @@ const autocompleteItemRenderer = props => <AutocompleteItem {...props} />
...
+    /**
+     * In case the array of items is not an array of strings, this function is used on each item to return the string that will be shown on the filter
+     */
     itemToString: PropTypes.func.isRequired,
+    /**
+     * Function that will render the 'filter' component.
+     */
     children: PropTypes.func.isRequired,
+    /**
+     * Proptype to set the lits items size beforehand for the virtual list to render properly.
```
jeroenransijn: The height of each item. Because the list is virtualized this is required beforehand.

src/autocomplete/src/Autocomplete.js
```diff
@@ -15,18 +15,61 @@ const autocompleteItemRenderer = props => <AutocompleteItem {...props} />
...
+     */
     itemSize: PropTypes.number,
+    /**
+     * Function that returns a component to render the iist item.
```
jeroenransijn: Function that returns a component to render the item.

src/autocomplete/src/Autocomplete.js
```diff
@@ -15,18 +15,61 @@ const autocompleteItemRenderer = props => <AutocompleteItem {...props} />
...
+     */
     renderItem: PropTypes.func,
+    /**
+     * Function used to filter items it should just return a subset of the initial items.
```
jeroenransijn: A function that is used to filter the items. It should return a subset of the initial items. By default the "fuzzaldrin-plus" package is used.

src/autocomplete/docs/index.js
jeroenransijn: We can ignore the introduction and just use implementationDetails instead.

fforres:
Updated comments @jeroenransijn  :D

src/autocomplete/docs/index.js
```diff
@@ -0,0 +1,166 @@
...
+})
+
+const title = 'Autocomplete'
+const subTitle = 'A component to filter through a list of options'
+
+const implementationDetails = (
+  <div>
+    <p>
+      The <code>Autocomplete</code> components renders a filterable list of
+      options in a Popover — it does not render the text input - If you need a
```
jeroenransijn: Being a type nazi here. hyphen (-) is not an em dash (—). Please use a em dash in this case. Learn more here:  https://practicaltypography.com/hyphens-and-dashes.html

jeroenransijn: Note that because atom uses a monospace font, it doesn't look any different in the code editor.

docs/src/utils/getComponent.js
```diff
@@ -41,8 +41,9 @@ const map = {
   colors: colorsDocs,
   'select menu': selectMenuDocs,
   'text input': textInputDocs,
-  'search input': searchInputDocs,
-  icons: iconsDocs
+  icons: iconsDocs,
+  autocomplete: autocompleteDocs
```
jeroenransijn: Unfortunate merge :(

fforres: Woops :(

jeroenransijn:
@fforres 🥇 👍 🎉 🌲

PR #170, opened 2018-03-26
Focus support dialog

jeroenransijn:
Improves focus management for Overlay + Dialog.

### DIFF
--- a/src/dialog/src/Dialog.js
+++ b/src/dialog/src/Dialog.js
@@ -257,7 +257,12 @@ class Dialog extends React.Component {
                   <Heading is="h4" size={600} flex="1">
                     {title}
                   </Heading>
-                  <IconButton appearance="ghost" icon="close" onClick={close} />
+                  <IconButton
+                    tabindex={3}
+                    appearance="ghost"
+                    icon="close"
+                    onClick={close}
+                  />
                 </Pane>
               )}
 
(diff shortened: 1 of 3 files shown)


src/dialog/src/Dialog.js
```diff
@@ -257,7 +257,12 @@ class Dialog extends React.Component {
                   <Heading is="h4" size={600} flex="1">
                     {title}
                   </Heading>
-                  <IconButton appearance="ghost" icon="close" onClick={close} />
+                  <IconButton
+                    tabindex={3}
```
Rowno: We probably don't want to use a fixed tab order because then the Dialog buttons will always get focus before the Dialog content. Instead we should let the DOM order decide the tab order.

src/overlay/src/Overlay.js
```diff
@@ -181,13 +181,29 @@ class Overlay extends React.Component {
         const autofocusElement = this.containerElement.querySelector(
           '[autofocus]'
         )
-        const wrapperElement = this.containerElement.querySelector('[tabindex]')
-        // eslint-disable-next-line eqeqeq, no-eq-null
+
+        // eslint-disable-next-line eqeqeq, no-eq-null, no-negated-condition
         if (autofocusElement != null) {
           autofocusElement.focus()
```
Rowno: You should move the `autofocusElement.focus()` to the else and vise versa instead of ignoring the problem.

PR #178, opened 2018-03-27
Improved focus handling

jeroenransijn:
This is based on the feedback from @Rowno. I removed the hardcoded tabindex order in the Dialog in favor of using DOM order. Bringing the focus inside the Overlay (or Popover) now looks for:

1. `[autofocus]`
2. `[tabindex]`
3. `button`

### DIFF
--- a/src/overlay/src/Overlay.js
+++ b/src/overlay/src/Overlay.js
@@ -181,29 +181,18 @@ class Overlay extends React.Component {
         const autofocusElement = this.containerElement.querySelector(
           '[autofocus]'
         )
+        const wrapperElement = this.containerElement.querySelector('[tabindex]')
+        const buttonElement = this.containerElement.querySelector('button')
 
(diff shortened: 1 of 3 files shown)


src/overlay/src/Overlay.js
```diff
@@ -181,29 +181,18 @@ class Overlay extends React.Component {
         const autofocusElement = this.containerElement.querySelector(
           '[autofocus]'
         )
+        const wrapperElement = this.containerElement.querySelector('[tabindex]')
+        const buttonElement = this.containerElement.querySelector('button')
 
-        // eslint-disable-next-line eqeqeq, no-eq-null, no-negated-condition
+        // eslint-disable-next-line eqeqeq, no-eq-null
         if (autofocusElement != null) {
```
Rowno: You should use `if (autofocusElement) {` instead and remove all the `eslint-disable-next-line`'s.

PR #187, opened 2018-04-17
Fix stacking with Stack component, StackingContext and React 16.3

jeroenransijn:
This PR makes use of the `createContext` API in React 16.3.0 to allow automatic stacking of components. 

**This is a breaking change and will have to be released as v4.**

## Changes

* Use React 16.3.0
* Use `createContext` API
* Export `StackingContext` from `evergreen-ui`
* Export `Stack` component from `evergreen-ui`
* Export `StackingOrder` object with z-index presets from `evergreen-ui`.
* Remove `zIndex` props from Popover, Combobox and Positioner.

## StackingContext
The StackingContext is a React context with a default value (z-index) of `10`. The StackingContext is currently only used within the `Stack` component within Evergreen.

## Stack component

The Stack component uses the StackingContext which accepts a function as children. That function takes in the zIndex and should return a React node:

```
static propTypes = {
  /**
   * Function that takes the current z-index and returns a React Node.
   * (zIndex) => ReactNode.
   */
  children: PropTypes.func.isRequired,

  /**
   * Set the value of the stack. This will increment for children.
   */
  value: PropTypes.number
}
```

Inside of the render function the Stack component first looks for the current value. Passing a value to the component will make sure the highest of the two are used. This is useful because Overlays start at a `z-index` of 20. See more info below about `StackingOrder`. The Stack component will pass the current value to the current children, and increment the value for the next consumer.

```jsx
render() {
  const { children, value } = this.props
  return (
    <StackingContext.Consumer>
      {previousValue => {
        const currentValue = Math.max(value, previousValue)
        const nextValue = currentValue + 1
        return (
          <StackingContext.Provider value={nextValue}>
            {children(currentValue)}
          </StackingContext.Provider>
        )
      }}
    </StackingContext.Consumer>
  )
}
```

### Stack component usage

In most cases Stack will be an internal component, we are exposing it if you want to build custom components on top of this logic.

```jsx
<Stack value={StackingOrder.POSITIONER}>
  {zIndex => {
    return (
      /* ... code omitted */
    )
  }}
</Stack>
```

## StackingOrder

The values here are somewhat random, the reason why POSITIONER and OVERLAY are 10 apart is that in between the Stack component can increment the z-index — giving a head room of 10 z-indexes.

```js
/**
 * Stacking order contains z-index values that are used through.
 * Note that the Stack component might increase the z-index for certain components.
 */
export default {
  /**
   * Used for focused buttons and controls.
   */
  FOCUSED: 2,

  /**
   * Used as the default for the StackingContext.
   */
  STACKING_CONTEXT: 5,

  /**
   * Used as the default for the Positioner.
   */
  POSITIONER: 10,

  /**
   * Used for the Overlay and everything that's inside such as Dialog + SideSheet.
   */
  OVERLAY: 20,

  /**
   * Used for the toasts in the toaster. Appears on top of everything else.
   */
  TOASTER: 30
}
```

## See Example

I tweeted about this with a video attached. Click the link below to see.

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr">React.createContext is a great solution for solving automatic stacking order when using portals for modal dialogs, popovers, panels and sheets. <a href="https://twitter.com/hashtag/reactjs?src=hash&amp;ref_src=twsrc%5Etfw">#reactjs</a> <a href="https://t.co/9RImVl2ISV">pic.twitter.com/9RImVl2ISV</a></p>&mdash; Jeroen Ransijn 🌲 (@Jeroen_Ransijn) <a href="https://twitter.com/Jeroen_Ransijn/status/986346971580395520?ref_src=twsrc%5Etfw">April 17, 2018</a></blockquote>

## Stuff that came up

When I was doing this PR, I realized there is an issue with the  `getPosition` for the `Positioner` that creates an infinite loop when there is not enough space on the top or the bottom of the screen. I will fix this outside of the scope of this PR so it will be included in `v3` as well.

Another issue that arose from creating examples is when a Popover contains a Combobox — which means another Popover. The Popover closes when trying to select an item from the Combobox. This is because the Popover inside the Popover does not live within the same tree. This should be fixed in a different PR as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "evergreen-ui",
-  "version": "3.0.1",
+  "version": "3.1.0",
   "description": "🌲 React UI Kit by Segment 🌲",
   "contributors": [
     "Jeroen Ransijn (https://jssr.design/)",
--- a/package.json
+++ b/package.json
@@ -51,8 +51,8 @@
     "ui-box": "^1.1.0"
   },
   "peerDependencies": {
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react": "^16.3.0",
+    "react-dom": "^16.3.0"
   },
   "devDependencies": {
     "@reactions/component": "^2.0.2",
@@ -81,9 +81,9 @@
     "husky": "^0.14.3",
     "lint-staged": "^6.0.0",
     "prettier": "^1.9.1",
-    "react": "^16.2.0",
+    "react": "^16.3.0",
     "react-chromatic": "^0.7.10",
-    "react-dom": "^16.2.0",
+    "react-dom": "^16.3.0",
     "react-test-renderer": "^16.2.0",
     "sinon": "^4.4.2",
     "size-limit": "^0.14.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "evergreen-ui",
-  "version": "3.1.0",
+  "version": "3.1.2",
   "description": "🌲 React UI Kit by Segment 🌲",
   "contributors": [
     "Jeroen Ransijn (https://jssr.design/)",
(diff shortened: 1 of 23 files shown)


jeroenransijn:
This is in v4.

PR #193, opened 2018-04-30
Enabling Dialog on Firefox 

fforres:
Currently evergreen Dialog behavior on FF is extending over the boundaries of the container.

I did extract footer from the scrollable component and put it at the same level as header/body of the dialog

Behavior for chrome stays the same :) 

### Current behavior on Firefox
image


fforres:
min-height situation

Rowno:
I'm not sure, but I think the scrollbar should be outside the padding.

Rowno:
No one approved the PR?

PR #194, opened 2018-05-02
Removing minheight from dialog implementation

fforres:
Removing `minHeight` on dialog update to avoid a breaking changes


Rowno:
But won't your other changes break without a `min-height`?

fforres:
Closing. as planning to rewrite Dialog component https://github.com/segmentio/evergreen/pull/193#issuecomment-385566872

Issue #196, opened 2018-05-16
Idea for Multi-Select Approachability via Deselect or Unselect


JMFURY:
Loving the components, and huge fan of Segment in general by the way.. 

I had been trying to figure out how best to leverage `SelectMenu`, and couldn't quite think through how I would manage a Multi-Select (my current use case). 

I installed the library and played around with the Storybook and realized it might possible (& nice) to be able hook into some sort of `onDeSelect()` (or unselect) event handler in order to give more obvious freedom to leverage the existing 'multi' (selected array) support in the component.

I rolled a **very hacky example** which started at top level like this:

![image](https://user-images.githubusercontent.com/7191639/40116061-19a05f6c-58d8-11e8-8691-f515c7f9e79e.png)

Happy to share details, but basically just added `onDeSelect()` as a similar prop/proptype to `onSelect()` further and further in components until I realized the following would probably be necessary to prove it out... 

I hooked in a hack to the  `handleClick()` of `TableRow` based on `isSelected`:

![image](https://user-images.githubusercontent.com/7191639/40117726-5d48d5be-58dd-11e8-9aa6-89c8b0f02d30.png)

Aside from a warning via `div, Box ` ....
![image](https://user-images.githubusercontent.com/7191639/40118246-ff074ae2-58de-11e8-862e-72c46461807b.png) 
(note: maybe I missed a declaration somewhere, or don't yet fully understand composition) 

This did actually work and produce this when clicking a selected item :

![ondeselect-example](https://user-images.githubusercontent.com/7191639/40116026-ff3afcf4-58d7-11e8-9a1c-d233b2c871c2.JPG)

Obviously `console.log` is just to show possibility but I will next start to manage my array of selected values in state/props/Redux, etc. and see what happens.

I could be totally off base, but figured I would drop it in here to discuss with you whether this is an approach you've considered for `SelectMenu` and/or `TableRow`. Again, I may be missing some part how this interplays/impacts other standards or components so definitely not confident this is perfectly helpful, but had to share. 

If there's a better approach I would love to hear about it! Thanks very much!

jeroenransijn:
Hey @JMFURY thanks for showing praise and taking the time to write this issue. We appreciate it a lot!

Select Menu is a continuously evolving component and we get more requirements everyday from our own teams as well. Multi select is not something we ran into just yet, but definitely on our radar.

[...]

Issue #191, opened 2018-04-27
Question: Multiline Button


graysonhicks:
Love this component library.  Quick question:

The Button seems to keep the text on a single line, even when the height is increased.  What's the best way to wrap the text?

Thanks!

jeroenransijn:
Hey @graysonhicks thanks for checking out Evergreen and kicking the tires. Could you give a little bit more details on the result you are trying to achieve — maybe a screenshot or napkin sketch works. There might be some ways to hack around this shortcoming, otherwise we'll have to figure out if we want to support that use case.

Thanks!

graysonhicks:
Sure, no problem. (This is just a coding exercise so it's not a deal-breaker, but I do really like Evergreen and thought I would ask about it).

 Basically I have some buttons with some very long text.  The text overflows the button, and the button stays on one line and does not grow in heigh to accommodate the text.

Here's a screenshot:

<img width="495" alt="screen shot 2018-04-29 at 12 28 11 am" src="https://user-images.githubusercontent.com/16200795/39428815-bfe26e44-4c56-11e8-8d2c-de4e7e062a30.png">

Expected behavior: button grows in height to fit both lines of text.

jeroenransijn:
Thanks for sharing your use case. Seems like something we might want to support. It will be tricky to support this without adding a new API or creating a new component sadly. This is because a lot of the components such as the buttons are controlled by a fixed height.

graysonhicks:
I understand, thanks.  Would also note though, that this would seem pretty common on small screen widths with a lot fewer words.  

I may mess around to see if there is some simple styling that would allow a button to grow with text with a special prop.

jeroenransijn:
This has become an issue for as well! Looping in @vadimdemedes here who is experiencing this issue in our app. I think we probably need to patch this before v4.

![image 5](https://user-images.githubusercontent.com/564463/40745811-5390f372-640d-11e8-8264-46c49bd7677e.png)

Issue #175, opened 2018-03-27
Use @reactions/component for storybook and docs examples


jeroenransijn:
I want move over the examples and stories to use https://reactions.github.io/component/ for state and component management. This will allow examples to contain all state and data in the example, instead of referring to outside scope.

## Why?

- Allows all the context to be in the examples. No referring to outside scope.
- All examples will use one single convention.
- Easier copy-paste/migration between stories and examples.
- The https://reactions.github.io/component/ package is documented.

## How examples will look

Let's take the `SegmentedControl` component as an example. On the docs page you will see the following snippet of code with all of the state and data visible.

```jsx
<Component
  initialState={{
    options: [
      { label: 'Hourly', value: 'hourly' },
      { label: 'Daily', value: 'daily' },
      { label: 'Monthly', value: 'monthly' },
    ],
    value: 'hourly'
  }}
>
  {({ state, setState }) => (
    <SegmentedControl
      width={240}
      options={state.options}
      value={state.value}
      onChange={value => setState({ value })}
    />
  )}
</Component>
```

## What we are doing now in some cases

As a side note, this component is almost the same to using the Manager that is being used in some components. The reason for moving over to https://reactions.github.io/component/ is a single documented way of doing this (instead of reinventing the wheel).

```jsx
import React from 'react'
import PropTypes from 'prop-types'

export default class Manager extends React.Component {
  static propTypes = {
    children: PropTypes.func,
  }

  constructor(props) {
    super(props)
    this.state = {
      ...props,
    }
  }

  render() {
    return this.props.children({
      setState: (...args) => {
        this.setState(...args)
      },
      state: this.state,
    })
  }
}
```

JMFURY:
@jeroenransijn  So for updating Multi Select example, and for future, I would love your thoughts on the following: 

1. Example (abbreviated for the parts that actually change):
![image](https://user-images.githubusercontent.com/7191639/41755052-01843626-759b-11e8-9513-d47b36517897.png)


I did this with the Stories too and it worked. And you're right, more straightforward to copy/paste in! 

2. Some concerns I had

After you pointed it out and I read through, I knew to add `import Component from "@reactions/component";`  to stories, which might not be obvious unless folks are already familiar with this pattern. Maybe this is just a docs/contrib guide thing.

[...]

PR #198, opened 2018-05-18
Add DeSelect func to SelectMenu & affected components

JMFURY:
Follow up from issue #196 - 

Thanks for checking out my idea! Super stoked if I can help in any way here.

### DIFF
--- a/src/select-menu/docs/examples/SelectMenu-multi.example
+++ b/src/select-menu/docs/examples/SelectMenu-multi.example
@@ -5,14 +5,17 @@
                 options={options}
                 selected={state.selected}
                 onSelect={item => {
-                  let selected;
-                  if (!Array.isArray(state.selected)) {
-                    selected = arrify(item.value);
-                  } else {
-                    selected = [...state.selected, item.value];
-                  }
-                  setState({ selected: selected, selectedNames: selected.toString().substr(0,25) + '...' });
-                }}
+            let selected
+            if (Array.isArray(state.selected)) {
+              selected = [...state.selected, item.value]
+            } else {
+              selected = arrify(item.value)
+            }
+            setState({
+              selected,
+              selectedNames: selected.toString().substr(0, 25) + '...'
+            })
+          }}
                 onDeselect={item => {
             let DeselectedItemIndex = state.selected.indexOf(item.value);
             let selectedItems = state.selected.filter((_item, i) => i !== DeselectedItemIndex);
--- a/src/select-menu/docs/examples/SelectMenu-multi.example
+++ b/src/select-menu/docs/examples/SelectMenu-multi.example
@@ -16,26 +16,29 @@
               selectedNames: selected.toString().substr(0, 25) + '...'
             })
           }}
-                onDeselect={item => {
-            let DeselectedItemIndex = state.selected.indexOf(item.value);
-            let selectedItems = state.selected.filter((_item, i) => i !== DeselectedItemIndex);
-          
-            let selectedItemsLength = selectedItems.length;
-            let selectedNames = "";
+          onDeselect={item => {
+            const DeselectedItemIndex = state.selected.indexOf(item.value)
+            const selectedItems = state.selected.filter(
+              (_item, i) => i !== DeselectedItemIndex
+            )
+
+            const selectedItemsLength = selectedItems.length
+            let selectedNames = ''
             switch (true) {
-              case (selectedItemsLength == 0):
-                selectedNames = "";
-                break;
-              case (selectedItemsLength == 1):
-                selectedNames = selectedItems.toString();
-                break;
-              case (selectedItemsLength > 1):
-                selectedNames = selectedItems.toString().substr(0, 25) + '...';
-                break;
+              case selectedItemsLength === 0:
+                selectedNames = ''
+                break
+              case selectedItemsLength === 1:
+                selectedNames = selectedItems.toString()
+                break
+              case selectedItemsLength > 1:
+                selectedNames = selectedItems.toString().substr(0, 25) + '...'
+                break
+              default:
+                break
             }
-            setState({ selected: selectedItems, selectedNames: selectedNames })
-          }
-               }
+            setState({ selected: selectedItems, selectedNames })
+          }}
               >
              <Button>{state.selectedNames || 'Select multiple...'}</Button>
      </SelectMenu>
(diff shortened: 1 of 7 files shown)


jeroenransijn:
This is awesome! Sorry for my late reply — have been a lot of moving parts lately on Evergreen. I don't have a ton of time to prioritize this PR atm, the code looks good. Could you provide a gif of the usage in the storybook and docs site? Let me know if you need any help with that.

JMFURY:
Yes absolutely. Can whip something like that up soon. Will do!

JMFURY:
@jeroenransijn  Here's a GIF of what I have so far.. 

select-multiplie-deselect

Not sure I am quite clear on whether to craft my own 'example' file under the docs tree or not, it looks like you run a command to do this? 

If that's the process should I complete the example manually in a file there (once generated/created) to match what I have here within index.stories.js ?

Thanks in advance for clarity!

JMFURY:
Also sorry the quality sort of stinks, if you have a GIF-making toolset you use, would love to know what it is! I converted an mp4 to GIF..

fforres:
Hijacking the thread for Gif recording info
@JMFURY for mac, [giffox](https://gifox.io/) or [kap](https://getkap.co/) + an afterpass with [giflossy](https://github.com/kornelski/giflossy) if gif ends up being too big :p

jeroenransijn:
@JMFURY docs can be somewhat confusing. We can always add that later, the storybook is the most important thing! Everything looks good to me, the only thing we should change for moving this in is renaming `DeSelect` => `Deselect`. I found that deselect is a single word after some googling.

JMFURY:
Awesome. Will update everywhere and commit.

JMFURY:
@jeroenransijn  Just committed what I have, of course the example is quite a bit of code to be sitting on a component like that, without Redux or a more contextual local state to work with, it's what I could think of as a good representation. 

If you look it over and think I can consolidate for purposes of example or just in general a better way to handle it let me know. 

A lot of the time I spent really just trying to get the names formatted properly so the Button wasn't huge and rendering oddly in some of the same ways referenced here: #191

src/select-menu/docs/examples/SelectMenu-multi.example
```diff
@@ -0,0 +1,46 @@
...
+            } else {
+              selected = arrify(item.value)
+            }
+            setState({
+              selected,
+              selectedNames: selected.toString().substr(0, 25) + '...'
+            })
+          }}
+          onDeselect={item => {
+            const DeselectedItemIndex = state.selected.indexOf(item.value)
```
jeroenransijn: `DeselectedItemIndex` => `deselectedItemIndex`. UpperCamelCase is only for class names (components) and enum like objects.

src/select-menu/docs/examples/SelectMenu-multi.example
```diff
@@ -0,0 +1,46 @@
...
+            const selectedItems = state.selected.filter(
+              (_item, i) => i !== DeselectedItemIndex
+            )
+
+            const selectedItemsLength = selectedItems.length
+            let selectedNames = ''
+            switch (true) {
```
jeroenransijn: Wow, I just learned something new in JS. Never thought you could do this with a switch case. We should probably dumb it down to use if statements instead, I think this might not familiar for a lot of people (including me haha).

src/select-menu/docs/examples/SelectMenu-multi.example
```diff
@@ -0,0 +1,46 @@
...
+              case selectedItemsLength === 0:
+                selectedNames = ''
+                break
+              case selectedItemsLength === 1:
+                selectedNames = selectedItems.toString()
+                break
+              case selectedItemsLength > 1:
+                selectedNames = selectedItems.toString().substr(0, 25) + '...'
```
jeroenransijn: This is cool and a nice solution to this problem! Alternatively, I could see simply `2-n items` as the label.

src/select-menu/docs/examples/SelectMenu-multi.example
```diff
@@ -0,0 +1,46 @@
+<Manager>
```
jeroenransijn: This is not a blocker to merge this in, but long term we hope to migrate Evergreen over to always use `@reactions/component` in examples: https://github.com/segmentio/evergreen/issues/175. Just so you are aware next time 👍

src/select-menu/src/OptionsList.js
```diff
@@ -33,6 +33,7 @@ export default class OptionsList extends PureComponent {
      */
     selected: PropTypes.arrayOf(PropTypes.string),
     onSelect: PropTypes.func,
+    onDeselect: PropTypes.func,
```
jeroenransijn: Awesome, you are my hero! Thanks for renaming 👍

src/select-menu/stories/index.stories.js
```diff
@@ -24,5 +25,51 @@ storiesOf('select-menu', module).add('SelectMenu', () => (
...
+          }}
+          onDeselect={item => {
+            const DeselectedItemIndex = state.selected.indexOf(item.value)
+            const selectedItems = state.selected.filter(
+              (_item, i) => i !== DeselectedItemIndex
+            )
+
+            const selectedItemsLength = selectedItems.length
+            let selectedNames = ''
+            switch (true) {
```
jeroenransijn: See other comment on this statement.

src/select-menu/docs/examples/SelectMenu-multi.example
JMFURY: Ahh makes sense! I started with C++ via Arduino a couple years back actually, so this is partly that shining through. Good deal.

src/select-menu/docs/examples/SelectMenu-multi.example
JMFURY: Exactly the wisdom I was looking for!

src/select-menu/docs/examples/SelectMenu-multi.example
JMFURY: Got it! Definitely want to get this down. Will investigate.

JMFURY:
@jeroenransijn  

All committed now, had a weird version issue with my local and VS Code.. 
Clearly I will need to get this whole process down. 

Thanks a ton for your help, insight, and patience in reviewing the code!

jeroenransijn:
Thanks @JMFURY for putting in the first meaningful PR outside of Segment! Your help is hugely appreciated 👍

jeroenransijn:
` evergreen-ui 3.2.0 published 🎉` Thanks @JMFURY !

JMFURY:
@jeroenransijn So awesome! Thanks again! 

I am all-in on Evergreen. 

Cheers and congrats on the release!

PR #205, opened 2018-06-15
Fix TableRow onDeselect and onSelect

### LINKED ISSUE #202: onDeselect bug
Hey @JMFURY! Spotted this while testing something else. Seems like we are accidentally passing the `onDeselect` on a `Box` somewhere. I need to dig a little deeper, my guess here is that we are spreading props and not taking this prop out.

image

jeroenransijn:
This fixed #202.


JMFURY:
Sorry about that, sir. I should have made more of an effort to squash that. 

If you need me to jump on anything else that might be affected components in the future (as things grow/change) too, let me know!

jeroenransijn:
@JMFURY no worries!

PR #207, opened 2018-06-23
Update to Select-Menu multiselect docs, example, & story

JMFURY:
Per @jeroenransijn guidance on docs/examples usage here #175 & #198 

- Add import & scope ref for `<Component />`  per #175
- Simplify the state management with options & selected array for the example & story
- Still use options from starwars-options that is scoped in

### DIFF
--- a/src/select-menu/docs/examples/SelectMenu-multi.example
+++ b/src/select-menu/docs/examples/SelectMenu-multi.example
@@ -10,7 +10,7 @@
       options={state.options}
       selected={state.selected}
       onSelect={item => {
-        let  selected = [...state.selected, item.value]
+        const selected = [...state.selected, item.value]
         const selectedItems = selected
         const selectedItemsLength = selectedItems.length
         let selectedNames = ''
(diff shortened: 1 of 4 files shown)


jeroenransijn:
This is awesome! Starwars options is fine with me! I think it's important to show the a giant list so people understand it's virtualized.

src/select-menu/docs/examples/SelectMenu-multi.example
```diff
@@ -1,51 +1,49 @@
...
+    selected: [] 
+  }}
+>
+  {({ state, setState }) => (
+    <SelectMenu
+      title="Select multiple names"
+      options={state.options}
+      selected={state.selected}
+      onSelect={item => {
+        let  selected = [...state.selected, item.value]
```
jeroenransijn: Seems this should be `const`?

jimmyredcaffeine: You are correct; the `let` vs. `const` mutability approach was a trip. All that, then I failed on the very copy/paste I commented on. Great catch!

PR #208, opened 2018-06-29
Fixes keyup/keydown/enter in select-menu component

### LINKED ISSUE #201: Select Menu Selection + Filtering bug
select menu filtering selection

jssridhar:
Fixes #201.

Please let me know if this fix makes sense. I am slightly confused as to what the expected behaviour should be in case of multi select.

### DIFF
--- a/src/select-menu/src/OptionsList.js
+++ b/src/select-menu/src/OptionsList.js
@@ -78,6 +78,7 @@ export default class OptionsList extends PureComponent {
     window.setTimeout(() => {
       this.searchRef.querySelector('input').focus()
     }, 1)
+    window.addEventListener('keyup', this.handleKeyUp)
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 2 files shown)


jeroenransijn:
Hey @jssridhar thanks so much for taking the time and put in this PR 👍  — much appreciated!
Select menu, dropdowns, they are super confusing to get right in terms of functionality and accessibility. It will take some time for us to get it perfectly right.

It looks good from just eyeballing the code. Could you provide a gif of the behavior? Here is a tool to do that if you don't have one: https://getkap.co/

A gif would be appreciated, if you show that, I would say let's merge it in!

jssridhar:
True that! I am still looking at ways to get this fixed in the multi-select component, will submit another pr for it if I can get it working.
functionality gif
selectmenu

jeroenransijn:
@jssridhar thanks for sharing the gif! Long term, we might want to consider using the arrow keys to focus an item, but use enter to select/deselect an item. I am working on an implementation in v4 that uses this approach and it works pretty well.

> True that! I am still looking at ways to get this fixed in the multi-select component, will submit another pr for it if I can get it working.

Is this PR ready to merge?

jssridhar:
Yep, this is ready for merge!

jeroenransijn:
🎉  Published in `v3.2.4`: 
 https://circleci.com/gh/segmentio/evergreen/538

PR #219, opened 2018-07-11
Fix popover not closed when toggle button has children

### LINKED ISSUE #218: [V4] Popover not toggleable when clicking on icon in the icon button
👋 **Use case**

IconButton used as popover toggle.

😢 **Issue** 
Popover doesn't close popover content when clicking on the **icon** area(in the screenshot, click on the triangle icon). **Note**  if clicking on button itself it does close the popover.

```
<Popover content={<PopoverContentWithTextInput />}>
  <IconButton icon={IconNames.SYMBOL_TRIANGLE_DOWN} />
</Popover>
```

image

I didn't see any stories about this use case.

jfmaggie:
Closes #218 

**In this PR**
Fix the popover not being closed when clicking on button that has children (icon...)

kapture 2018-07-10 at 17 32 05


src/popover/src/Popover.js
```diff
@@ -168,7 +168,7 @@ export default class Popover extends Component {
 
   onBodyClick = e => {
     // Ignore clicks on the popover or button
-    if (this.targetRef === e.target) {
+    if (this.targetRef === e.target || this.targetRef.contains(e.target)) {
```
mshwery: `contains` will check the node itself as well, so you can make this:

```js
if (this.targetRef.containers(e.target)) {
```

For instance, this will be true: 
```js
const node = document.getElementById('foo')
node.contains(node)
```

That said, it doesn't read as explicit as you have it here, so I'm not opposed to this.

mshwery: Oh, you might also encounter a race condition where `targetRef` is actually undefined still, so it might be safe to check that it has been defined first:

```js
if (this.targetRef && this.targetRef.contains(e.target)) {
```

package.json
```diff
@@ -127,7 +127,7 @@
   "size-limit": [
     {
       "path": "commonjs/index.js",
-      "limit": "70 KB"
+      "limit": "180 KB"
```
mshwery: I'm not sure what this does, @jeroenransijn is this ok?

Rowno: It checks the size of the compiled evergreen package using webpack. Apparently v4 increases the package size by 100kb 😱 

https://github.com/ai/size-limit

mshwery: @jeroenransijn looks like blueprint icons contribute significantly to the build size.

PR #232, opened 2018-07-18
increase the contrast of n1-level colors and fix typo

mshwery:
The current N1* color is a bit too light (not enough contrast) even on very high contrast/bright screens (currently on a MacBook Pro at full brightness, stock color balance).

screen shot 2018-07-18 at 11 13 58 am

This also fixes a small typo `isselecteable` -> `isselectable`


src/theme/src/default-theme/foundational-styles/scales.js
```diff
@@ -39,7 +39,7 @@ export default {
     B10: '#084b8a', // Normal Text AAA
 
     // Transparent variants.
-    B1A: 'rgba(16, 112, 202, 0.03)',
+    B1A: 'rgba(16, 112, 202, 0.04)',
```
jeroenransijn: I think we should update `B1` as well?

mshwery: Probably so.

mshwery: @jeroenransijn done

Issue #229, opened 2018-07-17
[v4] `SegmentedControl` missing focus state


mshwery:
This component doesn't have the same focus state as other form controls. In the below screenshot you can't tell the difference between the `SegmentedControl`'s focus state and unfocused state:

<img width="688" alt="screen shot 2018-07-16 at 11 27 57 pm" src="https://user-images.githubusercontent.com/710752/42796474-f8fae8ec-894f-11e8-8093-4485d3f8faa2.png">

jeroenransijn:
Thanks for documenting, have been thinking how to support this properly. Will put this in v4 shortly.

PR #233, opened 2018-07-18
add focus feedback to segmented controls

bmcmahen:
I needed focus feedback in segmented controls for accessibility purposes... so here's my quick solution. Just noticed that this should address #229.

### DIFF
--- a/src/segmented-control/src/SegmentedControl.js
+++ b/src/segmented-control/src/SegmentedControl.js
@@ -94,7 +94,8 @@ export default class SegmentedControl extends PureComponent {
         {options.map((option, index) => (
           <SegmentedControlRadio
             key={option.value}
-            name={name}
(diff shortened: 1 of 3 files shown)


jeroenransijn:
Awesome @bmcmahen! Could you provide a gif or screenshot of the result?

bmcmahen:
here ya go

segmented-controls-focus

I think this is the proper behaviour for accessibility. When working through a form using tabs it focuses on the selected radio button initially. Left - right keys shift both focus and radio selection. And the selected state remains when focus shift to somewhere else on the page. The actual shadow used already existed in your code.

jeroenransijn:
I will have to move this in v4 as well!

jeroenransijn:
`evergreen-ui 3.2.5 published`

PR #235, opened 2018-07-19
[V4] Use Tooltip inside of a Popover

jeroenransijn:
This PR implements the support of putting a `Tooltip` inside of a `Popover`.

## Preview
popover tooltip styled

## Code Example

The above example is rendered by the following code.

```jsx
<Popover content={<PopoverContentWithTextInput />}>
  <Tooltip content="Click me">
    <Button marginRight={20}>Tooltip Card + Popover</Button>
  </Tooltip>
</Popover>
<Popover content={<PopoverContentWithTextInput />}>
  <Tooltip
    appearance="card"
    content={
      <React.Fragment>
        <Heading>Heading</Heading>
        <Paragraph color="muted" marginTop={4}>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
          eiusmod tempor incididunt ut labore et dolore magna aliqua.
        </Paragraph>
      </React.Fragment>
    }
    statelessProps={{
      paddingY: 24,
      paddingX: 24,
      maxWidth: 280
    }}
  >
    <Button>Tooltip + Popover</Button>
  </Tooltip>
</Popover>
```

## Other additions

This PR also adds the support for 2 appearances on a `Tooltip`: `default` and `card`.

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { Position, Positioner } from '../../positioner'
+import { Tooltip } from '../../tooltip'
 import PopoverStateless from './PopoverStateless'
 
 export default class Popover extends Component {
@@ -237,12 +238,16 @@ export default class Popover extends Component {
 
   renderTarget = ({ getRef, isShown }) => {
     const { children } = this.props
+    const isTooltipInside = children.type === Tooltip
 
     const getTargetRef = ref => {
       this.targetRef = ref
       getRef(ref)
     }
 
+    /**
+     * When a function is passed, you can control the Popover manually.
+     */
     if (typeof children === 'function') {
       return children({
         toggle: this.toggle,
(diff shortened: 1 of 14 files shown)


src/popover/src/Popover.js
```diff
@@ -237,12 +238,16 @@ export default class Popover extends Component {
 
   renderTarget = ({ getRef, isShown }) => {
     const { children } = this.props
+    const isTooltipInside = children.type === Tooltip
```
mshwery: Is it possible to _not_ pass `children`? If so we may want to check that it exists first.

src/theme/src/default-theme/component-specific/getTooltipProps.js
```diff
@@ -1,11 +1,21 @@
...
 import palette from '../foundational-styles/palette'
 
-const getTooltipProps = () => {
-  return {
-    backgroundColor: tinycolor(palette.neutral.base)
-      .setAlpha(0.95)
-      .toString()
+const getTooltipProps = appearance => {
+  switch (appearance) {
+    case 'card':
```
mshwery: This might be a misunderstanding of the semantics, but I've often followed the rule of thumb that a `Tooltip` is simple, short content. It cannot contain html or other components, only strings. With those semantics, things that look like a raised "card" would qualify as a `Popover` IMO.

Not sure if this meshes with Evergreen's component semantics, but thought it was worth mentioning here (if it allows us to simplify things).

src/tooltip/src/Tooltip.js
```diff
@@ -36,10 +41,17 @@ export default class Tooltip extends PureComponent {
...
      * Properties passed through to the Tooltip.
      */
-    statelessProps: PropTypes.object
+    statelessProps: PropTypes.object,
+
+    /**
+     * This is an implementation detail. Please ignore.
+     * This is passed when a Tooltip is inside a Popover.
+     */
+    popoverProps: PropTypes.object
```
mshwery: I would omit it from the propTypes if this is meant for internal use only.

src/typography/src/ListItem.js
```diff
@@ -1,12 +1,70 @@
 import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
```
mshwery: Were these changes to `ListItem` intentional?

jeroenransijn: I am not sure why this is showing up, since this is already in v4.

src/theme/src/default-theme/component-specific/getTooltipProps.js
jeroenransijn: The distinction currently between `Tooltip` and `Popover` is the interaction kind: hover vs. click. We might want to consider supporting something like `<Popover interaction={InteractionKind.HOVER} />` similar to BlueprintJS at one point. I think we should leave that for a future major release.

src/tooltip/src/Tooltip.js
jeroenransijn: It will throw a warning if it's not in propTypes right?

PR #237, opened 2018-07-20
Bug/select icon margin

mshwery:
This fixes a style issue with `<Select />` when placed inside flex containers or when allowed to have an unconstrained width (`width: auto`) by including `paddingRight={sizeOfIconPlusPaddingX}`.

It also adds `/docs` and a `SelectField` component that matches the implementation for `TextInputField`.

We may have opportunities for refactoring later, but for now making this available is really great.

screen shot 2018-07-20 at 11 31 04 am

### DIFF
--- a/src/select/src/Select.js
+++ b/src/select/src/Select.js
@@ -105,13 +105,14 @@ class Select extends PureComponent {
     const textSize = theme.getTextSizeForControlHeight(height)
     const borderRadius = theme.getBorderRadiusForControlHeight(height)
     const iconSize = theme.getIconSizeForSelect(height)
+    const iconMargin = height >= 36 ? 12 : 8
 
     return (
       <Box
         display="inline-flex"
         flex={1}
         position="relative"
-        width={200}
+        width="auto"
         height={height}
         {...props}
       >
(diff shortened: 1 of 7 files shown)


src/text-input/src/TextInputField.js
```diff
@@ -96,7 +93,7 @@ export default class TextInputField extends PureComponent {
       ...props
     } = this.props
 
-    const { id } = this.state
+    const id = `TextInputField-${this.state.id}`
```
mshwery: This is to avoid collisions with other components using a naive `id` incrementer (cant have the same `id` in a document more than once).

src/select/src/Select.js
```diff
@@ -105,13 +105,14 @@ class Select extends PureComponent {
...
     const iconSize = theme.getIconSizeForSelect(height)
+    const iconMargin = height >= 36 ? 12 : 8
 
     return (
       <Box
         display="inline-flex"
         flex={1}
         position="relative"
-        width={200}
+        width="auto"
```
mshwery: This is up for debate, but if you are using a native select, this is closer to the normal behavior. I would suggest if we want a particular size we may implement that via `minWidth` instead. But I like the `width="auto"` better (because it explicitly says we want auto-sized width based on the select's option contents.

jeroenransijn:
All looks great to me! One thing I have been debating is wether `TextInputField` and `SelectField` should implement `required` vs. `isRequired` to better match native behavior. Not a blocker for getting this in, but thought I mention it here.

mshwery:
@jeroenransijn I would suggest sticking to `required`. 

I would also suggest providing an escape hatch for consumers who want to mark the field as required via html attributes, but **dont** want the `*` asterisk next to the label.

`showRequiredIndicator: PropTypes.bool` (defaults to true?)

Issue #249, opened 2018-07-23
[v4] Remove icon from sidebar tab


jfmaggie:
Ref https://github.com/segmentio/evergreen/issues/247

This issue is to remove the icon from sidebar tab. Or we could make it optional? I think the use case for us in segment app would be not using the icon for selected side bar. 

cc @jeroenransijn

PR #250, opened 2018-07-23
Remove sidebar tab icon

jfmaggie:
Closes https://github.com/segmentio/evergreen/issues/249

Remove the caret-right icon from side bar tab.

image


jfmaggie:
I think it is a good idea to make it configurable. 👍

jeroenransijn:
Let's move in render prop as soon as we need it! Merging this in, thanks @jfmaggie 💯 !

Issue #242, opened 2018-07-20
[v4] Allow "deselecting" in comboboxes


Flaque:
At the moment, a combobox item can be deselected by highlighting the text and deleting it. However, it would be nice if clicking on an item that you've already selected would "deselect" it as a checkmark would normally imply. 

<img width="271" alt="screen shot 2018-07-20 at 2 58 15 pm" src="https://user-images.githubusercontent.com/5942769/43027071-5b06b91e-8c2d-11e8-9226-b8eada91d435.png">

JMFURY:
I took a quick look and I think it should be straightforward since `Combobox` has `Autocomplete` options that derive from `SelectMenu` > `Option` - From there the `onDeselect` method we added for multiselect is available.  

I could take a stab at this.

cc: @jeroenransijn

JMFURY:
@Flaque It turns out I was wrong in a lot of the above assumption (ooops..). This definitely has its differences over Select-Menu. But that's actually good since it is a bit more straightforward. 

@jeroenransijn I have the interaction running after some trial and error. Here are some details:

Within `Autocomplete`'s `renderItem()` we can target the selected item in the same way the 'isSelected' boolean is applied. See comment in the code below:

![image](https://user-images.githubusercontent.com/7191639/43057252-bfbc5618-8e06-11e8-8e62-22203644da87.png)

Changing `onMouseUp` to `onClick` solved an issue where the item would get selected again after clicking it. The deselect and clearSelection (see below) all worked, but the item would get selected again -- all way too fast to see, but seen in the console. 

You may know why the `onMouseUp` was used in the first place so it's probably worth talking it through.

The other key to this was to use the [`Downshift`](https://github.com/paypal/downshift) component's built in clearSelection method / prop. Cool component..

I have some cleaning up to do from trying things before I'd be confident sharing code but I can prep a GIF of the functionality sometime mid-week. 

Seems like you have a ton going on for v4. (stoked) Happy to help whenever this is useful. Cheers!

mshwery:
We may want to expose this as an opt-in: `isClearable` or something.

Flaque:
@JMFURY Feel free to open an "work-in-progress" PR if you'd like!

JMFURY:
All, 

Doing it at the Autocomplete level felt to me like fighting Downshift. Doing the logic within onMouseUp() came with a lot of caveats, fixing one bug caused another and so on..

I am sure with some review/help it could be done more (and cleanly)..

[...]

PR #256, opened 2018-07-24
add isClearable iconbutton & story for review

JMFURY:
Per conversation in #242 - WIP PR for evaluating deselection options in Combobox

Will refer to this PR in the issue thread


Flaque:
Adding reviewers for visibility

jeroenransijn:
I haven't looked too much into this, but it might be worth looking into upgrading to Downshift 2. Not sure about benefits/downsides. If we are able to get it in `v4`, we don't have to do another major release just to update Downshift in Evergreen.

JMFURY:
Awesome - I will look into all these ideas/considerations - version as well. I noticed some selected item selection bugs in Downshift's own issues threads, these were pretty recent conversations too. I don't assume it's all apples to apples given the difference in implementation though. 

Hopefully I can get something workable together for `v4` that'd be sweet!

JMFURY:
**Here's what this looks like now with last commit:** 

isclearable with swap out on iconbutton

Flaque:
The combination `X` and `V` buttons are super clever.

JMFURY:
The Downshift example(s) with this swap really made this happen. Such a great component.

jeroenransijn:
The button sure is clever! However, I am curious if it makes sense to follow how `react-select` implements this thing https://react-select.com/home

reat-select

JMFURY:
Yeah definitely! That's basically what was in the first commit (https://github.com/segmentio/evergreen/pull/256/commits/6a4d6672b186e333a4f4211560b757063a816513) except it only showed [ X ] when the menu `isShown` -- totally doable to have that opt-in with the isClearable and be present same as [ V ]

That GIF looks real slick with the bar | in between though! I'm a fan!

JMFURY:
Oh I was at multi-select which has the [ X ] always.. 

Good stuff though. Let me know what you all think and I'll update to the preference.

mshwery:
I had used `react-select` on a previous project and discovered that many users did not find the X + V intuitive, especially when presented side-by-side. So, YMMV.

JMFURY:
Been thinking about it.. and I think it's almost like we would need a low-fi version of Downshift that doesn't emphasize a selection menu or autocomplete.. I feel like a lot of the principles apply, and wrapping these feels like the right approach.. no?

jeroenransijn:
@JMFURY I am not entirely sure if I follow what you mean, could you provide some more context or an example. What functionality would be in a low-fi Downshift? What I was describing is to add a `isClearable` and `onClear` to the `TextInput` component, we wouldn't have to change all that much within the current components such as Autocomplete or SelectMenu.  I might be missing something obvious, it's been a while since I touched these parts of the code.

JMFURY:
@jeroenransijn Ahh! Got it. Yes. I think you're spot on. I thought there might have been a need to have a wrapper like Downshift simply with `isClearable` and `clearSelection`  -- and I think `clearSelection`  == `onClear` to your point 

I don't have a great handle on Downshift but I also wondered about `getInputProps` 

That was my thought process, but I think I was over-complicating it. 

I will give this a shot on `TextInput`

JMFURY:
@jeroenransijn & team

Just checking in on this. Have had only a little time here and there, but I have something working and would love some feedback! 

I'm not certain what would be the best path forward to implement this across the many types of `<TextInput/>` usage if the default component (even with `isClearable` applied) is **_uncontrolled_** in this context.  

**Big question:** Were you thinking this would still be **_uncontrolled_**? 

The reason I ask is that getting it done in a **controlled** context is pretty simple. 

### What I Have Now
A **controlled** `isClearable`  `<TextInput/>` Implementation (on the `<Autocomplete />` **controlled** use case ... to prove it out)

```jsx
      <Autocomplete
        title="Starwars names with isClearable"
        onChange={handleChange}
        items={items}
      >
        {({ getInputProps, getRef, inputValue, clearSelection }) => (
          <TextInput
            placeholder="Starwars names and isClearable"
            value={inputValue}
            innerRef={ref => getRef(ref)}
            {...getInputProps()}
            onClear={clearSelection}
            isClearable
          />
        )}
      </Autocomplete>

```
Of course the Downshift `clearSelection` makes clearing this nice and clean, but you can imagine it not being tough to just simply clear a solo `<TextInput />` _input value_ in some state.

**Huge GIF of this**
autocompleteisclearableonclear

Going controlled, to me, also gives a clear and distinct hook for when to show/hide the [X] that I don't think is there otherwise.. i.e. I'd like to make the condition based on both `isClearable` and having a value populated -- I think this would be the way to scale across the differences in usage.

I have hardly ever used **uncontrolled** inputs since learning React in the first place, so I'm probably not jiving with it if there is a solution. I am personally not a huge fan of the whole 'ref' thing. That said, I've tried quite a few ideas on this but keep hitting roadblocks. 

Maybe there's something I'm not thinking of? Any thoughts?

mshwery:
@JMFURY first off: awesome work on this!

Regarding your question:
>Big question: Were you thinking this would still be uncontrolled?

I think we need to account for both scenarios because `TextInput` in its various forms can be both controlled or uncontrolled. 

In this situation there are a couple ways to handle this. 
1. Make the component _internally_ controlled. That is to say, consumers would use it as an uncontrolled component, but internally we would use state, change handlers, and generally have complete control of the input value at any time. The downside (or design no-no) with this option is that switching a component from controlled to uncontrolled (or vice versa) will be extra gymnastics TextInput will need to handle. For both types of components you can bubble an event for `onClear`.
2. Use refs and DOM apis to clear the input like you would in the old days. :) (e.g. `myRef.value = ''`)

I think either way we'll need to detect when consumers have provided the `onChange` and `value` props so we don't clear the value for them. 

Will clearing the input also bubble to `onChange` (with an empty string)?

JMFURY:
@mshwery I owe you a coffee! This was driving me nuts, so thanks for the clarity.

Your #**1.**  definitely covered an implied question on whether we were intending to bring a state into the `TextInput.js` so that is great to know it's something you're cool with.

> I think either way we'll need to detect when consumers have provided the onChange and value props so we don't clear the value for them.

Perfect, I'm thinking this is doable in props and covers the bases. 

> Will clearing the input also bubble to onChange (with an empty string)?

I think I experienced this exact concern in making a **controlled** demo too.

Here's that:
```jsx
// Story Example
            <Component
              initialState={{inputValue: ""}}
            >
              {({  setState, state}) => (
                <TextInput
                  appearance={appearance}
                  height={28}
                  name={28}
                  id={28}
                  value={state.inputValue}
                  placeholder="You can clear this one"
                  onClear={() => setState({ inputValue: "" })}
                  onChange={e => setState({ inputValue: e.target.value })}
                  isClearable
                />
              )}
            </Component>

// Important snippet from within TextInput.js
      {(isClearable && props.value)  && (
          <Icon
          color="muted"
          icon="cross"
          appearance="default"
          onClick={onClear}
          marginLeft={-20}
          marginTop={6}
        />
      )}
    
```

`onChange` did not work properly until I used `props.value` (in the second snippet). I had been explicitly including and evaluating on 'value' as a prop from the `const {..} = this.props` statement - that was wrong.

**Controlled works with code from above**
textinputclearwillclear

jeroenransijn:
@mshwery thanks for explaining the controlled/uncontrolled use cases! Onboard with Matt on how to tackle the API.

JMFURY:
Good deal. Working thru it early this week.

JMFURY:
Quick update: I think I've gotten nearly all the controlled/uncontrolled `TextInput` concerns out of the way, examples of each work side by side in stories.

Still need to make sure the variety of needs with `TextInputField`, `Autocomplete`, `Combobox`, etc. are factored into the API. 

Noticing some bugs / unexpected behavior, likely to do with how `onChange` needs to be handled.

JMFURY:
Just committed. I am already seeing now that it could be refactored a bit, maybe make the icon a renderer function for cleanliness; that type of thing..  

See comment on commit as well, just wanting to make things as obvious as possible. 

I think you will all have better ideas on these fronts. 

Cheers!

JMFURY:
Not sure exactly what's going on in FilePicker, but I saw no obvious console errors if that helps.. 
image

Also removed the console.log message I used for testing but still not matching the 'snapshot' in tests 

image

If you have clues on this let me know!

JMFURY:
@jeroenransijn Sorry for the radio silence, I just figure you all were busy on other elements of V4

I definitely need help/clarity! For starters, if you haven't seen yet, here is the failure in the tests:

image


I'm pretty sure I know why this is happening -- see my previous comment as well.

**So with that, I am not sure if...**

**A)** The new shape of `TextInput` is breaking a necessary design pattern that has to stay in place i.e. failed tests for `FilePicker`. 

If this is the case then a review of code committed here would help me out big time! 

**OR** 

**B)** The FilePicker test file simply needs to be updated to accommodate the new `TextInput` shape (This assumes you are all good with wrapping `<Text is="input">` in a `<Box>` to accommodate `<Icon>` for the [X])

--------
I'm sure the new implementation of `TextInput` could be refactored to pass somehow and could probably be improved in general. Open to ideas. 

That said, the `isClearable` and `onClear` functionality with [X] rendering does work in all the necessary controlled and uncontrolled contexts. So pending thoughts on the above it is ready for review.

src/text-input/src/TextInput.js
```diff
@@ -88,25 +116,62 @@ class TextInput extends PureComponent {
     const borderRadius = theme.getBorderRadiusForControlHeight(height)
 
     return (
-      <Text
-        is="input"
-        className={cx(themedClassName, className)}
-        type="text"
-        size={textSize}
-        width={width}
+      <Box
```
JMFURY: @jeroenransijn Here's what I am referring to as changing the shape, which I think causes `<div>` that breaks with the `FilePicker` snapshot test

src/text-input/src/TextInput.js
```diff
@@ -88,25 +116,62 @@ class TextInput extends PureComponent {
...
+          placeholder={placeholder}
+          paddingLeft={Math.round(height / 3.2)}
+          paddingRight={Math.round(height / 3.2)}
+          borderRadius={borderRadius}
+          spellCheck={spellCheck}
+          aria-invalid={isInvalid}
+          {...(disabled ? { color: 'muted' } : {})}
+          css={css}
+          {...props}
+          value={this.handleValueCase()}
```
JMFURY: @jeroenransijn  This is kind of a style choice I would love feedback on here with `handleValueCase()` .. I think a handler like this might make sense, but I did not do the same with `onChange`

src/text-input/src/TextInput.js
```diff
@@ -57,6 +69,13 @@ class TextInput extends PureComponent {
      */
     className: PropTypes.string
   }
+  constructor(props) {
```
jeroenransijn:
```
state = {
  value: ''
}
```

Should this take in account `defaultValue`?

src/text-input/src/TextInput.js
jeroenransijn: I will need to spend some more time playing around with this, in this case it's semantically not a handler. As in, it doesn't respond to a event (`onChange`, `onKeyDown` etc.). In this case we could rename it to `this.getValue()` or something.

src/text-input/src/TextInput.js
jeroenransijn: Yeah, this change is definitely tricky to make because we now have to wrap this component inside of component, while trying to mimic the API as if it was one single component.

I suggest taking a look at `SearchInput` as well, since we are doing almost the same there.

src/text-input/src/TextInput.js
JMFURY: Got this part working

mshwery:
@JMFURY looks like this got a little stale since the release of v4. I would recommend rebasing on latest master to get a much cleaner diff here (and resolve the conflicts).

Happy to help collab to push this through.

JMFURY:
@mshwery Music to my ears! 

I gave it a shot a week or two ago, but no change on the FilePicker test front. 

I think the logic on the Input component is pretty sound but you might have some better insight.
 
I will see what I can do this week post-Holiday! Will reach out with concerns.

mshwery:
@JMFURY sgtm! You could probably end up using `git checkout` even to cherry pick the files you want to keep (or copypasta, whatever). Good luck!

PR #263, opened 2018-07-26
🌲 CI status image

fforres:
We where pointing to `ui-box` CI status image, now , we point to 🌲's CI status image.

image


jeroenransijn:
Haha, I am so sloppy — at least I can trust @fforres to do things the right way!

PR #265, opened 2018-07-26
Fix hiding <Tooltip> by explicitly setting `isShown` to `false`

vadimdemedes:
Fixes https://github.com/segmentio/evergreen/issues/251.

Previously `<Tooltip>` was still visible on hover, even if `isShown` was explicitly set to `false`.

### DIFF
--- a/src/tooltip/src/Tooltip.js
+++ b/src/tooltip/src/Tooltip.js
@@ -156,8 +156,13 @@ export default class Tooltip extends PureComponent {
     } = this.props
     const { isShown: stateIsShown, isShownByTarget } = this.state
 
-    const shown =
+    let shown =
       (isShown || stateIsShown || isShownByTarget) && !this.isPopoverShown()
+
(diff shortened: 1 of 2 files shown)


src/tooltip/src/Tooltip.js
```diff
@@ -156,9 +156,14 @@ export default class Tooltip extends PureComponent {
     } = this.props
     const { isShown: stateIsShown, isShownByTarget } = this.state
 
-    const shown =
+    let shown =
       (isShown || stateIsShown || isShownByTarget) && !this.isPopoverShown()
```
mshwery: Wouldn't we also want to respect `this.props.isShown` here, regardless of `!this.isPopoverShown()`?

vadimdemedes: I don't think I understand, @jeroenransijn definitely will though! Is this an error I can reproduce?

Issue #211, opened 2018-07-06
TableRow that has `isSelectable` prop is not propagating onChange event


emadabdulrahim:
I'm working on implementing editable table cells using Evergreen components.

My issue comes when I have a `TextInput` component that implements `onChange` event inside a `TableRow` that has `isSelectable` prop.

What happens is `onChange` event does NOT get triggered **when hitting the space key on the keyboard.**

If I remove `isSelectable` from `TableRow` things work as expected`

I suspect this line in `TableRow` is causing the problem
https://github.com/segmentio/evergreen/blob/de5ebddc8a07bfc910cde63c9fd0389025d5a12e/src/table/src/TableRow.js#L61

```jsx
handleChange = e => {
    const { target } = e

    console.log("onChange", target.value)

    this.setState(currentState => ({
      [target.id]: {
        ...currentState[target.id],
        editValue: target.value,
      },
    }))
  }


render() {
  return (
    <TableRow isSelectable>
      <TextTableCell borderRight={borderRight}>
        {isEditing ? (
          <TextInput
            height={28}
            width="100%"
            value={children}
            // onChange doesn't fire when I hit the space key
            onChange={this.handleChange}
            {...rest}
          />
        ) : (
          children
        )}
      </TextTableCell>
    </TableRow>
  )
}
```

Thanks!

jeroenransijn:
This is using Evergreen v4.

PR #276, opened 2018-08-04
allow default TableRow keypress event behavior (#211)

JustAboutJeff:
allow default TableRow keypress event behavior (#211)

Thanks for creating evergreen! 🌲

_tl;dr remove the `e.preventDefault()` function call so that `input` and `change` events can be fired normally after the `keypress` event ._

* [Web standards specify that a `keypress` event should be fired **prior** to the corresponding `input` event for the same key](https://www.w3.org/TR/DOM-Level-3-Events/#keypress-event-order)

* [Web standards also specify that cancelling a `keypress` event will also prevent subsequent `input` events from being fired](https://www.w3.org/TR/DOM-Level-3-Events/#ref-for-input-31)

* [Web standards further specify that a `input` event should be fired **prior** to any `change` events for a given input](https://html.spec.whatwg.org/multipage/input.html#common-input-element-events)
  
* From all this, we can conclude that `keypress` event handlers will be invoked prior to any `change` handlers attached in our DOM tree. These two examples confirm this for both [native DOM](https://codesandbox.io/s/r7qllw2r5p?expanddevtools=1&hidenavigation=1) and [React SyntheticEvent](https://codesandbox.io/s/l56198rlzq?expanddevtools=1&hidenavigation=1). In addition, cancelling the default behavior for a `keypress` event will prevent `input` and `change` events from firing for the same user interaction, the cause of our bug here!

### DIFF
--- a/src/table/src/TableRow.js
+++ b/src/table/src/TableRow.js
@@ -54,11 +54,21 @@ export default class TableRow extends PureComponent {
     }
   }
 
+  onRef = node => {
+    this.node = node
+  }
+
   handleKeyPress = e => {
     if (this.props.isSelectable) {
       if (e.key === 'Enter' || e.key === ' ') {
         this.props.onSelect()
-        e.preventDefault()
+
+        // NOTE: In scenarios where the target for key press is our
+        // tabIndex <Pane> below, prevent the default event action
+        // so that browser scroll will not occur.
+        if (e.target === this.node) {
+          e.preventDefault()
+        }
       }
     }
 
@@ -83,6 +93,7 @@ export default class TableRow extends PureComponent {
 
     return (
       <Pane
+        innerRef={this.onRef}
         display="flex"
         {...(isSelectable
           ? {
--- a/src/table/stories/index.stories.js
+++ b/src/table/stories/index.stories.js
@@ -89,6 +89,17 @@ storiesOf('table', module)
       <TableRow>TableRow</TableRow>
     </Box>
   ))
+  .add('TableRow isSelectable with nested UI', () => (
+    <Box padding={40}>
+      {(() => {
+        document.body.style.margin = '0'
+        document.body.style.height = '100vh'
+      })()}
+      <TableRow isSelectable>
+        <input type="text" placeholder="type a space" />
+      </TableRow>
+    </Box>
+  ))
   .add('TableHeaderCell', () => (
     <Box padding={40}>
       {(() => {


src/table/src/TableRow.js
```diff
@@ -58,7 +58,6 @@ export default class TableRow extends PureComponent {
     if (this.props.isSelectable) {
       if (e.key === 'Enter' || e.key === ' ') {
         this.props.onSelect()
-        e.preventDefault()
```
mshwery: Nice! Should we find another way to allow consumers to _opt-in_ to preventing scroll of the page upon these key presses?

JustAboutJeff: @mshwery Thanks for the feedback and additional context on the original implementation here. Appreciate it!

As an alternative to additional configuration for opt-in behavior, it occurred to me that we could instead continue to invoke `preventDefault` but first wrap the function call within a conditional that would first check if the event target is in fact the TableRow itself.

This way, if consumers of 🌲 decide to nest additional UI within an `isSelectable` `TableRow` that can take focus, or accept user input, the browser can and will scroll and direct focus to these child UI elements when the event originates there. This is good, and helps with web accessibility standards.

For an `isSelectable` `TableRow` that contains text or other children that do not provide a target to fire a keyPress event, `preventDefault` will not be called and browser scrolling will not occur when selecting a row. This scenario is captured in the existing [storybook example here](https://github.com/segmentio/evergreen/blob/master/src/table/stories/index.stories.js#L36-L42)

src/table/src/TableRow.js
```diff
@@ -82,22 +92,24 @@ export default class TableRow extends PureComponent {
...
-              tabIndex: 0
-            }
-          : { css })}
-        onClick={this.handleClick}
-        onKeyPress={this.handleKeyPress}
-        {...props}
-      >
-        {children}
-      </Pane>
+      <div ref={this.handleRef}>
```
JustAboutJeff: I'm using a wrapping `div` here so that we can attach a ref to the underlying node and avoid a call to `findDOMNode` which is discouraged for React 16 onward.

src/table/src/TableRow.js
```diff
@@ -54,11 +54,21 @@ export default class TableRow extends PureComponent {
...
   handleKeyPress = e => {
     if (this.props.isSelectable) {
       if (e.key === 'Enter' || e.key === ' ') {
         this.props.onSelect()
-        e.preventDefault()
+
+        // NOTE: In scenarios where the target for key press is our
+        // tabIndex <Pane> below, prevent the default event action
+        // so that browser scroll will not occur.
+        if (e.target.parentNode === this.node) {
```
JustAboutJeff: Because of the wrapping `div` below, we compare the event target's parent node to our ref node here. If they match, the event was fired while keyboard focus was on the `tabIndex` `<Pane>` below and we should stop browser scroll by preventing the default event action.

If the event target was some other node, then the user has interacted with other types of nested UI, and we should *not* prevent the default action so that `input` and `change` events can fire normally

src/table/src/TableRow.js
jeroenransijn: Not sure why we would need to do to this? If you need the ref of the Pane you can use `innerRef={ref => this.paneRef = ref}`. Eslint will probably complain, so you might have to create a func to do it `innerRef={this.onRef}`

JustAboutJeff: Thanks @jeroenransijn. I took this approach because using a wrapping `div` was seemingly the easiest way preserve encapsulation. However, as you pointed out, `Pane` (via `ui-box`) has already been wired up to expose a `ref` prop that we can use. I wasn't aware of this, and this approach works just fine as well.

For reference:
https://github.com/segmentio/ui-box/blob/master/src/box.js#L50-L54

I'll amend my commit to use this approach, and include a new storybook example that will make it easier to verify the bug has been fixed.

jeroenransijn:
Thanks for the work @JustAboutJeff ! Highly appreciated! Will take a look soon to move this in v4 as well. 🌲 💯 🎉

PR #299, opened 2018-08-22
add Positioner support for Position.LEFT and Position.RIGHT

bmcmahen:
I've got basic left / right positions working for the positioner. The one thing I haven't done is messed with pushing the popover incrementally to ensure that it stays on the screen. I think it could make sense for this to be done on the y-axis for `Position.LEFT` and `Position.RIGHT`... leaving the current x-axis logic in place for the other positions. 

The `getPosition` function is getting kinda huge. Not sure if it's worth breaking that up into smaller functions, but I think it works okay-ish as is. 

I also threw in some quick tests. 

positioner-positions


bmcmahen:
Took a stab at this - basically just copied the other functions, but now applied to the y-axis properties.

src/positioner/src/getPosition.js
```diff
@@ -328,16 +341,24 @@ function getPosition({
...
     bottomRect,
     viewport,
     viewportOffset
   )
 
-  if (positionIsAlignedOnTop && topRectFitsOnTop) {
-    return {
-      position,
-      rect: topRect
+  if (positionIsAlignedOnTop) {
```
bmcmahen: This behaviour makes more sense to me - that it would alter the position to fit below if the popover fits within the viewport... but perhaps I'm missing something.

jeroenransijn: I am not exactly sure how this is different in behavior from what's currently there. An example screenshot would help!

bmcmahen: @jeroenransijn Here are some gifs that demonstrate the difference. With the old logic, the popover will be moved down even if the entire popover could instead be positioned below the button. With the new logic, the popover will check if it can appear entirely below the button. Only if it can't, will it reposition it within its preferred position. 

Old:
evergreen-position-old

New:
evergreen-position-new

jeroenransijn: Looks better!

jeroenransijn:
@bmcmahen is this one good to go? Could you update the branch — I might get this in this or next week!

bmcmahen:
@jeroenransijn i've rebased off of the current v4 branch. Everything should be good to go.

jeroenransijn:
`evergreen-ui@4.0.0-39`
https://github.com/segmentio/evergreen/releases/tag/v4.0.0-39

@bmcmahen 🎉

PR #301, opened 2018-08-23
Make the Dialog mobile friendly

Rowno:
This change makes the Dialog resize gracefully to fit the available viewport.

It should be a non-breaking change and the Dialog should behave the same on desktop as it did before.

kapture 2018-08-23 at 0 21 39


mshwery:
@Rowno I think once we get to a reasonably small viewport width (or height) we should reduce the margin/padding between the edge of the screen and the dialog.

src/dialog/src/Dialog.js
```diff
@@ -209,94 +217,91 @@ class Dialog extends React.Component {
...
-    if (Number.isInteger(topOffset)) {
-      maxHeight = `calc(100vh - ${topOffset}px)`
-    } else {
-      maxHeight = `calc(100vh - ${topOffset})`
-    }
+    const topOffsetWithUnit = Number.isInteger(topOffset)
+      ? `${topOffset}px`
+      : topOffset
+    const sideOffsetWithUnit = Number.isInteger(sideOffset)
+      ? `${sideOffset}px`
```
jeroenransijn: I don't think we need a unit here

src/dialog/src/Dialog.js
```diff
@@ -209,94 +217,91 @@ class Dialog extends React.Component {
...
     } = this.props
 
-    let maxHeight
-    if (Number.isInteger(topOffset)) {
-      maxHeight = `calc(100vh - ${topOffset}px)`
-    } else {
-      maxHeight = `calc(100vh - ${topOffset})`
-    }
+    const topOffsetWithUnit = Number.isInteger(topOffset)
+      ? `${topOffset}px`
```
jeroenransijn: I don't think we need a unit here

Rowno: We do need this one for the `maxHeight` `calc()` below.

jeroenransijn: `${topOffset}px` => `topOffset` right?

Rowno: No, because then `calc(100% - ${topOffsetWithUnit} * 2)` won't work.

PR #310, opened 2018-08-28
fix githubLink pathname on ComponentReadme

emersonlaurentino:
The component links to be seen in github are broken. I used the kebabCase function in the component headings and this fixes this problem.

### DIFF
--- a/docs/src/components/ComponentReadme.js
+++ b/docs/src/components/ComponentReadme.js
@@ -1,5 +1,6 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
+import kebabCase from '../utils/kebabCase'
 import ComponentBlock from './ComponentBlock'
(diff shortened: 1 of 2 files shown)


wangcch:
@emersonlaurentino  
Not just two words. You can test.
like:
'a b c'.toLowerCase().replace(/\s+/g, '-')   // a-b-c
'a       b     c'.toLowerCase().replace(/\s+/g, '-')   // a-b-c

But I really didn't think about that  'Remove hyphens from both ends'
Thanks~

emersonlaurentino:
Hey, @wangcch. 

The problem is when you have a title component like `buttonGroup`, the current path would be `buttongroup`. But the correct would be `button-group`.

wangcch:
Sure, you are right.
I'm going to keep trying.

Thanks~ 🙈 @emersonlaurentino

PR #313, opened 2018-08-30
Add "indeterminate" prop to <Checkbox>

### LINKED ISSUE #312: [v4] Checkbox Indeterminate
Add a `indeterminate` prop to the Checkbox.

```jsx
<Checkbox indeterminate />
```

## Design Example

image

nucleartide:
Closes #312.

screen shot 2018-08-30 at 4 52 00 pm

### DIFF
--- a/src/checkbox/src/Checkbox.js
+++ b/src/checkbox/src/Checkbox.js
@@ -105,11 +105,12 @@ class Checkbox extends PureComponent {
     appearance: 'default',
     onChange: () => {},
     indeterminate: false,
-    value: ''
+    checked: false
   }
 
   setIndeterminate = el => {
     if (!el) return
+    console.log(el, this.props.indeterminate)
     el.indeterminate = this.props.indeterminate
   }
 
@@ -151,7 +152,7 @@ class Checkbox extends PureComponent {
           onChange={onChange}
           disabled={disabled}
           {...(isInvalid ? { 'aria-invalid': true } : {})}
-          ref={this.setIndeterminate}
+          innerRef={this.setIndeterminate}
         />
         <Box
           boxSizing="border-box"
--- a/src/checkbox/src/Checkbox.js
+++ b/src/checkbox/src/Checkbox.js
@@ -103,7 +103,14 @@ class Checkbox extends PureComponent {
 
   static defaultProps = {
     appearance: 'default',
-    onChange: () => {}
+    onChange: () => {},
+    indeterminate: false,
+    value: ''
+  }
+
+  setIndeterminate = el => {
+    if (!el) return
+    el.indeterminate = this.props.indeterminate
   }
 
   render() {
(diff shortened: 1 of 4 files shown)


src/checkbox/stories/index.stories.js
```diff
@@ -5,9 +5,42 @@ import { Checkbox } from '../../checkbox'
 
 storiesOf('checkbox', module).add('Checkbox', () => (
   <Box padding={40}>
-    <Checkbox label="Checkbox default" />
-    <Checkbox checked label="Checkbox checked" />
-    <Checkbox disabled label="Checkbox disabled" />
-    <Checkbox disabled checked label="Checkbox checked disabled" />
+    {permutations({
```
vadimdemedes: What's the benefit of having this? I would just add:

```jsx
<Checkbox indeterminate label="Checkbox indeterminate" />
```

src/checkbox/stories/index.stories.js
```diff
@@ -9,5 +9,6 @@ storiesOf('checkbox', module).add('Checkbox', () => (
     <Checkbox checked label="Checkbox checked" />
     <Checkbox disabled label="Checkbox disabled" />
     <Checkbox disabled checked label="Checkbox checked disabled" />
+    <Checkbox checked indeterminate label="Checkbox checked disabled" />
```
vadimdemedes: Label should be `checked indeterminate`

nucleartide: Fixed!

nucleartide:
@jeroenransijn Updated!

src/checkbox/src/Checkbox.js
```diff
@@ -119,10 +147,11 @@ class Checkbox extends PureComponent {
           type="checkbox"
           name={name}
           value={value}
-          checked={checked}
+          checked={checked || indeterminate}
           onChange={onChange}
           disabled={disabled}
           {...(isInvalid ? { 'aria-invalid': true } : {})}
+          ref={this.setIndeterminate}
```
jeroenransijn: `ref` is actually not available on `Box` since it's a React component — instead `Box` uses the non-standard but conventional `innerRef` prop. 

Also, this might actually not work completely since I think `ref/innerRef` is only called once in the lifecycle of the component. We might need to opt-in to `componentDidUpdate`: https://reactjs.org/docs/react-component.html#componentdidupdate

nucleartide:
@jeroenransijn Addressed your comment about `ref` + `componentDidUpdate`. With `innerRef`, it seems to work when you pass in new props.

Try this component in the story:

```jsx
import React from 'react'

class CheckboxWithState extends React.Component {
  state = {
    checked: false,
  }

  handleChange = () => {
    this.setState(() => ({ checked: !this.state.checked }))
  }

  render() {
    return <Checkbox checked={this.state.checked} onChange={this.handleChange} />
  }
}
```

PR #315, opened 2018-08-31
fix prop warnings and make list components more flexible

mshwery:
This fixes those pesky prop-type warnings by providing some sensible default values. It also only passes down props when defined (and removes unnecessary props from the wrapper Box via destructuring).


jeroenransijn:
Thanks @mshwery!

PR #317, opened 2018-08-31
add iconSize to IconButton component. closes #316

### LINKED ISSUE #316: [v4] iconSize prop on IconButton
In some cases the default icon size is too small, or some icons appear better slightly bigger. I want to add a `iconSize` prop to the `IconButton` that would override the value returned from the theme object.

mshwery:
This allows consumers to specify the size of an icon when the theme-based size isn't meeting their needs.

Closes https://github.com/segmentio/evergreen/issues/316

### DIFF
--- a/src/buttons/src/IconButton.js
+++ b/src/buttons/src/IconButton.js
@@ -97,7 +97,7 @@ class IconButton extends PureComponent {
       intent,
       ...props
     } = this.props
-    const size = iconSize || this.props.theme.getIconSizeForIconButton(height)
+    const size = iconSize || theme.getIconSizeForIconButton(height)
 
     return (
       <Button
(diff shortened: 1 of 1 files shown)


src/buttons/src/IconButton.js
```diff
@@ -83,8 +88,16 @@ class IconButton extends PureComponent {
...
+    const {
+      theme,
+      iconAim,
+      icon,
+      iconSize,
+      height,
+      intent,
+      ...props
+    } = this.props
+    const size = iconSize || this.props.theme.getIconSizeForIconButton(height)
```
jeroenransijn: `this.props.theme.getIconSizeForIconButton(height)` => `theme.getIconSizeForIconButton(height)` since you already deconstruct it!

Issue #320, opened 2018-09-03
Spelling mistake in Tooltip proptypes


PMCorbett:
Hi there is a spelling mistake here: https://github.com/segmentio/evergreen/blob/de5ebddc8a07bfc910cde63c9fd0389025d5a12e/src/tooltip/src/Tooltip.js#L34

That is causing the following error to be logged to the console in a browser:

```
Warning: Failed prop type: Tooltip: prop type `children` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
```

PR #321, opened 2018-09-03
Fix typo in Tooltip proptypes

PMCorbett:
Fixes issue: https://github.com/segmentio/evergreen/issues/320

Spelling mistake here: https://github.com/segmentio/evergreen/blob/de5ebddc8a07bfc910cde63c9fd0389025d5a12e/src/tooltip/src/Tooltip.js#L34

That is causing the following error to be logged to the console in a browser:

```
Warning: Failed prop type: Tooltip: prop type `children` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
```

### DIFF
--- a/src/tooltip/src/Tooltip.js
+++ b/src/tooltip/src/Tooltip.js
@@ -31,7 +31,7 @@ export default class Tooltip extends PureComponent {
     /**
      * The target button of the Tooltip.
      */
-    children: PropTypes.node.isRequried,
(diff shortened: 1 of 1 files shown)


jeroenransijn:
Thanks @PMCorbett 👍 

I would also recommend checking out `evergreen-ui@next`  (which is v4) since it is a more stable branch at the moment.

PR #323, opened 2018-09-05
Allow grammarly to be disabled for Textarea

gurdasnijor:
[Grammarly](https://www.grammarly.com/) mutates the backing textarea for the `<Textarea>` component in unexpected ways that leads to issues like these:

screen shot 2018-09-05 at 1 08 50 am

This [PR makes use of an attribute provided to selectively disable grammarly](https://github.com/facebook/draft-js/issues/616#issuecomment-343596615) as a configuration prop

### DIFF
--- a/src/textarea/src/Textarea.js
+++ b/src/textarea/src/Textarea.js
@@ -32,6 +32,11 @@ class Textarea extends PureComponent {
     spellCheck: PropTypes.bool,
 
     /**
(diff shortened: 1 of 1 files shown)


gurdasnijor:
@mshwery No catchall attributes I know of that all extensions would respect -- so far Grammarly has been (by far) the biggest cause of support tickets so I figured we could address this in isolation for now

PR #339, opened 2018-10-01
Pass event to Table onSelect handlers

colinking:
If you have a clickable element inside of a selectable table row, the table row's `onSelect` handler will fire before the handlers on the inner element. Without passing the raw event to the `onSelect` table handler, you can't determine whether the user clicked on the inner element or not.

Here is a simple example ([Source](https://github.com/segmentio/evergreen/blob/68bd08de16511c90244428000035738c3afba85f/src/table/stories/index.stories.js#L18)):
image


src/table/src/TableRow.js
```diff
@@ -92,7 +92,7 @@ class TableRow extends PureComponent {
     if (this.props.isSelectable) {
       const { key } = e
       if (key === 'Enter' || key === ' ') {
-        this.props.onSelect()
+        this.props.onSelect(e)
```
mshwery: To bubble the event you may want to use [`event.persist()`](https://reactjs.org/docs/events.html) first:

>If you want to access the event properties in an asynchronous way, you should call `event.persist()` on the event, which will remove the synthetic event from the pool and allow references to the event to be retained by user code.

colinking: That's a good point, but is this async code? I thought these callbacks were synchronous, CMIIW. I noticed that we're already passing the event on to other handlers in this function without `event.persist()`.

colinking:
@mshwery: it looks like what I would have to do is:

```
<Checkbox
  checked
  onClick={e => {
    // perform click logic
    e.stopPropagation()
    e.preventDefault()
  }}
/>
```

To prevent the click handler from being fired multiple times per-click.

This is good enough for my use case, and the event `e` is already passed to this handler, so I'm no longer blocked by this PR.

However, I'm still quite confused about what is causing this event to fire twice. AFAICT, [the SVG doesn't receive a handler](https://github.com/segmentio/evergreen/blob/v4/src/checkbox/src/Checkbox.js#L166) -- what would cause it to bubble the event twice?

jeroenransijn:
@mshwery @colinking just chiming in to this conversation, maybe `pointer-events: none;` can help on the SVG in the checkbox 🤞

jeroenransijn:
Closing this for now. If functionality is still valid, might be worth doing another PR based on the new v4 master!

colinking:
Sounds good, will keep in mind. We just ended up shipping the workaround above, which is working for now. Thanks!

PR #344, opened 2018-10-15
 Upgrade most of the dependencies

Rowno:
This upgrades all the dependencies in the v4 branch except for downshift (which would cause breaking changes for evergreen). None of these changes should be breaking changes.

I didn't upgrade any of the docs dependencies.


Rowno:
@babel/runtime reduces the overall evergreen bundle size by 10kb and will have even bigger savings for people using @babel/runtime in their apps (like we are). [@blueprintjs/icons v2 to v3 only added new icons](https://github.com/palantir/blueprint/wiki/3.0-Changelog#blueprintjsicons-300).

PR #347, opened 2018-10-16
Update colors.mdx docs to explain programmatic usage

Flaque:
# Overview 
Updates docs to better explain how to get colors from a theme.

### DIFF
--- a/docs/src/pages/components/colors.mdx
+++ b/docs/src/pages/components/colors.mdx
@@ -18,4 +18,12 @@ The color system in Evergreen is located in the theme and is used throughout the
 There is no real dependency on any of the colors directly within components.
 Components always access a theme color or property through a get function. For example, `theme.getTextColor` is a required function in the Evergreen theme, theme.colors is not a required property.
 
+## Accessing Colors
+
+```js
+import { defaultTheme } from 'evergreen-v4'
+
+console.log(palette.blue.lightest) // #F7F9FD
+```
+
 <ColorExamples marginTop={40} />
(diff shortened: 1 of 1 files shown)


docs/src/pages/components/colors.mdx
```diff
@@ -18,4 +18,14 @@ The color system in Evergreen is located in the theme and is used throughout the
...
 Components always access a theme color or property through a get function. For example, `theme.getTextColor` is a required function in the Evergreen theme, theme.colors is not a required property.
 
-<ColorExamples marginTop={40} />
\ No newline at end of file
+## Accessing Colors
+
+You can programatically access colors with a theme's `palette` property.
+
+```js
+import { defaultTheme } from 'evergreen-v4'
```
mshwery: This isn't really recommended since it's referencing a static theme. If you customize the theme none of those customized values will be found in this import.

jeroenransijn: Agree with @mshwery, should probably use `withTheme` instead. Seems like a great addition though!

mshwery: @Flaque typically this is what we should be recommending:

```jsx
import React from 'react'
import { Pane, withTheme } from 'evergreen-ui'

class FooComponent extends React.Component {
  render() {
    // `theme` is passed into props when using `withTheme`
    const { theme } = this.props
    return (
      <Pane backgroundColor={theme.palette.blue.lightest}>
        This is light blue
      </Pane>
    )
  }
}

export default withTheme(FooComponent)
```

PR #356, opened 2018-10-17
Upgrade Downshift

Rowno:
This upgrades Downshift from v1 to v3 and fixes a couple of bugs.

### Bug fixes
- Pressing the up arrow key in the text input now correctly takes you to the bottom of the list.
- The Popover's Portal is now properly supported which improves accessibility and fixes touch screen support.
- Anything using Positioner will now render correctly if `isShown` is `true` on initial render.

### Breaking changes
Because all the Downshift props are exposed directly, these breaking changes are introduced to `Autocomplete` and `Combobox`:
- https://github.com/paypal/downshift/releases/tag/v2.0.0
- https://github.com/paypal/downshift/releases/tag/v3.0.0

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -35,7 +35,7 @@ export default class Autocomplete extends PureComponent {
     /**
      * The selected item to be selected & shown by default on the autocomplete
      */
-    defaultSelectedItem: PropTypes.any,
+    initialSelectedItem: PropTypes.any,
 
     /**
      * In case the array of items is not an array of strings,
@@ -115,9 +115,9 @@ export default class Autocomplete extends PureComponent {
     width,
     inputValue,
     highlightedIndex,
-    selectItemAtIndex,
     selectedItem,
-    getItemProps
+    getItemProps,
+    getMenuProps
   }) => {
     const {
       title,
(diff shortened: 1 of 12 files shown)


src/autocomplete/src/Autocomplete.js
```diff
@@ -163,10 +168,7 @@ export default class Autocomplete extends PureComponent {
                   index,
                   style,
                   children: itemString,
-                  onMouseUp: () => {
-                    selectItemAtIndex(index)
-                  },
-                  isSelected: itemToString(selectedItem) === itemString,
+                  isSelected: selectedItem === item,
```
Rowno: This avoids an unnecessary string conversion.

src/table/src/TableRow.js
```diff
@@ -144,8 +144,8 @@ class TableRow extends PureComponent {
           innerRef={this.onRef}
           className={cx(themedClassName, className)}
           display="flex"
-          aria-selected={isHighlighted}
-          aria-current={isSelected}
+          data-ishighlighted={isHighlighted}
+          aria-selected={isSelected}
```
Rowno: Downshift adds `aria-selected` and it behaves like `isSelected`. Using both `aria-current` and `aria-selected` also probably isn't correct accessibility wise so I switched `isHighlighted` to a data attribute. I tested in a screen reader and highlights are still being announced correctly by Downshift.

jeroenransijn: Yeah those `aria-` attributes keep tripping me up. We might need to check if this works for all the consumers of `TableRow`.

src/autocomplete/src/Autocomplete.js
Rowno: This is how they do in the Downshift docs though? https://github.com/paypal/downshift#usage

Rowno: I think you're right, I'll revert this change.

src/table/src/TableRow.js
Rowno: I just tested `Menu`, which from what I can tell is the only other consumer of `createRowAppearance` (nothing else uses `TableRow`), and the highlighted / selected states are still working there.

Rowno: `Table` itself also seems to be working fine.

PR #358, opened 2018-10-17
fix passing server-rendered css in ssr-next example

egorovli:
This fix avoid escaping of certain characters in style tag by using `dangerouslySetInnerHTML`. Previously it led to invalid css for some rules.

screenshot 2018-10-17 at 13 06 10
screenshot 2018-10-17 at 13 06 27


jeroenransijn:
Awesome!

PR #370, opened 2018-10-18
Convert NodeList to array with Array.from and prevent XO linter from fixing

sumanbh:
## Summary
* Switch to using the Array object to convert NodeList to an array. Reason being, the ES6 spread is being transpiled incorrectly. See the errors [here](https://evergreen.segment.com/components/menu) and [here](https://evergreen.segment.com/components/table) (try to sort the table in this case)

**Screenshot**

screen shot 2018-10-17 at 9 33 40 pm

### DIFF
--- a/src/menu/src/Menu.js
+++ b/src/menu/src/Menu.js
@@ -34,7 +34,8 @@ export default class Menu extends React.PureComponent {
 
   componentDidMount() {
     // Get the menu item buttons
-    this.menuItems = Array.prototype.slice.call(
+    // eslint-disable-next-line unicorn/prefer-spread
+    this.menuItems = Array.from(
       this.menuRef.querySelectorAll('[role="menuitemradio"], [role="menuitem"]')
     )
 
--- a/src/menu/src/Menu.js
+++ b/src/menu/src/Menu.js
@@ -34,11 +34,9 @@ export default class Menu extends React.PureComponent {
 
   componentDidMount() {
     // Get the menu item buttons
-    this.menuItems = [
-      ...this.menuRef.querySelectorAll(
-        '[role="menuitemradio"], [role="menuitem"]'
-      )
-    ]
+    this.menuItems = Array.prototype.slice.call(
+      this.menuRef.querySelectorAll('[role="menuitemradio"], [role="menuitem"]')
+    )
 
     if (this.menuItems.length === 0) {
       throw new Error('The menu has no menu items')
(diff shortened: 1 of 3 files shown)


src/menu/src/Menu.js
```diff
@@ -34,11 +34,9 @@ export default class Menu extends React.PureComponent {
 
   componentDidMount() {
     // Get the menu item buttons
-    this.menuItems = [
-      ...this.menuRef.querySelectorAll(
-        '[role="menuitemradio"], [role="menuitem"]'
-      )
-    ]
+    this.menuItems = Array.prototype.slice.call(
```
mshwery: I think we could use `Array.from` instead

sumanbh: I'm guessing the original author wrote `Array.from` as well but one of the pre-commit hook (xo) just auto changes/"fixes" it to use spread operator instead, and we're back to the original transpilaton issue.

bchelli: Would making it a two step process prevent the transpilaton issue: 
```javascript
    // Get the menu item buttons
    const menuItemsNodes = this.menuRef.querySelectorAll('[role="menuitemradio"], [role="menuitem"]');

    // Convert to an array
    this.menuItems = [ ...menuItemsNodes ];
```

sumanbh: @bchelli The issue is with the spread itself. Gatsby doesn't realize `menuItemsNodes` is array-like. This is what the result looks like

```javascript
var menuItemsNodes = this.menuRef.querySelectorAll('[role="menuitemradio"], [role="menuitem"]');
this.menuItems = menuItemsNodes.concat();
```

which is invalid. Perhaps an issue with the gatsby/babel configs for docs.

jeroenransijn: @sumanbh is right that some hook changed the code. I initially wrote `Array.from` this also happened in a different case. Not sure what the right solution here is. @Rowno any ideas?

jeroenransijn: image

jeroenransijn: @sumanbh might be nice to have update this too  in `./src/table/src/manageTableCellFocusInteraction.js` and `./src/table/src/manageTableRowFocusInteraction.js`

sumanbh: @jeroenransijn Updated.

Rowno: I think we can just change it back to `Array.from()` and add an `// eslint-disable-next-line unicorn/prefer-spread`

sumanbh: @Rowno Done.

jeroenransijn:
This should fix #383.

jeroenransijn:
https://github.com/segmentio/evergreen/releases/tag/v4.0.1

Issue #180, opened 2018-04-03
Use label prop in the Switch component


jeroenransijn:
The Switch component has an unused label property that should be used (or removed).

prateekgoel:
This issue can be closed now @jeroenransijn

bndw:
Feature request: It would be nice to have the `label` prop for `Switch`.

PR #375, opened 2018-10-18
Remove extra label prop from Switch

prateekgoel:
This PR targets issue https://github.com/segmentio/evergreen/issues/180.
I have removed the extra label prop from Switch component.
If there's a need to implement its use, we can implement it in another PR.


jeroenransijn:
Thanks for cleaning up! Agree that we might add this back in the future with actual support 👍

jeroenransijn:
https://github.com/segmentio/evergreen/releases/tag/v4.0.2

Issue #325, opened 2018-09-06
[v4] Switch should support a "value" prop


Flaque:
# Overview
In order for an `onChange` prop to be effective in a [switch](https://github.com/segmentio/evergreen/tree/master/src/switch), you have to pass a value in with an anonymous function, which is not consistent with other `input` type items in React.  

## How Evergreen currently does it 

```jsx
<Switch onChange={() => this.onChange(myValue)} />
```

## How Evergreen should do it

```jsx
<Switch onChange={this.onChange} value={myValue} />
```

## How other inputs work

```jsx
import { TextInput } from 'evergreen-ui'
<TextInput onChange={this.onChange} value={myValue} />
```

prateekgoel:
I would like to work on this.

Flaque:
@prateekgoel go for it! 🎉

prateekgoel:
Thank you @Flaque

prateekgoel:
@Flaque PR submitted.

jeroenransijn:
The switch is semantically a checkbox, I think it should be the `checked` prop instead of `value`. https://evergreen.segment.com/components/switch

```javascript
<Switch onChange={this.onChange} checked={isChecked} />
```

PR #377, opened 2018-10-18
Support value prop for Switch component

prateekgoel:
This pull request targets https://github.com/segmentio/evergreen/issues/325.
I have added support for value prop.
I am new contributing to this repository. Please suggest any improvements and I would be happy to do the changes.


Flaque:
> Relaying @jeroenransijn's comments in here:

The switch is semantically a checkbox, I think it should be the `checked` prop instead of `value`. https://evergreen.segment.com/components/switch

```javascript
<Switch onChange={this.handleChange} checked={isChecked} value="some value" />
```

The `value` prop is pretty much only useful in non-ajax traditional form submissions.

To give a bit more context you can read about checkboxes here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

We should optimize for passing the `event` into the `onChange` handler.

prateekgoel:
@Flaque I added one commit. Did you intend this?

jeroenransijn:
@prateekgoel we should pass the `event` (e) itself to the `onChange` handler. 

```javascript
handleChange = e => {
    if (isControlled(this)) {
      this.props.onChange(e)
    } else {
      this.setState(({ checked }) => ({
        checked: !checked
      }))
      this.props.onChange(e)
    }
  }
```

prateekgoel:
Got it. I will do the changes. Thanks.

Issue #261, opened 2018-07-25
[v4] Overlay lock body scrolling


jeroenransijn:
Consider adding `overflow: hidden;` to the body when the Overlay is shown:

```javascript
  componentDidMount() {
    document.body.style.overflow = 'hidden'
  }

  componentWillUnmount() {
    document.body.style.overflow = null
  }
```

prateekgoel:
CanI work on this?

mshwery:
This is release as v4.9.0

PR #385, opened 2018-10-20
Add overflow:hidden when overlay is shown

prateekgoel:
Targets https://github.com/segmentio/evergreen/issues/261

Rather than setting `overflow` to `null` on `componentWillUnmount`, I have set it to the default value of overflow. i.e. `visible`

### DIFF
--- a/src/overlay/src/Overlay.js
+++ b/src/overlay/src/Overlay.js
@@ -162,6 +162,10 @@ class Overlay extends React.Component {
     }
   }
 
+  componentDidMount() {
+    document.body.style.overflow = 'hidden'
+  }
+
   componentWillReceiveProps(nextProps) {
     if (nextProps.isShown && !this.props.isShown) {
       this.setState({
@@ -172,6 +176,7 @@ class Overlay extends React.Component {
 
   componentWillUnmount() {
     document.body.removeEventListener('keydown', this.onEsc, false)
+    document.body.style.overflow = 'visible'
   }
 
   /**


src/overlay/src/Overlay.js
```diff
@@ -162,6 +162,10 @@ class Overlay extends React.Component {
     }
   }
 
+  componentDidMount() {
+    document.body.style.overflow = 'hidden'
```
mshwery: Shouldn't this be applied only when `this.props.isShown`?

prateekgoel: @mshwery I am not sure about that.

jeroenransijn: Yeah I think we need to monitor the prop from changing. Maybe we can use `onEnter` and `onExit` event instead.

```jsx
<Transition
  appear
  unmountOnExit
  timeout={ANIMATION_DURATION}
  in={isShown && !exiting}
  onExit={onExit} // onExit={this.handleExit}
  onExiting={this.handleExiting}
  onExited={this.handleExited}
  onEnter={onEnter} // => onEnter={this.handleEnter}
  onEntering={this.handleEntering}
  onEntered={this.handleEntered}
>
```

Then in the `handleEnter/Exit` method we can do the following:

```js
handleEnter = () => {
  this.previousOverflow = document.body.style.overflow
  document.body.style.overflow = 'hidden'
  this.props.onEnter()
}

handleExit = () => {
  document.body.style.overflow = this.previousOverflow
  this.prop.onExit() 
}
```

I think the above should work. @mshwery do you seen any issues here?

kcjpop: Would be nice if this can be moved forward, as I really need this feature.

mshwery:
This is now implemented (just need to make a new release soon)

src/overlay/src/Overlay.js
kcjpop: I use dialog to show content and usually it has its own scrollbar. Without this change, when users reach the end of the dialog and keep scrolling, the page will scroll too. Something like this https://codesandbox.io/s/18lprv8rr3

mshwery:
This is released as v4.9.0

PR #387, opened 2018-10-20
Fixed Menu error 'querySelectorAll of null'.

colshacol:
While browsing the docs, I found an error on the Menu component. I set out to fix the error, and this PR is the solution I came up with.

screen shot 2018-10-20 at 6 13 15 pm

This is the piece of problematic code:

```js
onMenuRef = ref => {
  this.menuRef = ref
}
```

I believe `componentDidMount` was invoking before the `onMenuRef` method was invoked by the ancestral child. So my solution was to change `componentDidMount` to `onRefReady`, which now gets invoked whenever the `onMenuRef` method is invoked.

```js
  onMenuRef = ref => {
    this.menuRef = ref
    if (ref) this.onRefReady()
  }
```

Also, every _other_ time `onMenuRef` was called, `ref` was `null`, so I added the condition to make sure that `ref` existed before trying to invoke `onRefReady`.

One other change I made was to `Positioner.js`, in which I updated to follow the same defensive pattern for invoking `this.props.innerRef`. 

**I am not sure if this is the right fix.** I do not yet fully understand the logic in the `componentDidMount` or what made it necessary to happen only on mount. **But this is the solution that worked for me** in the few minutes I had before having to take the misses out for dinner. :)

### DIFF
--- a/src/menu/src/Menu.js
+++ b/src/menu/src/Menu.js
@@ -32,7 +32,8 @@ export default class Menu extends React.PureComponent {
     children: PropTypes.node
   }
 
-  componentDidMount() {
+  onRefReady() {
+    console.log(this.menuRef)
     // Get the menu item buttons
     // eslint-disable-next-line unicorn/prefer-spread
     this.menuItems = Array.from(
@@ -112,6 +113,7 @@ export default class Menu extends React.PureComponent {
 
   onMenuRef = ref => {
     this.menuRef = ref
+    if (ref) this.onRefReady()
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


colshacol:
I have found that this bug affects the Table component, as well, and have confirmed that this PR fixes it too.

jeroenransijn:
Hey @colshacol thanks for contributing! I think this was fixed by #370 in `v4.0.1` (https://github.com/segmentio/evergreen/releases/tag/v4.0.1).

I have since deployed the documentation, but it seems GitHub is still moving through a backlog of pages builds: https://status.github.com/messages

jeroenransijn:
Closing this PR, let me know if I am missing something and this is actually still valid!

Issue #216, opened 2018-07-10
[v4] Consider `flexShrink={0}` for Avatar component by default


jeroenransijn:
See title.

mshwery:
Already implemented.

PR #390, opened 2018-10-22
Add default flexShrink={0} for Avatar

fraindz:
Addresses #216 
Default flexShrink={0} will be applied for Avatar component.
If property is specifically provided as a prop, default value will be overridden.


jeroenransijn:
Thanks for contributing @fraindz! This looks good.

PR #399, opened 2018-10-24
TagInput component

mshwery:
## What's in this PR?

This PR introduces a new component called `TagInput` which displays a list of values as "tags" (very similar to Badges, in fact we are using a variation of them under the hood) and allows you to enter or remove tags.

taginput

### Highlights

- backspace (when cursor is at position 0 in the input) removes items
- enter keypress appends items to the list
- you can define (or disable) a `separator` so that you can add multiple items in one go. by default splits on comma and new line
- values is controlled, which means it's up to consumers to do the work of actually adding/removing items via `onAdd`, `onRemove`, or conveniently using `onChange` to get the new array.
- prevent input clearing (for instance if an item is invalid) by returning `false` during `onAdd` or `onChange`
- styled the container to match text inputs

### For later PRs:
- Add keyboard navigation support _within_ the input while focused to select tags and delete with keypresses (aka fully keyboard accessible with no mouse)
- Add focus/active styles to tags ^
- Potentially split out Tag into its own component, if it continues to deviate from Badge
- Otherwise, consider moving `removable` into Badge directly


mshwery:
Hah! 

```
  Package size limit has exceeded by 1 KB
  Package size: 201 KB
  Size limit:   200 KB
```

src/badges/src/Badge.js
```diff
@@ -36,20 +46,31 @@ class Badge extends PureComponent {
...
+    } = this.props
 
     const { color, backgroundColor } = theme.getBadgeProps({
-      isSolid,
-      color: propsColor
+      color: propsColor,
+      isSolid
     })
 
+    const appearance = isInteractive ? 'interactive' : 'default'
```
jeroenransijn: Instead of having a different appearance. What about using a `data-interactive` tag so we can use a selector instead?

mshwery: We could do that too. I think this is a little more straightforward in terms of "are interactive styles applied or not"

PR #400, opened 2018-10-24
built instead of build.

chrisdl:
minor spelling thingy that was driving me crazy =).


jeroenransijn:
Different PR with same change got merged in. Closing.

PR #405, opened 2018-10-25
Minor bugfixes for TagInput: shouldClearInput, prop warning, preventDefault

mshwery:
## What's in this PR?

This fixes `shouldClearInput` so that if _either_ `onAdd` or `onChange` explicitly `return false` the input will not be cleared.

This behavior allows consumers to do validation and reject changes while allowing the user a change to fix their input (not have to re-type stuff)

It also introduces `event.preventDefault()` when pressing enter because we do not want it to submit forms. It has special powers inside `TagInput` and we should preserve them.

### DIFF
--- a/src/tag-input/src/TagInput.js
+++ b/src/tag-input/src/TagInput.js
@@ -51,10 +51,10 @@ class TagInput extends React.Component {
      */
     onRemove: PropTypes.func,
     /** Value or RegExp to split on pasted text or on enter keypress */
-    separator: PropTypes.oneOf([
+    separator: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.instanceOf(RegExp),
-      false
+      PropTypes.oneOf([false])
     ]),
     /** Provide props to tag component (actually `Badge`, for now). */
     tagProps: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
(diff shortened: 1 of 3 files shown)


src/tag-input/src/TagInput.js
```diff
@@ -51,10 +53,10 @@ class TagInput extends React.Component {
      */
     onRemove: PropTypes.func,
     /** Value or RegExp to split on pasted text or on enter keypress */
-    separator: PropTypes.oneOf([
+    separator: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.instanceOf(RegExp),
-      false
+      PropTypes.oneOf([false])
```
mshwery: Couldn't find a better way to do this. `false` is allowed but `true` isn't.

src/tag-input/src/TagInput.js
```diff
@@ -80,9 +83,11 @@ class TagInput extends React.Component {
   addTags = (value = '') => {
     const { onAdd, onChange, values } = this.props
     const newValues = this.getValues(value)
-    const shouldClearInput =
-      safeInvoke(onAdd, newValues) ||
-      safeInvoke(onChange, values.concat(newValues))
+    let shouldClearInput = safeInvoke(onAdd, newValues)
+
+    if (typeof onChange === 'function') {
```
mshwery: This fixes the `shouldClearInput` bug. `return false` in either `onAdd` OR `onChange` and the input will not clear 👍

src/tag-input/src/TagInput.js
jeroenransijn: lgtm

src/tag-input/src/TagInput.js
jeroenransijn: Sweet! Makes sense.

PR #410, opened 2018-10-29
[SCH-932] Fix Tooltip on disabled buttons; tooltip doesn't close

colinking:
Bug re-created: http://localhost:6006/?selectedKind=tooltip&selectedStory=Tooltip

Solution WIP


jeroenransijn:
Maybe `Tooltip` should have never render when `disabled` is `true`?

Issue #398, opened 2018-10-24
Extra div wrapper in Combobox prevents full width to work


LukaszGrela:
- Issue and steps to reproduce.

  - Insert `Pane` and inside add a `Combobox`
  - Set `width="100%"` on the `Combobox`
 - or set `flexGrow={1}` on the `Combobox`

e.g.

```
  render() {
    return (
      <Pane
        display="flex"
        background="greenTint"
        padding={16}>
        <Combobox
          width="100%"
          defaultSelectedItem={{ label: 'Banana' }}
          items={[{ label: 'Banana' }, { label: 'Pear' }, { label: 'Kiwi' }]}
          itemToString={item => item.label}
          onChange={selected => console.log(selected)}
          placeholder="Choose project code"
        />
      </Pane>
    );
  }
```

which renders as:
```
<div class="📦bg_y3l8zn 📦dspl_flex 📦pb_16px 📦pl_16px 📦pr_16px 📦pt_16px 📦box-szg_border-box css-nil">
  <div>
    <div class="📦dspl_inline-flex 📦w_100prcnt 📦box-szg_border-box">
      <input class="css-12835yy 📦color_425A70 📦fnt-fam_b77syt 📦fnt-sze_12px 📦f-wght_400 📦ln-ht_16px 📦ltr-spc_0 📦w_0px 📦h_32px 📦pl_10px 📦pr_10px 📦bblr_3px 📦bbrr_0px 📦btlr_3px 📦btrr_0px 📦flx_1 📦box-szg_border-box"
        type="text" placeholder="Choose project code" spellcheck="true" aria-invalid="false" role="combobox"
        aria-autocomplete="list" aria-expanded="false" autocomplete="off" id="downshift-0-input" value="Banana">
      <button class="css-18rijld 📦color_66788A 📦fnt-fam_b77syt 📦mt_0px 📦fnt-sze_12px 📦f-wght_500 📦ln-ht_32px 📦ltr-spc_0 📦btrr_3px 📦bbrr_3px 📦btlr_0px 📦bblr_0px 📦pt_0px 📦pb_0px 📦pr_0px 📦pl_0px 📦ml_-1px 📦mr_0px 📦mb_0px 📦h_32px 📦pst_relative 📦dspl_flex 📦algn-itms_center 📦flx-wrap_nowrap 📦w_32px 📦just-cnt_center 📦box-szg_border-box"
[...]

PR #412, opened 2018-10-31
Make combobox to take full width of parent el

kcjpop:
This PR will attempt to solve #398. Initially I thought the solution is to set inline style for that DIV `<div style={{ width: '100%' }}>` but finally decided to use `Pane` for a more "evergreen"-way of doing this.

Any feedback is pretty much appreciated.

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -199,9 +199,10 @@ export default class Autocomplete extends PureComponent {
           selectedItem,
           highlightedIndex,
           selectItemAtIndex,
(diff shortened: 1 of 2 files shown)


jeroenransijn:
@kcjpop thanks for contributing! Looking good so far. Could you provide some screenshots to illustrate the new behavior?

kcjpop:
Definitely. Here is the change in Safari, Chrome and Firefox.

screenshot 2018-10-31 at 22 02 31
screenshot 2018-10-31 at 22 03 14
screenshot 2018-10-31 at 22 05 51

PR #415, opened 2018-11-01
Runtime warnings

vadimdemedes:
This PR starts an initiative to introduce runtime warnings in development to help developers notice and fix common mistakes. Being an active user of Evergreen, I still use onClick prop instead of onSelect sometimes. After this is merged, this is what I would see if I did it again:

cleanshot 2018-10-31 at 21 24 36 2x

Happy to tweak error messages, if needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "react-tiny-virtual-list": "^2.1.4",
     "react-transition-group": "^2.5.0",
     "tinycolor2": "^1.4.1",
-    "ui-box": "^1.4.0"
+    "ui-box": "^1.4.0",
+    "warning": "^4.0.2"
   },
   "peerDependencies": {
     "react": "^16.3.0",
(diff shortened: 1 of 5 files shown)


vadimdemedes:
> Is there a way to flag this for removal from most bundlers?
It will be auto removed on minify step. It's the same way warnings are shown in React, Redux, etc.

> I don't know that the extra package is worth these warnings.
This package weighs 130 bytes after minification.

> I think the typeof checking is good enough.
`warning` doens't check anything, it's responsible only for showing warnings and ensuring there's only one warning shown per session.

jeroenransijn:
@vadimdemedes as a fyi, @Rowno did this in `ui-box` https://github.com/segmentio/ui-box/blob/756f9f915bb89cfb8c0805200249ad63f9f0ea00/src/box.js#L35

Not sure what tradeoffs would be

vadimdemedes:
What tradeoffs do you mean? I don't follow.

mshwery:
LGTM then

PR #420, opened 2018-11-01
Add onBeforeClose to SideSheet and Overlay for conditional closing

jeroenransijn:
## Use case

When a `SideSheet` contains unsaved user state the `onBeforeClose` method can conditionally prevent the `SideSheet` from closing when there are unsaved changes.

## Example

sidesheet-onbeforeclose

## Code

```js
<Component
  initialState={{
    value: 'Change me',
    newValue: 'Change me',
    isShown: false,
    hasChanges: false,
    isLoading: false,
    isConfirmationShown: false
  }}
>
  {({ state, setState }) => (
    <Box>
      <SideSheet
        isShown={state.isShown}
        onBeforeClose={() => {
          if (state.hasChanges) {
            setState({ isConfirmationShown: true })
            return false
          }
          return true
        }}
      >
        {({ close }) => {
          return (
            <Box padding={40}>
              <Dialog
                intent="danger"
                isShown={state.isConfirmationShown}
                title="You have unsaved changes"
                onConfirm={closeDialog => {
                  setState(
                    {
                      newValue: state.value,
                      hasChanges: false
                    },
                    () => {
                      closeDialog()
                    }
                  )
                }}
                onCloseComplete={() => {
                  setState({
                    isConfirmationShown: false
                  })
                  if (!state.hasChanges) {
                    close()
                  }
                }}
                confirmLabel="Discard Changes"
              >
                Are you sure you want to discard your unsaved changes?
              </Dialog>
              <Pane>
                <TextInput
                  value={state.newValue}
                  onChange={e => {
                    setState({
                      newValue: e.target.value,
                      hasChanges: state.value !== e.target.value
                    })
                  }}
                  marginRight={16}
                />
                <Button
                  appearance="primary"
                  isLoading={state.isLoading}
                  disabled={!state.hasChanges}
                  onClick={() => {
                    setState({ isLoading: true })
                    setTimeout(() => {
                      setState(
                        {
                          value: state.newValue,
                          hasChanges: false,
                          isLoading: false
                        },
                        close
                      )
                    }, 2000)
                  }}
                >
                  Save
                </Button>
              </Pane>
              <Paragraph marginTop={16}>
                Current value: {state.value}
              </Paragraph>
            </Box>
          )
        }}
      </SideSheet>
      <Button onClick={() => setState({ isShown: true })}>
        Show Side Sheet
      </Button>
    </Box>
  )}
</Component>
```

### DIFF
--- a/src/overlay/src/Overlay.js
+++ b/src/overlay/src/Overlay.js
@@ -91,6 +91,14 @@ class Overlay extends React.Component {
     shouldCloseOnEscapePress: PropTypes.bool,
 
     /**
+     * Function called when overlay is about to close.
+     * Return `true` to close the sheet. (default)
+     * Return `false` to prevent the sheet from closing.
+     * type: `Function -> Boolean`
+     */
+    onBeforeClose: PropTypes.func,
+
+    /**
      * Callback fired before the "exiting" status is applied.
      * type: `Function(node: HtmlElement) -> void`
      */
@@ -145,6 +153,7 @@ class Overlay extends React.Component {
     onHide: () => {},
     shouldCloseOnClick: true,
     shouldCloseOnEscapePress: true,
+    onBeforeClose: () => true,
     onExit: () => {},
     onExiting: () => {},
     onExited: () => {},
@@ -245,7 +254,9 @@ class Overlay extends React.Component {
   }
 
   close = () => {
-    this.setState({ exiting: true })
+    if (this.props.onBeforeClose()) {
+      this.setState({ exiting: true })
+    }
   }
 
   handleEntering = node => {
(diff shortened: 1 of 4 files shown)


src/overlay/src/Overlay.js
```diff
@@ -90,6 +90,14 @@ class Overlay extends React.Component {
      */
     shouldCloseOnEscapePress: PropTypes.bool,
 
+    /**
+     * Function called when overlay is about to close.
+     * Return `true` to close the sheet. (default)
+     * Return `false` to prevent the sheet from closing.
+     * type: `Function -> Boolean`
+     */
+    onBeforeClose: PropTypes.func,
```
mshwery: Would it make more sense to call this `onRequestClose`? Where only explicit `false` returns would prevent closing?

jeroenransijn: Hey Matt!  I know you have some similar patterns like this in `Pagination` currently, and I think there is a real use case for it. But for this I think `onBeforeClose` sounds slightly simpler and more in line with browser events such as `onbeforeunload`.

PR #438, opened 2018-11-09
Adds filterProps prop

### LINKED ISSUE #431: Change placeholder in SelectMenu search input
Is there a way or prop to modify the placeholder from the search input?

captura de pantalla 2018-11-05 a la s 21 41 57

lndgalante:
Fix #431 adding `filterProps` prop to SelectMenu so we can modify the SearchTableHeaderCell props like for example the placeholder.

### DIFF
--- a/src/select-menu/src/SelectMenu.js
+++ b/src/select-menu/src/SelectMenu.js
@@ -77,7 +77,12 @@ export default class SelectMenu extends PureComponent {
      * rendered on the right side of the Select Menu to give additional
      * information when an option is selected.
      */
-    detailView: PropTypes.oneOfType([PropTypes.func, PropTypes.node])
+    detailView: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
+
+    /**
+     * Change the filters options in Option List
+     */
+    filterProps: PropTypes.shape({ placeholder: PropTypes.string })
   }
 
   static defaultProps = {
@@ -115,6 +120,7 @@ export default class SelectMenu extends PureComponent {
       hasFilter,
       detailView,
       isMultiSelect,
+      filterProps,
       ...props
     } = this.props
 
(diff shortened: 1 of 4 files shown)


src/select-menu/src/SelectMenu.js
```diff
@@ -77,7 +77,12 @@ export default class SelectMenu extends PureComponent {
      * rendered on the right side of the Select Menu to give additional
      * information when an option is selected.
      */
-    detailView: PropTypes.oneOfType([PropTypes.func, PropTypes.node])
+    detailView: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
+
+    /**
+     * Change the filters options in Option List
+     */
+    filterProps: PropTypes.shape({ placeholder: PropTypes.string })
```
mshwery: Curious if this should be broader in terms of which properties are allowed?

src/select-menu/src/SelectMenuContent.js
```diff
@@ -16,6 +16,7 @@ export default class SelectMenuContent extends PureComponent {
     hasTitle: PropTypes.bool,
     hasFilter: PropTypes.bool,
     listProps: PropTypes.shape(OptionsList.propTypes),
+    filterProps: PropTypes.shape({ placeholder: PropTypes.string }),
```
mshwery: Seeing `listProps` right next to this, it seems like `filterProps` might be redundant? Can/should we hook into the existing `listProps` prop?

src/select-menu/src/SelectMenu.js
```diff
@@ -132,6 +138,7 @@ export default class SelectMenu extends PureComponent {
             hasFilter={hasFilter}
             hasTitle={hasTitle}
             isMultiSelect={isMultiSelect}
+            filterProps={filterProps}
             listProps={{
```
mshwery: I'm thinking we can spread `filterProps` here into `listProps`?

PR #445, opened 2018-11-14
Add emptyView to <SelectMenu>

vadimdemedes:
Adds prop to display a custom empty view when there are no options passed to `<SelectMenu>`.

cleanshot 2018-11-14 at 10 54 13 2x

```jsx
<SelectMenu
  title="Empty view"
  options={[]}
  emptyView={(
    <Pane height="100%" display="flex" alignItems="center" justifyContent="center">
      <Text size={300}>No options found</Text>
    </Pane>
  )}
>
  <Button>Select option...</Button>
</SelectMenu>
```

### DIFF
--- a/docs/src/pages/components/select-menu.mdx
+++ b/docs/src/pages/components/select-menu.mdx
@@ -131,6 +131,41 @@ Available positions:
 </Component>
 ```
 
+## Empty view
+
+It's possible to display a custom empty view instead of options list via `emptyView`, when there are no properties supplied.
+Note that empty view won't be shown when options are being filtered and there are no search results.
+
+```jsx
+<SelectMenu
+  title="Empty view"
+  options={[]}
+  emptyView={(
+    <Pane height="100%" display="flex" alignItems="center" justifyContent="center">
+      <Text size={300}>No options found</Text>
+    </Pane>
+  )}
+>
+  <Button>Select option...</Button>
+</SelectMenu>
+```
+
+It's also possible to close `<SelectMenu>` from within empty view:
+
+```jsx
+<SelectMenu
+  title="Empty view"
+  options={[]}
+  emptyView={({ close }) => (
+    <Pane height="100%" display="flex" alignItems="center" justifyContent="center">
+      <Button onClick={close}>Close</Button>
+    </Pane>
+  )}
+>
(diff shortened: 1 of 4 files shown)


docs/src/pages/components/select-menu.mdx
```diff
@@ -131,6 +131,41 @@ Available positions:
 </Component>
 ```
 
+## Empty view
+
+It's possible to display a custom empty view instead of options list via `emptyView`, when there are no properties supplied.
+Note that empty view won't be shown when options are being filtered and there are no search results.
```
mshwery: Do/should we have a "no results" view as well?

src/select-menu/src/SelectMenuContent.js
```diff
@@ -81,14 +88,19 @@ export default class SelectMenuContent extends PureComponent {
...
-            height={optionsListHeight}
-            hasFilter={hasFilter}
-            options={options}
-            isMultiSelect={isMultiSelect}
-            close={close}
-            {...listProps}
-          />
+
+          {options.length === 0 && hasEmptyView ? (
+            <Pane height={optionsListHeight}>{emptyView}</Pane>
```
mshwery: Should we push this deeper (into `OptionsList` itself)?

docs/src/pages/components/select-menu.mdx
vadimdemedes: That would be a separate PR and a separate prop.

src/select-menu/src/SelectMenuContent.js
vadimdemedes: No, because then `emptyView` would appear when there are no search results after filtering. There should be a separate prop for it.

vadimdemedes: `emptyView` is useful for CTA views, for example when there menu is responsible for selecting warehouses and customer didn't connect any yet. Empty view could say that and show a button to go to connect a new warehouse. It would be incorrect to show the same view for empty search results.

mshwery: Yep, I agree with all of that. I was wondering if OptionsList should be responsible for those distinctions though. Sounds like that's not the case in your opinion.

Thanks!

PR #448, opened 2018-11-16
Fix manual control over <Popover> visibility

vadimdemedes:
Explicitly passing `isShown = true/false` didn't affect `<Popover>`'s behavior and didn't show/hide the popover correctly. This PR fixes it.

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -92,7 +92,6 @@ export default class Popover extends Component {
 
   static defaultProps = {
     position: Position.BOTTOM,
-    isShown: false,
     minWidth: 200,
     minHeight: 40,
     animationDuration: 300,
@@ -316,7 +315,8 @@ export default class Popover extends Component {
     } = this.props
     const { isShown: stateIsShown } = this.state
 
-    const shown = isShown || stateIsShown
+    // If `isShown` is a boolean, popover is controlled manually, not via mouse events
+    const shown = typeof isShown === 'boolean' ? isShown : stateIsShown
 
     return (
       <Positioner
--- a/src/positioner/src/Positioner.js
+++ b/src/positioner/src/Positioner.js
@@ -119,13 +119,6 @@ export default class Positioner extends PureComponent {
     this.state = initialState()
   }
 
-  componentDidMount() {
-    // If `isShown` is `true` on first render, <Transition> won't emit any enter/entered events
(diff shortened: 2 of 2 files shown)


src/popover/src/Popover.js
```diff
@@ -316,7 +315,8 @@ export default class Popover extends Component {
     } = this.props
     const { isShown: stateIsShown } = this.state
 
-    const shown = isShown || stateIsShown
+    // If `isShown` is a boolean, popover is controlled manually, not via mouse events
+    const shown = typeof isShown === 'boolean' ? isShown : stateIsShown
```
mshwery: I think this might break some of our consumers in the app that only use isShown for initial display (not using it as a controlled component)

src/positioner/src/Positioner.js
```diff
@@ -119,6 +119,13 @@ export default class Positioner extends PureComponent {
     this.state = initialState()
   }
 
+  componentDidMount() {
+    // If `isShown` is `true` on first render, <Transition> won't emit any enter/entered events
+    // because there was no transition from `false` to `true`
+    // In order to still display and position the content properly, `update()` must be fired after first render
+    this.update()
```
mshwery: Should transition emit on appear instead of this?

src/popover/src/Popover.js
vadimdemedes: I don't understand, could you point to an example?

src/positioner/src/Positioner.js
vadimdemedes: That's a good idea, will try that.

src/popover/src/Popover.js
mshwery: I think we have some places where we set `setState({ isShown: true })` without also updating `setState({ isShown: false })` on cancel/confirm/close.

vadimdemedes: I see, I'll check it out.

vadimdemedes: I checked all cases where `<Popover>` is being used and I couldn't find any examples where `isShown` prop is being passed. Could you point me to those places you had in mind?

mshwery: I might have been thinking of `Dialog`

vadimdemedes: Ah ok then.

PR #453, opened 2018-11-19
Proposal: DatePicker component [WIP]

kcjpop:
I built a date picker for my project and thought it would be useful for other Evergreen's users. If the team already has a plan to build this component in future versions, please let me know 🙂

Lots of inspiration come from Antd and Atlaskit.

### Functionalities

The DatePicker itself does not expose any input, but just a calendar where users can select a date. By doing this way it's possible to show DatePicker inline, or customize to use with `TextInput`, `Button` or any other controls.

image
_Inline date pickers, with Today button and dates being disabled on demand_

image
_Date input using `Popover` and `TextInput`_

image
_Can be used to make a date range input_

image
image
_Supports l10n_

### Props

- `value {string|number|Date}`: (required) the initial date or value of DatePicker
- `shouldShowTodayButton {boolean}`: if a button is displayed to jump to the current date
- `todayButtonLabel {string}`: label for ShowToday button
- `locale {string}`: locale to format dates, as defined in `Intl.DateTimeFormat`
- `localeOptions {object}`: options to format dates and weekdays' name, as defined in `Intl.DateTimeFormat`
- `disableDates {Date -> boolean}`: function to determine if a date should be disabled
- `onChange {Date -> void}`: callback function when a date is selected

### Todos

- [ ] Allow to show weeks starting from Mondays
- [ ] Support a11y via ARIA attributes (don't have much experience about this)
- [ ] Most likely support keyboard usage
- [ ] Add unit tests for date calculating

### Final thoughts

**Should we make `DateInput` and `DateRangeInput`?** It's up to consumers to customize the UI of their date pickers, but I guess we can export ready-made controls for most common use-cases.

**Do we need a time picker too?** It could mimic the one in Atlaskit.

**Look and feel** I'm not a designer so pretty much unsure if the UI matched with Evergreen design guideline. Please have comments for this.

All feedback are pretty much appreciated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@blueprintjs/icons": "^3.2.0",
     "arrify": "^1.0.1",
     "classnames": "^2.2.6",
+    "date-fns": "^1.29.0",
     "dom-helpers": "^3.2.1",
     "downshift": "^1.31.16",
     "fuzzaldrin-plus": "^0.6.0",
--- a/src/date-picker/index.js
+++ b/src/date-picker/index.js
@@ -1,0 +1,1 @@
+export { default as DatePicker } from './src/DatePicker'
--- a/src/date-picker/index.js
+++ b/src/date-picker/index.js
@@ -1,1 +1,3 @@
+export { default as InlineDatePicker } from './src/InlineDatePicker'
+export { default as DateRangePicker } from './src/DateRangePicker'
 export { default as DatePicker } from './src/DatePicker'
--- a/src/date-picker/src/Calendar.js
+++ b/src/date-picker/src/Calendar.js
@@ -54,7 +54,8 @@ function makeCalendarData(pivotDate) {
 
   const present = makeDaysArray(addDays(firstDay, -1), totalDays, {
     isCurrentMonth: true,
-    isToday: date => isSameDay(date, today)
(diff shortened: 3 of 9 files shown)


vadimdemedes:
Hey @kcjpop, I appreciate your incredible effort in implementing this component, but I don't think it should be in Evergreen just yet. Date picker is a tricky component, in a way that everyone has custom needs for it. We actually have an internal date picker implementation already, which looks like this:



As you see, it's something very different from the standard date picker look. Even our design team is not settled on it yet either, it's something we still experiment on. With that said, feel free to release it as a standalone component :)

kcjpop:
Hi @vadimdemedes, thank you so much for the feedback. No worry, I understand it's not that easy to come up with a general solution for date pickers. Love your screenshot though. Hopefully we can use the official DatePicker component soon :smile: V5 maybe?

vadimdemedes:
@kcjpop Yep, it's probably going to end up in v5! Thank you so much for your understanding :)

mikecole:
@kcjpop Did you ever release this as a standalone component?

kcjpop:
@mikecole Ah no. I was waiting for v5 and finally moved to another project that didn't use Evergreen 😅

PR #461, opened 2018-11-28
Add scroll props to Table.VirtualBody

creedarky:
Adds props from `react-tiny-virtual-list` to control the scroll of the table
image

### DIFF
--- a/docs/src/components/examples/VirtualTable.js
+++ b/docs/src/components/examples/VirtualTable.js
@@ -40,42 +40,37 @@ const users = range(1000)
 
 export default class VirtualTable extends React.PureComponent {
   state = {
-    scrollToIndex: '',
-    scrollOffset: '',
-    scrollToAlignment: ''
+    scrollToIndex: null,
+    scrollOffset: null,
+    scrollToAlignment: null
+  }
+
+  setValue = property => {
+    return e => {
+      const value = e.target.value ? parseInt(e.target.value, 10) : null
+      this.setState({
+        [property]: value
+      })
+    }
   }
 
   render() {
-    const scrollOffset =
-      this.state.scrollOffset.length > 0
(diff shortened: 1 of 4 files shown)


src/table/src/TableVirtualBody.js
```diff
@@ -46,7 +46,20 @@ export default class TableVirtualBody extends PureComponent {
...
+     */
+    scrollToIndex: PropTypes.number,
+    /**
+     * The scrollOffset property passed to react-tiny-virtual-list
+     */
+    scrollOffset: PropTypes.number,
+    /**
+     * The scrollToAlignment property passed to react-tiny-virtual-list
+     */
+    scrollToAlignment: PropTypes.oneOf(['start', 'center', 'end', 'auto'])
```
mshwery: Since you are just passing props to another lib it's probably ok to omit the storybook, but maybe since we are only supporting these specific props it would be useful to add those to the docs and explain why you might need them (why we're exposing them)

creedarky: Added to the doc and to the example as inputs to control it

mshwery: Great! Thank you for doing that. Would you mind posting a screenshot or gif of the new addition to the docs?

docs/src/components/examples/VirtualTable.js
```diff
@@ -36,36 +39,93 @@ const users = range(1000)
   })
 
 export default class VirtualTable extends React.PureComponent {
+  state = {
+    scrollToIndex: '',
```
vadimdemedes: Since `scrollToIndex` and `scrollOffset` are not strings, but numbers, default values here should be `null`, not `''`.

creedarky: Nvm, fixed

PR #467, opened 2018-12-22
Fix empty input in ComboBox

### LINKED ISSUE #447: Combobox breaks when input is blank
Great job on the library! I really like the approach and decision you've made. 👍 
Here's something I found while using the combobox. 

- Issue 
Combobox breaks when the input is cleared.

- Steps to reproduce
1. Go to https://evergreen.segment.com/components/combobox
2. Scroll to the section "Custom items and default selected item"
3. Select the text on the input and delete it

- Versions
"evergreen-ui": "^4.7.0"

- Screenshots.
evergreen-combobox

- Expected.
User can leave the value blank.

- Actual.
The component breaks throwing an error:  
`TypeError: Cannot read property 'label' of null`

- Link to your project.
N/A

- Browser name and version.
Chrome 70

- Operating system.
Mac OS

jyash97:
Fixes #447

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -155,7 +155,10 @@ export default class Autocomplete extends PureComponent {
             scrollToAlignment="auto"
             renderItem={({ index, style }) => {
               const item = items[index]
-              const itemString = itemToString(item)
+              const itemString = item ? itemToString(item) : ''
+              const selectedItemString = selectedItem
+                ? itemToString(selectedItem)
+                : ''
               return renderItem(
                 getItemProps({
                   item,
@@ -166,7 +169,7 @@ export default class Autocomplete extends PureComponent {
                   onMouseUp: () => {
                     selectItemAtIndex(index)
                   },
-                  isSelected: itemToString(selectedItem) === itemString,
+                  isSelected: selectedItemString === itemString,
                   isHighlighted: highlightedIndex === index
                 })
               )


colinking:
Thanks for the PR @jyash97! Just ran into this problem myself.

Looks like you'll need to wrap the `itemToString` prop we pass to `Downshift` with a `null` handler: https://github.com/segmentio/evergreen/blob/0a49844152d02248c8aa11121a847aaafd6713a7/src/autocomplete/src/Autocomplete.js#L197

For example (roughly):

```js
const wrappedItemToString = (i => i ? itemToString(i) : '')

return (
      <Downshift {...props} defaultSelectedItem={defaultSelectedItem} itemToString={wrappedItemToString}>
```

Since `Downshift` expects an `itemToString` that can handle `null` values: https://github.com/downshift-js/downshift#itemtostring

You can verify this with the "Custom item objects" example in the storybook (just run `yarn dev`).

Otherwise, cc @mshwery on whether we want to expose `null` handling in `itemToString`. IMO, this seems like a sane default, since I can't think of a use case.

jyash97:
Hey @colinking thanks for the review.
This sounds good to me we can handle it on AutoComplete , & also I have tested this implementation this works properly.

jyash97:
Hey @colinking ,
Thanks for the review. I tried the same approach of wrapping the `itemToString` function and return an empty string. 

But consider this : 

If we will clear the input we get a empty string back and trying to access `label` from the string, which will break things.
```js
 <Combobox
   defaultSelectedItem={customItems[0]}
   items={customItems}
   itemToString={i => i.label} // i will be empty string and this will break things again
   onChange={handleChange}
 />
```

Thoughts @colinking @mshwery

PR #468, opened 2018-12-28
Feat/body scroll

### LINKED ISSUE #261: [v4] Overlay lock body scrolling

mshwery:
This introduces scroll-locking support for anything consuming Overlay. It's opt-in, which makes this a non-breaking change.

When you pass `preventBodyScrolling` to Overlay, Dialog, or SideSheet the body content will no longer be scrollable (opposite the default behavior of these components). When the component unmounts or `isShown={false}` then the scroll behavior will return.

I've also updated the docs and stories.

Replaces #385 and closes #261 

cc @jeroenransijn @prateekgoel @kcjpop

### DIFF
--- a/docs/src/pages/components/dialog.mdx
+++ b/docs/src/pages/components/dialog.mdx
@@ -55,7 +55,7 @@ This is normally the button that triggered the `Dialog`.
 
 ## Default with a danger intent
 
-The `intent` prop determines the appearance of the confirm button, `danger` is red. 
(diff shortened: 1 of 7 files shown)


vadimdemedes:
@mshwery Could you please use more descriptive PR title next time?

PR #474, opened 2019-01-14
Check if invalid aria is true rather than present

### LINKED ISSUE #464: SelectField isInvalid not treated as boolean
The `<SelectField />` component has a property `isInvalid`, which should be treated as a boolean (as with the `<TextInputField />` component).

It seems the component is checking for presence of the property, rather than it's value.

## What I did

Set `isInvalid={false}` on the `<SelectField />` component:

```jsx
<SelectField
  isInvalid={false}
  required
  label="A required text input field"
  description="This is a description."
>
  <option value="foo" checked>Foo</option>
  <option value="bar">Bar</option>
</SelectField>
```

## What I expected to happen

The component to render with a _valid_ state (and no red border) - just like the `<TextInputField />` component.

## What happened

The component rendered with the _invalid_ state. From your live docs:

screenshot 2018-12-07 16 28 29

## Workaround

Remove `isInvalid` entirely so the field renders as expected.

---

Thanks for the awesome component library 🙌

dillonhafer:
Fixes #464

### DIFF
--- a/src/select/src/Select.js
+++ b/src/select/src/Select.js
@@ -85,7 +85,8 @@ class Select extends PureComponent {
 
   static defaultProps = {
     appearance: 'default',
-    height: 32
+    height: 32,
+    isInvalid: false
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Could you also modify https://github.com/segmentio/evergreen/blob/master/src/select/src/Select.js#L136 to be `aria-invalid={String(isInvalid)}`?

vadimdemedes:
This will also involve adding a default `false` value for `isInvalid` prop.

dillonhafer:
Sure thing!

vadimdemedes:
Thanks!

PR #477, opened 2019-01-18
Add titleView to SelectMenu

notfelineit:
**This PR:**
- Fixes `onFilterChange` warning in the console
- Adds an optional tooltip `?` icon option to the SelectMenu header. (Needed for some Protocols UI work)

selectmenu

If this seems to specialized, and we should instead pass a render function for the header (or something else!) let me know :-)

### DIFF
--- a/src/select-menu/src/SelectMenu.js
+++ b/src/select-menu/src/SelectMenu.js
@@ -3,6 +3,9 @@ import PropTypes from 'prop-types'
 import arrify from 'arrify'
 import { Popover } from '../../popover'
 import { Position } from '../../constants'
+import { Pane } from '../../layers'
+import { Heading } from '../../typography'
+import { IconButton } from '../../buttons'
 import SelectMenuContent from './SelectMenuContent'
 import OptionShapePropType from './OptionShapePropType'
 import SelectedPropType from './SelectedPropType'
@@ -13,12 +16,6 @@ export default class SelectMenu extends PureComponent {
      * The title of the Select Menu.
      */
     title: PropTypes.string,
-
-    /**
-     * The content for the tooltip in the header of the Select Menu.
-     */
-
-    tooltipContent: PropTypes.string,
 
     /**
      * The width of the Select Menu.
@@ -62,11 +59,6 @@ export default class SelectMenu extends PureComponent {
     hasTitle: PropTypes.bool,
 
     /**
-     * When true, show the tooltip.
-     */
-    hasTooltip: PropTypes.bool,
-
-    /**
      * When true, show the filter.
      */
     hasFilter: PropTypes.bool,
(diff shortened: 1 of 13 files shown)


vadimdemedes:
I think this may be too small of a change to trigger such major addition as header render function.

What do you think about removing tooltip in favor of putting text below the header? Reason I'm suggesting this is to avoid 2 popovers at the same time (select menu and tooltip).

cleanshot 2019-01-18 at 19 53 21 2x

Either way, I'm good with this change 👍

src/select-menu/src/SelectMenuContent.js
```diff
@@ -45,9 +49,11 @@ export default class SelectMenuContent extends PureComponent {
       height,
       options,
       hasTitle,
+      hasTooltip,
       hasFilter,
       close,
       title,
+      tooltipContent,
```
vadimdemedes: Could you rename this prop to `tooltip`, so that it matches `hasTooltip` prop naming?

notfelineit:
@vadimdemedes That's a good idea! Will it work with really long helper text?

notfelineit:
@mshwery I agree 🤔 Before the tooltip, I had a subtitle (like Vadim's suggestion) with the following: `To create a new tag type a key:value format into the search field`. But that wrapped and made the title wonky, hence the tooltip. Do you guys have any suggestions on how else to surface this to users within the SelectMenu? Maybe as a detailPane?

src/select-menu/src/SelectMenuContent.js
```diff
@@ -47,15 +51,16 @@ export default class SelectMenuContent extends PureComponent {
       hasTitle,
       hasFilter,
       close,
-      title,
       listProps,
+      titleView,
       detailView,
       emptyView,
+      headerHeight,
```
mshwery: Should we provide a default value?

notfelineit: Absorootintootinly

src/select-menu/src/SelectMenuContent.js
```diff
@@ -68,26 +74,7 @@ export default class SelectMenuContent extends PureComponent {
...
-              </Pane>
-              <IconButton
-                icon="cross"
-                appearance="minimal"
-                height={24}
-                onClick={close}
-              />
-            </Pane>
-          )}
+          {hasTitle && titleView}
```
mshwery: Hm, I think we should provide the below (removed) jsx as the _default prop_ value for `titleView`. That way consumers of this component can continue using it as is.

The way you've moved the code around makes this a breaking change, technically.

src/select-menu/src/SelectMenu.js
```diff
@@ -128,6 +138,40 @@ export default class SelectMenu extends PureComponent {
     return {}
   }
 
+  getTitleView = (close, title, headerHeight, titleView) => {
```
mshwery: I think you could get rid of this code in favor pushing it into the component that actually uses `titleView` instead.

src/select-menu/src/SelectMenuContent.js
```diff
@@ -36,26 +42,48 @@ export default class SelectMenuContent extends PureComponent {
   static defaultProps = {
     options: [],
     hasTitle: true,
-    hasFilter: true
+    hasFilter: true,
+    titleView: ({ close, title, headerHeight }) => (
```
mshwery: Nit: I would pull this out into a separate component so the defaultProps are easier to scan.

But this is exactly what I was thinking of!

notfelineit: Whoo! ok, I'll do that.

PR #481, opened 2019-01-24
Add autoFocus prop for EditableCell

calthomson:
# What's in this PR?

This PR adds an `autoFocus` prop to the EditableCell component. When `autoFocus` is true the EditableCell will initialize with it's `isEditing` state set to true. The purpose of this is to allow this component to initialize in an editable state, so the user can immediately start typing into the cell.

### DIFF
--- a/src/table/src/EditableCell.js
+++ b/src/table/src/EditableCell.js
@@ -54,7 +54,8 @@ class EditableCell extends React.PureComponent {
 
   static defaultProps = {
     size: 300,
-    isSelectable: true
+    isSelectable: true,
+    autoFocus: false
   }
 
   static getDerivedStateFromProps(props, state) {
--- a/src/table/src/EditableCell.js
+++ b/src/table/src/EditableCell.js
@@ -44,7 +44,12 @@ class EditableCell extends React.PureComponent {
     /**
      * Function called when value changes. (value: string) => void.
      */
-    onChange: PropTypes.func
+    onChange: PropTypes.func,
+
+    /**
+     * When true, the cell will initialize in the editing state.
+     */
+    autoFocus: PropTypes.bool
   }
 
   static defaultProps = {
@@ -62,7 +67,8 @@ class EditableCell extends React.PureComponent {
   }
 
   state = {
-    value: this.props.children
+    value: this.props.children,
+    isEditing: this.props.autoFocus
   }
 
   onMainRef = ref => {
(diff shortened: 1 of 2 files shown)


src/table/src/EditableCell.js
```diff
@@ -44,7 +44,12 @@ class EditableCell extends React.PureComponent {
     /**
      * Function called when value changes. (value: string) => void.
      */
-    onChange: PropTypes.func
+    onChange: PropTypes.func,
+
+    /**
+     * When true, the cell will initialize in the editing state.
+     */
+    autoFocus: PropTypes.bool
```
mshwery: We should probably provide a default value of `false` in `defaultProps`.

src/table/src/EditableCell.js
```diff
@@ -62,7 +67,8 @@ class EditableCell extends React.PureComponent {
   }
 
   state = {
-    value: this.props.children
+    value: this.props.children,
+    isEditing: this.props.autoFocus
```
mshwery: What happens if the component has already initialized/mounted and the `autoFocus` prop changes after the fact? Should changes to the prop be propagated to state?

calthomson: I don't have a strong opinion on this, but I don't think it would be a bad thing to give the parent of `EditableCell` the ability to make the cell editable. I could imagine a scenario where it could be useful.

I'm imagining an editable table. A row is added, and is auto focused. A second row is added and takes the focus. Now the user deletes the second row. Maybe we want the first row to become focused again. I think in this case we would need to be able to use the autoFocused prop to update the `isEditing` state of the cell. Maybe the prop should be `focused` instead of `autoFocused` in this case. What do you think?

mshwery: `autoFocus` is kind of an interesting prop. It maps to `autofocus` which is a real `input` attribute and its behavior is to make the input focused upon presentation. That said I don't think it makes much sense for us to respect post-mount changes to the prop because that would seemingly deviate from the native behavior.

(I agree with your thoughts on changing the name of the prop should we change the behavior)

Thoughts?

calthomson: Yeah, originally I chose the name `autoFocus` to resemble the html input `autofocus` but you're right that this behaves differently. 

I'm fine with either a) changing the prop name to `focused` or b) changing the implementation so that the `autoFocus` prop is only used to seed the `isEditing` state, and is never used again. 

I think I'll do **b** because it's functionality we know we need now, without speculation about what we might need in the future.

mshwery: Agreed!

PR #482, opened 2019-01-24
Add missing letterSpacing prop to 200 heading

zeroslope:
This PR Fixes _#479 Heading 200 doesn't have Letter spacing_

This is my first PR. I finished this PR in the following order.

1. fork it
2. create a new branch named "fix-heading-200-style"
3. change the heading 200 theme
4. run the doc to see if it is correct

I think the missing value is letterSpacing: '0'. It is different from the heading 300 just in color.

### before change
image

### after change
image

### DIFF
--- a/src/theme/src/default-theme/typography/headings.js
+++ b/src/theme/src/default-theme/typography/headings.js
@@ -81,6 +81,7 @@ export default {
     fontSize: '12px',
     fontWeight: 600,
     lineHeight: '16px',
+    letterSpacing: '0',
     marginTop: 16,
     fontFamily: fontFamilies.ui,
     color: colors.text.muted


src/theme/src/default-theme/typography/headings.js
```diff
@@ -81,6 +81,7 @@ export default {
     fontSize: '12px',
     fontWeight: 600,
     lineHeight: '16px',
+    letterSpacing: '0',
```
mshwery: Actually, I wonder if we want this to be `normal` (the default value of this property). 

> The normal letter spacing for the current font. Unlike a value of `0`, this keyword allows the user agent to alter the space between characters in order to justify text.

https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing

vadimdemedes: I would also use `'0'`, since size `300` also has it: https://github.com/segmentio/evergreen/blob/ec4552dee7561178f0a0291dd3cd34040a5a578d/src/theme/src/default-theme/typography/headings.js#L75.

mshwery: Fair enough!

vadimdemedes:
Thanks @zeroslope!

PR #491, opened 2019-02-05
Register first keydown on the editableCell component

hellooimkat:
When using the `editableCell`, on the first keydown the key value isn’t being set to state, and so requires 2 key presses to register the key value:
master

Functionality changes made in this PR:
- The first keydown value is registered, value is appended to cell
- Capital letters are registered on keydown

feb-11-2019 13-38-12

### DIFF
--- a/src/table/src/EditableCell.js
+++ b/src/table/src/EditableCell.js
@@ -60,8 +60,13 @@ class EditableCell extends React.PureComponent {
 
   static getDerivedStateFromProps(props, state) {
     if (props.children !== state.value) {
+      if (state.value !== '') {
+        return {
+          value: props.children
+        }
+      }
       return {
-        value: props.children
+        value: state.value
       }
     }
     return null
--- a/src/table/src/EditableCell.js
+++ b/src/table/src/EditableCell.js
@@ -59,14 +59,15 @@ class EditableCell extends React.PureComponent {
   }
 
   static getDerivedStateFromProps(props, state) {
+    console.log(props.children)
     if (props.children !== state.value) {
-      if (state.value !== '') {
+      if (state.value !== '' && props.children === '') {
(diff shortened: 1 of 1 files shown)


mshwery:
Ah, I see the problem is that you want editable cells to become _active_ when tab focusing them? (the same as clicking into them?)

hellooimkat:
Talking with @mshwery on Slack, there was a potential race condition/delay between the value coming through as `props.children` or `state.value`.

hellooimkat:
@mshwery refactored!

src/table/src/EditableCell.js
```diff
@@ -96,30 +87,32 @@ class EditableCell extends React.PureComponent {
      * When the user presses a character on the keyboard, use that character
      * as the value in the text field.
      */
-    if (key.match(/^[a-z]{0,10}$/) && !e.metaKey && !e.ctrlKey && !e.altKey) {
-      this.setState({
-        isEditing: true,
-        value: key
-      })
-    } else if (key === 'Enter') {
+    if (key === 'Enter' || key === 'Shift' || this.state.value !== '') {
```
mshwery: Why add this `key === 'Shift' || this.state.value !== ''`?

hellooimkat: `this.state.value`: The functionality that I was trying to preserve was on keydown, if there was already text in the cell, instead of updating the cell immediately, to enter editing mode. 
`Shift`: Currently, the cells only take in a-z. This lets the user also input A-Z (changing the regex to [a-zA-z] automatically enters the word 'Shift' in the field). I thought this would be a nice feature, but I can take it off.

mshwery: Do we have a storybook or example in the docs for this?

PR #493, opened 2019-02-06
Fix Button whitespace so text does not wrap

tj:
I'm not sure if there are any current use-cases where you'd want wrapping but figured I'd send over a PR if it was an oversight! Easy enough to add in any case.

screen shot 2019-02-06 at 6 43 28 pm


tj:
oops tests ahah.. right, I'll just add it inline :D

PR #494, opened 2019-02-08
Allow SelectMenuContent to have editable placeholder icon & value

hellooimkat:
Previously, the `SelectMenuContent` component had a placeholder that could not be accessed through its props This change proposes to allow the filter placeholder to be editable through props. 

This feature is also added to the Storybook so users can actively edit the new props being passed in.

**Default values are still the same:**
screen shot 2019-02-07 at 5 07 39 pm

**With the placeholder text changed:**
screen shot 2019-02-07 at 5 07 55 pm

**With the placeholder icon gone:**
screen shot 2019-02-07 at 5 08 08 pm

**With the placeholder icon changed:**
screen shot 2019-02-07 at 5 08 28 pm

**Documentation:**
screen shot 2019-02-12 at 2 15 04 pm
screen shot 2019-02-12 at 2 20 00 pm

### DIFF
--- a/src/select-menu/src/OptionsList.js
+++ b/src/select-menu/src/OptionsList.js
@@ -43,7 +43,8 @@ export default class OptionsList extends PureComponent {
     hasFilter: PropTypes.bool,
     optionSize: PropTypes.number,
     renderItem: PropTypes.func,
-    placeholder: PropTypes.string,
+    filterPlaceholder: PropTypes.string,
+    filterIcon: PropTypes.string,
     optionsFilter: PropTypes.func,
     defaultSearchValue: PropTypes.string
   }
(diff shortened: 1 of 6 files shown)


src/select-menu/src/OptionsList.js
```diff
@@ -43,7 +43,8 @@ export default class OptionsList extends PureComponent {
     hasFilter: PropTypes.bool,
     optionSize: PropTypes.number,
     renderItem: PropTypes.func,
-    placeholder: PropTypes.string,
```
mshwery: Looks like we are still using `placeholder` in some props destructuring.

src/table/src/SearchTableHeaderCell.js
```diff
@@ -56,7 +56,8 @@ export default class SearchTableHeaderCell extends PureComponent {
   static defaultProps = {
     onChange: () => {},
     spellCheck: true,
-    placeholder: 'Filter...'
+    placeholder: 'Filter...',
+    icon: 'search'
```
mshwery: We should document this change.

hellooimkat:
@mshwery documentation updated! Screenshots in place.

Issue #215, opened 2018-07-10
Combobox with itemToString prop does NOT autocomplete as I type


emadabdulrahim:
I noticed when I use `<Combobox itemToString/>` such as in the last example provided in the docs

```jsx
<Combobox
  defaultSelectedItem={{ label: 'Banana' }}
  items={[{ label: 'Banana' }, { label: 'Pear' }, { label: 'Kiwi' }]}
  itemToString={item => item.label}
  onChange={selected => console.log(selected)}
/>
```

Autocomplete doesn't work in that case.

_P.S._
Also, in that same example in the docs if you try to backspace all the way to remove the selected item, you'll notice this error message
`TypeError: Cannot read property 'label' of null`

I'm guessing that's because `onSelects` fires and sends an empty object. I wonder if there is a better way to implement this without having the consumer of the component to code defensively.

timtutt:
In case anyone comes across this again, after about 2 hrs of digging through code I just figured out the issue and a solution...

By default the underlying AutoComplete Component uses `fuzzaldrin` for the filtering by default and passes items to the `fuzzaldrin.filter` function without the option to handle objects.

Was able to work around this by importing `fuzzaldrin` into my project using:
 ```bash
npm install --save fuzzaldrin-plus
```

And then creating my own fuzzy filter that takes advantage of the "key" option in `fuzzaldrin.filter` for filtering of objects:
```jsx
const fuzzyFilter = (items, input) => fuzzaldrin.filter(items, input, {key: "label"});
```

And passing that as an `autocompleteProp`:
```jsx
<Combobox
  defaultSelectedItem={{ label: 'Banana' }}
  items={[{ label: 'Banana' }, { label: 'Pear' }, { label: 'Kiwi' }]}
  itemToString={item => (item != null ? item.label : "")}
  onChange={selected => console.log(selected)}
  autocompleteProps={{    // Used for the title in the autocomplete.
     title: 'Dataset',
     itemsFilter: fuzzyFilter
  }}
/>
```

Also note the error checking to handle cases where there is no item returned if you backspace all the way out.

PR #495, opened 2019-02-09
Fix Combobox filtering with custom itemToString

colinking:
Fixes: #215

If you supply an `itemToString` to an `Autocomplete` / `Combobox`, we will now correctly fuzzy filter on the string values.

| Before | After |
| ------ | ----- |
| image | image |


package.json
```diff
@@ -24,7 +24,7 @@
   "sideEffects": false,
   "scripts": {
     "test": "xo && ava",
-    "prepublishOnly": "rm -rf esm commonjs && yarn run build",
+    "prepare": "rm -rf esm commonjs && yarn run build",
```
Rowno: Why did you change this? `prepublishOnly` is an npm lifecycle hook that makes sure evergreen is rebuilt when people publish.

colinking: Yep, so is `prepare` :) https://docs.npmjs.com/misc/scripts#description

I wish `npm` had a separate object for lifecycle hooks (`lifecycle: { prepare: ... }`), so that this is less so black magic 😬 

But from the docs (`prepare` is a superset of `prepublishOnly`):

```
prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.
prepublishOnly: Run BEFORE the package is prepared and packed, ONLY on npm publish. (See below.)
```

Admittedly, I forgot this was still here -- I was mainly using it for testing in another repo. But in general, this hook builds `evergreen` when you add it as a dependency, which allows you to reference a git branch (f.e. `"evergreen-ui": "segmentio/evergreen#colin/item-to-string-autocomplete",`). It's pretty useful for testing a branch like this out in a production setting.

However, not really sure if we want to include this change since it may slow down install times.

colinking: Reverting this change and dismissing the review so I can ship this :)

colinking:
Merged master + double-checked that all storybook entries for `Autocomplete` and `Combobox` worked.

PR #499, opened 2019-02-17
fixes #450

### LINKED ISSUE #450: Combobox does not scroll to selected item
Steps to reproduce:

bug

jongsue:
fixes #450
highlightedIndex from downshift seems to make this bug.
I've changed to get highlightedIndex from our props.

combobox-scroll

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -198,6 +198,7 @@ export default class Autocomplete extends PureComponent {
           getItemProps,
           selectedItem,
           selectItemAtIndex,
+          highlightedIndex,
           getRootProps,
           ...restDownshiftProps
         }) => (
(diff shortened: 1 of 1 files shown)


src/autocomplete/src/Autocomplete.js
```diff
@@ -237,7 +238,9 @@ export default class Autocomplete extends PureComponent {
                   },
                   inputValue,
                   selectedItem,
-                  highlightedIndex,
+                  highlightedIndex: this.props.items.includes(selectedItem)
+                    ? this.props.items.indexOf(selectedItem)
+                    : 0,
                   selectItemAtIndex,
                   ...restDownshiftProps
```
mshwery: By removing it from the destructure at the top it is being passed in `...restDownshiftProps`. Your code change for `children` here will be ignored because it is being overwritten by the value later.

jongsue: Oops, you're right! 
I'll add highlightedIndex to the destructure

PR #501, opened 2019-02-18
Adding the ability to pass a nonce attribute to the hydration script

### LINKED ISSUE #500: Not able to hydrate under a strict CSP
It appears that Evergreen throws an error and cannot properly run if a site enforces a strict [Content Security Policy](https://www.html5rocks.com/en/tutorials/security/content-security-policy/), specifically if the site disallows `'unsafe-inline'` for the `script-src` directive. The JSON inside the `#evergreen-hydrate` element cannot be accessed if the CSP directive is in place: 

```
Refused to execute a script because its hash, its nonce, or 'unsafe-inline' does not appear in the script-src directive of the Content Security Policy.
```

`'unsafe-inline'` is important when CSP is in place to prevent unwarranted inline scripts from running. Inline scripts are still allowed if the nonce passed from the server matches the `nonce` attribute on the inline script tag. 

To mitigate this, it should be possible to pass a nonce to the hydration script that populates a `nonce` attribute.

jasonbarry:
Fixes #500. This change allows you to pass a nonce to the hydration script tag in order to allow inline scripts to run under a CSP that has the `script-src` directive enabled. The following example shows how to use this change in a Next.js project:

```jsx
export default class extends Document {
  static getInitialProps({ renderPage, res }) {
    const page = renderPage()
    const { nonce } = res.locals
    // Pass the nonce to `extractStyles`
    const { css, hydrationScript } = extractStyles({ nonce })
    return { ...page, css, hydrationScript, nonce }
  }

  render() {
    const { css, hydrationScript, nonce } = this.props
    return (
      <html lang="en">
        <Head>
          <style dangerouslySetInnerHTML={{ __html: css }} />
        </Head>
        <body>
          <Main />
          {hydrationScript}
          <NextScript nonce={nonce} />
        </body>
      </html>
    )
  }
}
```

(See also [this example](https://github.com/zeit/next.js/tree/master/examples/with-strict-csp) for more context)

Note that this implementation is pretty loose — you could pass any object to `extractStyles` and it would get spread to the hydration script tag. I'm open to ideas on how to make this more strict while keeping it clean and backwards-compatible.


mshwery:
Would it make more sense to explicitly permit an optional `nonce` value that gets set as the `nonce` prop instead? Rather than allowing an anything-goes object?

```js
console.log(extractStyles({ nonce: 'whatever' }))
// {
//   type: 'application/json',
//   id: 'evergreen-hydrate',
//   dangerouslySetInnerHTML: { __html: JSON.stringify(evergreenCache) },
//   nonce: 'whatever'
// }
```

I think the biggest risk is that we are letting consumers pass in their own `dangerouslySetInnerHTML`, `id` and `type` that would override our values.

jasonbarry:
@mshwery what about if the `scriptAttributes` spread was first? That way, `type`, `id`, and `dangerouslySetInnerHTML` couldn't be overwritten. 

```js
const scriptProps = {
  ...scriptAttributes,
  type: 'application/json',
  id: 'evergreen-hydrate',
  dangerouslySetInnerHTML: { __html: JSON.stringify(evergreenCache) }
}
```

This approach is less explicit, but more future-proof – if it was desired to add another attribute in the future, all that would need to change would be a call-out in the docs.

vadimdemedes:
I agree with @mshwery, we should expose one prop that takes care of that functionality, without letting developers overwrite any prop on the script tag. Exposing specific props for specific bits of functionality is better long term and is easier to maintain, because we know exactly what input to expect.

jasonbarry:
@vadimdemedes @mshwery makes sense to me. I've updated the branch.

src/ssr/src/extractStyles.js
```diff
@@ -16,6 +16,7 @@ export default function extractStyles() {
   const scriptProps = {
     type: 'application/json',
     id: 'evergreen-hydrate',
+    nonce: options.nonce,
```
mshwery: Actually one last thing – can we only apply `nonce` when it is provided?

jasonbarry: @mshwery np, done

Issue #504, opened 2019-02-23
SelectMenu onDeselect 


rejifald:
Hi all. I am newbie in React and in evergreen too.
I use `4.12.1` evergreen version. MacOS, Chrome 72. Haven't published this project yet.
But i believe there is a bug with SelectMenu:onDeselect event.
My event handler isn't invoking. 
onSelect works ok, but onDeselect isn't. BTW, when i open menu at the first time, i see get this error in console:
![image](https://user-images.githubusercontent.com/42000283/53287982-be0cc300-378b-11e9-8dd6-286dafbae2aa.png)
Here is how i use it:
```
        <SelectMenu
          title={'Choose level'}
          selected={this.state.filters.levels}
          onSelect={item => this.onLevelSelect(item.value)}
          onDeselect={item => this.onLevelDeselect(item.value)}
          hasFilter={false}
          options={[
            'error',
            'warn',
            'info',
            'verbose',
            'debug',
            'silly'
          ].map(label => ({ label, value: label }))}>
          <Button>Select level</Button>
        </SelectMenu>
```

mshwery:
The proptype warning is a regression – nice find. As for the `onDeselect` behavior it's really hard to help without a link to a minimal reproduction of the issue in code @rejifald

rejifald:
@mshwery NP ! Here is repo with my experimentals. https://github.com/rejifald/k just clone it, and run `npm start`. 
And open `localhost:3000`. In top lever corner u will see `Select level` button. This is SelectMenu trigger.
Hope, it will help.

mshwery:
The example in the docs works. Have you tried the `isMultiSelect` prop?

rejifald:
Yes. I've checked docs, and i know that there it works.
Hm... I've tried... But i think it wasn't work. But now i see that it works... Thanks :)
But why i can't deselect item if **isMultiSelect** is false ?
And, i still see same error.

mshwery:
That is a good question. It might be by design (although I'm not sure I agree with that design now).

I think it behaves similar to a native browser `<select>` in that once you've selected something you cannot deselect it unless you have `multiselect` set. You can only select a _different item_ but not deselect the currently selected item.

rejifald:
If i want to let user ability deselect item, i can't anyway, right ?
It forces me to implement additional buttons, etc.
Combobox also disallow this. For example 
[...]

PR #505, opened 2019-02-25
fix proptype warning and rm package lock because yarn

mshwery:
This resolves the proptype warning from https://github.com/segmentio/evergreen/issues/504

I also removed the `package-lock.json` which got added at some point (unsure if intentional)


vadimdemedes:
If you're removing `package-lock.json` in this PR anyway, could you also add `.npmrc` with `package-lock=false` content to prevent anyone from ever creating package-lock.json again?

mshwery:
@vadimdemedes excellent, will do.

PR #506, opened 2019-02-25
Task/use np

mshwery:
This PR reintroduces [`np`](https://github.com/sindresorhus/np) for publishing releases and deploys the docsite alongside version releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "build": "concurrently --names 'commonjs,esm' 'yarn run build-commonjs' 'yarn run build-esm'",
     "clean": "git clean -Xdf",
     "release": "np",
-    "build-docs": "cd docs && yarn install && yarn deploy",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -33,7 +33,8 @@
     "build-esm": "BABEL_ENV=esm babel src --out-dir esm --ignore '**/stories','**/test' --source-maps inline",
     "build": "concurrently --names 'commonjs,esm' 'yarn run build-commonjs' 'yarn run build-esm'",
     "clean": "git clean -Xdf",
-    "release": "np --no-publish",
+    "release": "np",
+    "build-docs": "cd docs && yarn install && yarn deploy",
```
vadimdemedes: Should it be `deploy-docs` or `release-docs` (consistent with `release`) instead? Asking since it says `build`, but it actually deploys them.

mshwery: Ah, good call. That's much better!

PR #509, opened 2019-03-01
Add Startup banner to home page

williamgrosset:
## Overview
This PR introduces a banner at the top of the docs to the Startup Program (launched last week). The motivation for this was because of 🌲Evergreen's amazing SEO 🚀- suggested by @nettofarah.

## Screenshot
screen shot 2019-02-28 at 11 11 54 pm


vadimdemedes:
Could you put that banner at the very top of the page and make it closable? I don’t think it fits well between nav and the header, kind of interrupts the design of the page.

williamgrosset:
banner

Hiding banner on screen sizes > 1440 px since the image doesn't scale well.

vadimdemedes:
Could you code that banner instead of image btw? It's a bit odd to have it as an image anyway :) It would scale perfectly then and wouldn't be distorted.

Issue #497, opened 2019-02-12
Popover overflow


dy:
I have react-select inside of popover and it gets cropped:

![image](https://user-images.githubusercontent.com/300067/52657047-bf093f00-2ec5-11e9-9c80-6cddf5a63390.png)

What is the correct way to set `overflow="visible"` for the popover?

dy:
Ok, can be resolved as
```jsx
<Popover statelessProps={{ overflow: 'visible'}}/>
```
fixed by #511

PR #511, opened 2019-03-06
Fix `statelessProps` propTypes

### LINKED ISSUE #510: Failed prop type: Property `statelessProps` of component `Popover` has invalid PropType notation inside objectOf.
We have `statelessProps` property for Popover component https://evergreen.segment.com/components/popover,
but passing anything to that raises an error
```sh
Warning: Failed prop type: Property `statelessProps` of component `Popover` has invalid PropType notation inside objectOf.
```
The reason for that is that _prop-types_ expects argument for `objectOf` to be a function, not an object 
https://github.com/segmentio/evergreen/blob/master/src/popover/src/Popover.js#L60.

That is just logically wrong statement there - `objectOf` checks that an object passed to properties contains only properties of specific type. Instead, `PropType.shape` should be used.

dy:
This fixes #510 and #497, enabling passing properties to the popover container.

I wonder though, why `statelessProps` property has different name than Dialog's `containerProps`. The `containerProps` seems to be more natural/intuitive.

@rywall @mshwery @Flaque

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -57,7 +57,7 @@ export default class Popover extends Component {
     /**
      * Properties passed through to the Popover card.
      */
-    statelessProps: PropTypes.objectOf(PopoverStateless.propTypes),
+    statelessProps: PropTypes.shape(PopoverStateless.propTypes),
 
     /**
      * Duration of the animation.


src/popover/src/Popover.js
```diff
@@ -57,7 +57,7 @@ export default class Popover extends Component {
     /**
      * Properties passed through to the Popover card.
      */
-    statelessProps: PropTypes.objectOf(PopoverStateless.propTypes),
+    statelessProps: PropTypes.shape(PopoverStateless.propTypes),
```
mshwery: 👍 Nice catch

dy:
@mshwery thanks!
Any estimate on when can we expect the patch?

colinking:
Hey @dy! I just released it in [v4.13.1](https://github.com/segmentio/evergreen/releases/tag/v4.13.1) :)

PR #513, opened 2019-03-10
Add TextInput isClearable + onClear

JMFURY:
@mshwery -- Fresh branch stemming from the process over on #256 

For now I implemented the **TextInput** parts (state handling and storybook examples) here on this branch. I think this could use a once over by you from a style/design perspective. It's been a bit since I implemented it the first time around, and I wonder if there are any ideas you might have as to how to make it cleaner. 

Questions: 

- Are we still wanting to support both controlled and uncontrolled in the same component?
- Am I missing something in this implementation that could make it simpler to distinguish between these two use-cases? 

Regarding the second bullet... for example, should we maybe do an `if` in the render method based off whether or not `isClearable` is even present? That would give the opportunity to leave the non-clearable TextInput completely untouched.


src/text-input/src/TextInput.js
```diff
@@ -89,25 +118,55 @@ class TextInput extends PureComponent {
...
+          spellCheck={spellCheck}
+          aria-invalid={isInvalid}
+          {...(disabled ? { color: 'muted' } : {})}
+          css={css}
+          {...props}
+        />
+        {isClearable &&
+          (!this.props.value &&
+            this.state.value.length > 0 &&
+            !this.props.onChange) && (
```
mshwery: Small suggestion: I would pull all this conditional logic out into a `shouldShowClearIcon` or similar. That way we don't have a bunch of chained conditions in the JSX body.

src/text-input/src/TextInput.js
```diff
@@ -89,25 +118,55 @@ class TextInput extends PureComponent {
...
+              icon="cross"
+              appearance="default"
+              onClick={() => {
+                this.setState({ value: '' })
+              }}
+              marginLeft={-20}
+            />
+          )}
+        {isClearable &&
+          (this.props.value && this.props.onChange) && (
```
mshwery: Is there a good reason to have two separate `isClearable` behaviors?

src/text-input/src/TextInput.js
```diff
@@ -59,6 +70,14 @@ class TextInput extends PureComponent {
     className: PropTypes.string
   }
 
+  constructor(props) {
+    super(props)
+    // Create a state for 'uncontrolled' inputs, locally controlled
+    this.state = {
+      value: this.props.defaultValue || ''
```
mshwery: Rather we skip the constructor and do this: 

```js
class TextInput extends PureComponent {
  state = {
    value: this.props.defaultValue || ''
  }
}
```

src/text-input/src/TextInput.js
```diff
@@ -68,14 +87,24 @@ class TextInput extends PureComponent {
     spellCheck: true
   }
 
+  getValue() {
+    if ('value' in this.props) {
+      // Console.log('Input has controlled value prop')
```
mshwery: Could you remove this bit? :)

src/text-input/src/TextInput.js
```diff
@@ -68,14 +87,24 @@ class TextInput extends PureComponent {
...
+      return this.props.value
+    }
+    return this.state.value
```
mshwery: To simplify things like state vs props we might benefit from making two separate (both exported) classes. One `*Stateless` that's fully controlled, and another that's stateful. That way consumers can decide which one to use. What do you think?

I'm happy to provide a bit more feedback on this if you need a little more suggestion here.

src/text-input/src/TextInput.js
```diff
@@ -89,25 +118,55 @@ class TextInput extends PureComponent {
...
+          size={textSize}
+          width={width}
+          height={height}
+          required={required}
+          disabled={disabled}
+          placeholder={placeholder}
+          value={this.getValue()}
+          {...(onChange
+            ? { onChange }
+            : { onChange: e => this.setState({ value: e.target.value }) })}
```
mshwery: I'd recommend that we always call our own function, and only if the consumer also passes an `onChange` prop do we bubble to them as well (you can use `safeInvoke` for this):

```js
class TextInput extends PureComponent {
  handleChange = (event) => {
    this.setState({ value: e.target.value })
    safeInvoke(this.props.onChange, event)
  }

  render() {
    // later...
    <Text
      onChange={this.handleChange}
    />
  }
}
```

mshwery:
@JMFURY also, something else to note: I think some browsers automatically display a sort of "clear" icon in text inputs natively (like Internet Explorer and Edge). We'll either need to hide the native clear buttons if we are showing our own, OR defer to the native ones when available.

I think it'll end up being easier to show our own and hide the native ones, that way we'll get more uniform display and behavior across browsers.

src/text-input/src/TextInput.js
JMFURY: I love this idea, appreciate the guidance. I'll see what I can put together!

src/text-input/src/TextInput.js
JMFURY: Probably a naive reason way back when, but I think I was just confusing myself before. I think I know how to collapse now.

src/text-input/src/TextInput.js
mshwery: I should also follow up that we could use the `*Stateless` version _within_ the stateful one.

JMFURY: Can you give me a clue on the file structure this might impact? Where do you want each component ... Exact filenames and exports in some fashion I can follow would be appreciated.

JMFURY: I have done some legwork and have it working just would rather ask ahead of pushing

mshwery: I imagine the stateful component (`TextInput`) would look something like this. Note: I'm excluding a lot of the bits that you've added, but anything that's managing state would belong in this component instead of `StatelessTextInput`. Atlaskit actually does this quite a bit, iirc.

```jsx
import React, { PureComponent } from 'react'
import StatelessTextInput from './StatelessTextInput'

class TextInput extends PureComponent {
  static propTypes = {
    /**
     * Composes the StatelessTextInput component as the base.
     */
    ...StatelessTextInput.propTypes
  }

  state = {
    value: this.props.value
  }

  handleChange = (event) => {
    // handle state change and bubble event to `this.props.onChange` if provided?
  }

  render() {
    return (
      <TextStatelessTextInput
        {...this.props}
        value={this.state.value}
        onChange={this.handleChange}
      />
    )
  }
}

export default TextInput
```

Sidenote: I'm also beginning to wonder if this behavior will interfere with native browser support for clearing inputs (IE / Edge)?

JMFURY: Okay great. This is generally what I have already, and your naming preferences help a ton! 

ON the IE / Edge *edge* case, it seems to be the accepted practice to just `display:none` on it. 
https://stackoverflow.com/questions/14007655/remove-ie10s-clear-field-x-button-on-certain-inputs
https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-clear

I was thinking of just adding a style to the components with a prop. Do you think it should have a true/false that can be set by consumer of the component? ..or is it good enough to simply document that we override it completely since it is non-standard?

mshwery: It's probably fine for us to override it with a cross-browser version only if they've opted in to using `isClearable`, where `isClearable` defaults to `false`. That way we don't introduce a breaking change for consumers today, and if they prefer the native non-cross-browser behavior they can continue to us it.

src/text-input/src/StatelessTextInput.js
```diff
@@ -0,0 +1,153 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+import { css } from 'glamor'
+import { Text } from '../../typography'
+import { withTheme } from '../../theme'
+import { Icon } from '../../icon'
+
+// Create a style to override default IE, Edge native input=["text"] clear button
+const hideMsClearPseudoElement = css({
```
JMFURY: This is not tested yet.

src/text-input/src/StatelessTextInput.js
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Makes the input element clearable; enables clickable [X] icon
+     */
+    isClearable: PropTypes.bool,
+
+    /**
+     * Used to trigger the function that should clear the input value in Parent component.
+     * Note that this is required to leverage the click of the [X] icon.
+     */
+    onClear: PropTypes.func,
```
JMFURY: I took a stab at these messages on Proptypes, but they're probably not perfect

src/text-input/src/TextInput.js
```diff
@@ -118,55 +106,28 @@ class TextInput extends PureComponent {
...
-              marginLeft={-20}
-            />
-          )}
-      </React.Fragment>
+      <StatelessTextInput
+        value={this.state.value}
+        isClearable={isClearable}
+        onClear={() => this.setState({ value: '' })}
+        onChange={e => this.setState({ value: e.target.value })}
+        className={cx(themedClassName, className)}
```
JMFURY: I am all for consolidating these props into the ...restProps type reference, but again was not positive on the usage of the Stateless version and how this might look to consumers. 

I often inspect a render method when I'm looking at a library and while the spread might be cleaner it is less explicit that this component I'm consuming does XYZ with all these props. 

Happy to clean this up though.

src/text-input/src/TextInput.js
```diff
@@ -89,8 +106,11 @@ class TextInput extends PureComponent {
     const borderRadius = theme.getBorderRadiusForControlHeight(height)
 
     return (
-      <Text
-        is="input"
+      <StatelessTextInput
+        value={this.state.value}
+        isClearable={isClearable}
+        onClear={() => this.setState({ value: '' })}
+        onChange={e => this.setState({ value: e.target.value })}
```
mshwery: Rebinding new functions on every render will defeat the purpose of using PureComponent. I would extract these to bound functions:

```js
class Foo {
  myFunction = (event) => {}
}
```

JMFURY: My bad! -- fixed this bit (I think)

PR #514, opened 2019-03-11
add stalebot config

mshwery:
This adds stalebot to automatically handle closing out stale PRs/Issues.


vadimdemedes:
@mshwery Could you apply this only to PRs? This bot is going to close a lot of useful issues for no good reason.

mshwery:
Fair point. I wonder if we could enforce this on issues so long as they are being triaged. Basically ignore any issues that have our specific triage labels applied. // @vadimdemedes

PR #522, opened 2019-03-18
Revert "Scroll combobox to highlighted item (#499)"

tusbar:
This reverts commit 38be9cc6e95342c57ea7803f666f8a13f33ce441 from #499.

highlightedIndex is a controlled prop, you cannot force it to be `0`.

This will revert the intended feature by @jongsue (scrolling to the selected item when opening the combo box), but this fixes all the broken highlights (and keyboard navigation) for all components depending on downshift.

I will open a PR that upgrades downshift and fixes this.

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -212,8 +212,8 @@ export default class Autocomplete extends PureComponent {
           inputValue,
           getItemProps,
           selectedItem,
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Good catch!

vadimdemedes:
@mshwery you good with reverting this? It makes sense to me.

mshwery:
@vadimdemedes sgtm. go for it!

PR #523, opened 2019-03-18
Upgrade downshift to 3.2.10 & highlight autocomplete’s selected item by default

### LINKED ISSUE #450: Combobox does not scroll to selected item

tusbar:
This fixes #450 by creating a `stateReducer` that will set the `highlightedItem` to the selected item when opening the autocomplete – as suggested by @kentcdodds here: https://github.com/downshift-js/downshift/issues/645#issuecomment-451805753.

I wish this could be done without upgrading downshift but it doesn’t seem to work properly with the version evergreen is using (it highlights and scrolls to the selected element by default in Autocomplete but not in Combobox)

So this is breaking because Downshift renamed the following properties:

| Previous name | New name | |
|-|-|-|
| `getButtonProps` | `getToggleButtonProps` | _not used directly_ |
| `defaultSelectedItem` | `initialSelectedItem` | |
| `defaultInputValue` | `initialInputValue` | _unused_ |

If you don’t want to do a major release, we could alias and deprecated those properties first.

### DIFF
--- a/docs/src/pages/components/autocomplete.mdx
+++ b/docs/src/pages/components/autocomplete.mdx
@@ -99,7 +99,7 @@ Get the `openMenu` prop to manually open the menu on certain events such as `onF
 >
   {({
     getInputProps,
-    getButtonProps,
+    getToggleButtonProps,
     getRef,
     inputValue,
     toggleMenu
@@ -110,7 +110,7 @@ Get the `openMenu` prop to manually open the menu on certain events such as `onF
         value={inputValue}
         {...getInputProps()}
       />
-      <Button onClick={toggleMenu} {...getButtonProps()}>
+      <Button onClick={toggleMenu} {...getToggleButtonProps()}>
         Trigger
       </Button>
     </Pane>
@@ -133,7 +133,7 @@ Consider using that component before using the `Autocomplete` component directly
   {({
     key,
     getInputProps,
-    getButtonProps,
+    getToggleButtonProps,
     getRef,
     inputValue,
     openMenu,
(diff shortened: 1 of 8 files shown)


tusbar:
Rebased!

tusbar:
I added a deprecation message when using `defaultSelectedItem` and aliased to prop to `initialSelectedItem`.

What do you think. It would be nice to have this merged.

tusbar:
Upgraded downshift to 3.2.10.

Could we merge this @mshwery?

mshwery:
I think we'd have to merge this into the v5 branch (which already upgraded downshift iirc) because this is a breaking change.

tusbar:
@mshwery is the v5 branch still active?

The old version of downshift is annoying. `Autocomplete` doesn’t work on mobile, etc.
Should I reopen this and make it non breaking (by aliasing the 3 changing props)?

mshwery:
@tusbar i'd be in favor of getting this out the door before v5 if you think the aliasing approach works for now. We can make those properties deprecated while also supporting the new prop names?

Curious if there are any other "breakages" besides those prop names (i.e. behavior/functionality?)

tusbar:
The list of breaking changes is rather small:

- https://github.com/downshift-js/downshift/releases/tag/v3.0.0
  - All the `default*` props were renamed to `initial*`
  - Keyboard accessibility fixes (first item is selected when pressing down and last item is selected when pressing up)
- https://github.com/downshift-js/downshift/releases/tag/v2.0.0
  - `getButtonProps` renamed to `getToggleButtonProps`
  - `render` prop not supported anymore – but only the `children` prop is documented in evergreen’s doc.


Other than that, there were a lot of bug and accessibility fixes.

mshwery:
Hm. The safest thing would be a major release so we don't break anyone. But I wouldn't be opposed to introducing the aliasing prop names.

@vadimdemedes @Rowno @solon-aguiar thoughts?

tusbar:
In the meantime, https://github.com/segmentio/evergreen/pull/631. :)

Issue #532, opened 2019-03-26
FilePicker redux-form integration


dy:
Redux-form expects `onChange: (event, value?) => void` event handler to be defined on an input wrapped with Field (https://redux-form.com/6.6.3/docs/api/field.md/).

That works perfectly with input type="file":
```js
<Field name="file" component={({field}) => (
<input {...field.input} type="file"/>
)}/>
// selecting a file we see data in redux-form
```

But when we use FilePicker, we get nothing:
```js
<Field name="file" component={({field}) => (
<FilePicker {...field.input}/>
)}/>
// selecting a file we see nothing in form data
```

That is because `onChange` event handler is modified by evergreen-ui to return list of files instead of event.

Ideally onChange event should not modify arguments, following the consistency with other inputs. Or at least it should not hide native events to be able to customize that. Otherwise it takes over the part of business logic, not only the UI.

dy:
Workaround for now:

<Field name="file" component={({props}) => <Box onBlur={e => e.stopPropagation()}>
    <FilePicker
        name={props.input.name}
        onChange={props.input.onChange}
        onFocus={props.input.onFocus}
        height={40} />
</Box>}/>

PR #533, opened 2019-03-27
Filepicker onblur event

dy:
This fixes https://github.com/segmentio/evergreen/issues/532 and adds proper onBlur handling
@mshwery


src/file-picker/src/FilePicker.js
```diff
@@ -49,7 +49,12 @@ export default class FilePicker extends PureComponent {
     /**
      * Function called when onChange is fired
      */
-    onChange: PropTypes.func
+    onChange: PropTypes.func,
+
+    /**
+     * Function called when onBlur is fired
+     */
+    onBlur: PropTypes.func
```
dy: If we don't add onBlur handler here, there is no easy way to attach [redux-form field properties](https://redux-form.com/6.6.3/docs/api/field.md/) to the filepicker

src/file-picker/src/FilePicker.js
```diff
@@ -157,4 +164,17 @@ export default class FilePicker extends PureComponent {
...
     this.fileInput.click()
   }
+
+  handleBlur = e => {
+    const { onBlur } = this.props
+    const { files } = this.state
+
+    // Redux-form derives field value from bubbled onBlur event, which is fake here
+    if (e && e.stopPropagation) e.stopPropagation()
+    if (this.fileInput) this.fileInput.blur()
```
dy: Checks are needed only for tests, otherwise we'd have to introduce jsdom in tests.

src/file-picker/src/FilePicker.js
```diff
@@ -157,4 +164,17 @@ export default class FilePicker extends PureComponent {
...
+
+    if (onBlur) {
+      onBlur(files)
```
dy: First argument in onBlur is the same as in onChange - requirement of redux-form, if event with data is not available.

mshwery: We do have a `safeInvoke` function we could use instead of this `if` check.

src/file-picker/src/FilePicker.js
```diff
@@ -157,4 +164,17 @@ export default class FilePicker extends PureComponent {
   handleButtonClick = () => {
     this.fileInput.click()
   }
+
+  handleBlur = e => {
+    const { onBlur } = this.props
+    const { files } = this.state
+
+    // Redux-form derives field value from bubbled onBlur event, which is fake here
+    if (e && e.stopPropagation) e.stopPropagation()
```
mshwery: I'm concerned this change might be a breaking one.

Why must we stop the event from propagating?

src/file-picker/src/FilePicker.js
mshwery: Prefer new lines and curly braces for consistency.

src/file-picker/src/FilePicker.js
dy: If we don't, the event bubbles up to redux-form Field, which derives field value from `event.target.files`: https://github.com/erikras/redux-form/blob/master/src/events/getValue.js#L32
The problem with that event is that it's fake, not real `<input type="file/>` event.
We could try to augment that with `target.files` property, that'd also fix that.

mshwery: I am wondering why we wouldn't handle that in consumers since it's a dependency of `redux-form`. e.g. wrap your field with `onBlur` handling that does the stop propagation.

Otherwise this could be a breaking change for some consumers and would have to wait for v5.

dy: That works, but that is unknown hack for now.

```js
<Field name="file" component={({props}) => (
<Box onBlur={e => e.stopPropagation()}>
    <FilePicker
        name={props.input.name}
        onChange={props.input.onChange}
        onFocus={props.input.onFocus}
        height={40} />
</Box>)}/>
```

What about augmenting `event.target.files`?

dy:
Ok, no need for the PR then, considering #532 closed.

dy:
@mshwery actually - can we still introduce `onBlur` method - that would allow for easier integration with redux-form, it would be easier to preventDefault too, if needed.

Issue #531, opened 2019-03-24
Placeholder for FilePicker


mehmetnyarar:
Hi,

Could you provide `placeholder` attribute for the `FilePicker` component?

Can be useful in the following situations:

- translation of the placeholder text
- small description of the file requirements inside the input

I don't know why there's such a comment in the source code by the way:

> placeholder: "Select a file to upload\u2026" // There's a weird specifity issue when there's two differently sized inputs on the page

BR,
Mehmet

mehmetnyarar:
Actually, it would be great to customize button text, too :)

PR #536, opened 2019-03-31
Added placeholder and buttonLabels properties for FilePicker

narekmal:
Addresses #531

### DIFF
--- a/src/file-picker/src/FilePicker.js
+++ b/src/file-picker/src/FilePicker.js
@@ -48,6 +48,20 @@ export default class FilePicker extends PureComponent {
     height: PropTypes.number,
 
     /**
+     * The placeholder attribute of the input.
+     */
+    placeholder: PropTypes.string,
+
+    /**
+     * Button labels in different states of the component.
+     */
+    buttonLabels: PropTypes.shape({
+      selectLabel: PropTypes.string,
+      replaceFileLabel: PropTypes.string,
+      replaceFilesLabel: PropTypes.string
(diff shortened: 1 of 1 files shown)


src/file-picker/src/FilePicker.js
```diff
@@ -47,6 +47,20 @@ export default class FilePicker extends PureComponent {
...
 
+    /**
+     * The placeholder attribute of the input.
+     */
+    placeholder: PropTypes.string,
+
+    /**
+     * Button labels in different states of the component.
+     */
+    buttonLabels: PropTypes.shape({
```
mshwery: I think we could make all of these top level props instead and simplify some of the logic below.

mshwery:
Could you add a description to this PR describing the change and why?

narekmal:
PR addresses issue #531  by adding properties to FilePicker component for overriding default input placeholder, select file button label, replace file button label, replace files button label

src/file-picker/src/FilePicker.js
```diff
@@ -108,21 +116,19 @@ export default class FilePicker extends PureComponent {
     let buttonText
     if (files.length === 0) {
       buttonText =
-        buttonLabels === undefined || buttonLabels.selectLabel === undefined
+        selectFileButtonLabel === undefined
           ? 'Select file'
```
mshwery: If you use defaultProps you shouldn't need to do these `undefined` checks which will clean up the code quite a bit.

PR #537, opened 2019-04-07
[WIP] Migrate to typescript

jeroenptrs:
Hi all, sorry in advance for the unfinished PR. I tagged it as WIP but felt like opening it already because I didn't want this work to go to waste.

I was able to setup the bulk of the components with TS definitions already, and just got XO and storybook to make good with typescript (the latter being the final checkpoint I wanted to touch before opening this PR).

There's still a few things to do:
- [x] move css definition to `ui-box`
- [x] add props back in (I deleted them at first but they're critical for non-TS projects, as Evergreen is a component library)
- [ ] migrate all remaining components to TS
- [ ] migrate all stories to TSX
- [x] fix tests (would love help with this, not used to AVA)
- [ ] migrate all tests to Typescript
- [ ] fix build process for both CJS and ESM (would love help with this, I'm guessing 2 different tsconfig that extend from a base tsconfig) as per @jaredpalmer this would easily be solved by using TSDX
- [ ] add husky and lint-staged hooks back in

Hope you all appreciate the help, it's pretty new for me - contributing to a component library - helping out at this scale. But, I really dig Evergreen as a design language! I made the move to TS last year and would love to continue it in all its glory (but with the typescript upgrade!)

If you have any feedback be sure to leave it in this PR, I can update as I go 🙌


jeroenptrs:
Got `yarn test` to run locally, so now I can just migrate everything to Typescript with ease until the final steps.

One note: I took the entire definition of `ui-box` locally since all propTypes were untyped. I still need to migrate the `css` namespace back into ui-box as well (it's taken from glamor at the moment)

src/alert/src/Alert.tsx
```diff
@@ -1,78 +1,69 @@
...
 
-    /**
-     * Function called when the remove button is clicked.
-     */
-    onRemove: PropTypes.func,
+  // When true, show a remove icon button.
+  isRemoveable?: boolean
+
+  // Function called when the remove button is clicked.
+  onRemove?: AnyFunction
```
jaredpalmer:
```suggestion
  onRemove?: () => void
```

Using an alias like `AnyFunction` is a code smell and is annoying to update if you end up passing something later. It is also more explicit.

jeroenptrs: Thanks! I'm updating `any` related typing in my next commit :)

jaredpalmer:
I know that Evergreen and Blueprint share some stuff (like Icons). However, I would beg for Evergreen not to follow Blueprint's TS `interface` naming convention (where everything starts with `IXXX`. This convention is explicitly prohibited in TypeScript core's tslint rules and is losing popularity. Instead, using `ButtonProps` and `ButtonState` for  a `<Button />` is much more conventional nowadays.

**Libraries and UI Kits that do not use `Ixxx`**

- React and React DOM !!!
- [Atlassian AtlasKit](https://bitbucket.org/atlassian/atlaskit-mk-2/src/f8172af305826842a56ea6f0f31643c38e135656/packages/core/button/src/types.ts?at=master&fileviewer=file-view-default)
- [Shopify Polaris](https://github.com/Shopify/polaris-react/blob/master/src/components/Button/Button.tsx)
- [Apollo](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/QueryManager.ts)
- Formik
- React Router
- Styled Components
- Glamor 
- Emotion
- React Redux

**Libraries and UI Kits that use `Ixxx`**

- Blueprint
- [MobX React](https://github.com/mobxjs/mobx-react/blob/master/src/index.d.ts)

jeroenptrs:
@jaredpalmer I sort of agree. I'm used to using I as an interface prefix and T as a type prefix (I'm guessing you prefer neither to be enforced?) from past projects, but if you want shareable interfaces it looks a whole lot cleaner when it's not prefixed. I do not know what standards the Segment team wants to enforce, but I don't see prefixes in this PR either: segmentio/ui-box/pull/28

jaredpalmer:
One thing that isn’t really documented or spoken about with VSCode / TS is how fickle intellisense gets on large projects like this. My suggestion is to “Restart TypeScript Language Server” regularly. I probably do it 10-20 times a day. This will allow intellisense to re-evaluate the codebase from scratch without having to restart VSCode or kill any other processes.

jeroenptrs:
Upon restart it immediately errors again. Maybe I'm doing something wrong?

I declared the props of Text as the following (BoxProps are the propTypes of Box that Text extends):

```ts
export interface TextProps extends Partial<BoxProps> {
  ...
}
```

In one of the stories I've declared the following:

```ts
const Description = (props: TextProps) => (
  <Text is="p" marginTop={0} size={300} color="muted" {...props} />
)
```

Which prompts this error:
```const Text: React.FunctionComponent<TextProps>
import Text
Type '{ size: TextSize; fontFamily?: string; is: string; marginTop: number; color: string; }' is not assignable to type 'IntrinsicAttributes & TextProps & { children?: ReactNode; }'.
  Property 'is' does not exist on type 'IntrinsicAttributes & TextProps & { children?: ReactNode; }'.ts(2322)```

but property `is` *does* exist in BoxProps... I'm lost 😂

**Comment by jaredpalmer** [comment]
Yeah this isn’t documented well either. However, you need to add an annotation that tells TS that Description is a React Function Component. @types/react has a generic that does this. If you don’t use it, you have to manually declare the type of `children` or you’ll get the error you are seeing.

```tsx
import * as React from ‘react’

export const Description: React.FC<TextProps> = (props: TextProps) => ...

// this will give it a name in React DevTools
Description.displayName = ‘Description’
```

--
Jared

________________________________

jaredpalmer:
What's the status of this?

mshwery:
Hey @jeroenptrs @jaredpalmer ! Wow, love this. Sorry I'm just seeing it now. 

Segment has internally been building up typedefs for Evergreen (and there are `@types/ui-box` but we're migrating that entire codebase to TypeScript as well). 

Let me spend a little bit of time reviewing this effort.

jeroenptrs:
@jaredpalmer I needed the break 😬I was unable to fix intellisense which was pretty worrying for users of a library. I'm going to await feedback from @mshwery and the rest of team Evergreen before continuing as they are apparently working on it as well.

jeroenptrs:
Something I *did* come to realize is that proper typing of `ui-box` is desperately needed for `evergreen` to be succesfully available in typescript - pretty much the entire library depends on that package - so I hope that's available soon!

mshwery:
You can follow along the ui-box migration here: https://github.com/segmentio/ui-box/compare/types

.storybook/config.js
```diff
@@ -1,6 +1,9 @@
 import { configure } from '@storybook/react'
+import { setAddon } from '@storybook/react'
 import { setOptions } from '@storybook/addon-options'
 import 'react-chromatic/storybook-addon'
+import JSXAddon from 'storybook-addon-jsx'
+setAddon(JSXAddon)
```
mshwery: Are storybook changes/additions needed for the TypeScript migration?

package.json
```diff
@@ -42,10 +42,18 @@
...
+    "@types/classnames": "^2.2.7",
+    "@types/dom-helpers": "^3.4.1",
+    "@types/fuzzaldrin-plus": "^0.6.0",
+    "@types/react": "^16.8.10",
+    "@types/react-dom": "^16.8.3",
+    "@types/react-transition-group": "^2.0.16",
+    "@types/tinycolor2": "^1.4.1",
     "arrify": "^1.0.1",
     "classnames": "^2.2.6",
+    "csstype": "^2.6.3",
```
mshwery: Should all of these `@types/*` and `csstype` be in dev dependencies?

package.json
```diff
@@ -64,63 +72,111 @@
...
     ],
+    "extensions": [
+      "tsx",
+      "ts"
+    ],
     "envs": [
       "browser"
     ],
-    "space": true,
+    "4space": true,
```
mshwery: Is this change intentional? What is `4space`?

package.json
```diff
@@ -149,42 +205,37 @@
...
+          ".tsx"
+        ]
       }
     }
   },
   "prettier": {
     "semi": false,
     "singleQuote": true
   },
-  "lint-staged": {
```
mshwery: Was there a reason for removing `lint-staged` + `husky`?

mshwery: I think we could/should adapt this to lint ts files (and `--fix` what it can automatically).

package.json
```diff
@@ -149,42 +205,37 @@
...
-  },
   "ava": {
+    "compileEnhancements": false,
+    "extensions": [
+      "ts",
+      "tsx"
+    ],
     "files": [
-      "**/test/**/*.js"
+      "**/test/**/*.tsx"
```
mshwery: Should we include `.ts`?

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "noImplicitAny": true,
+    "module": "commonjs",
+    "target": "es5",
+    "jsx": "react",
+    "rootDir": "src",
+    "typeRoots": ["./src/types", "node_modules/@types"]
```
mshwery: Can we install the types from `@types/ui-box` instead of including it directly in the repo?

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
```
mshwery: I think we want to add `"declaration": true` here and include `"types": "./dist/evergreen-ui.d.ts" (or similarly named) in `package.json`.

mshwery:
Amazing effort. I have one main question:
Would it be possible for us to do a partial migration? i.e. setup TypeScript compilation and output a type declaration file, but without migrating the entire package in one PR?

jaredpalmer:
FWIW I have a fork building with typescript and tsdx, but with no implicit any turned off.

mshwery:
Looks like a neat lib @jaredpalmer  :)

What's the branch on your fork?

.storybook/config.js
jeroenptrs: It's no must, if it's becoming a bottleneck i can leave it out but if it's a succesful port to TSX I think it's better to make as much as possible Typescript :D

package.json
jeroenptrs: I believe csstype is only imported because of the ui-box types being taken internally *for now*

package.json
jeroenptrs: oops! not intentional

package.json
jeroenptrs: Yes, it's going back in once I'm done (it's in my todo in the OP). I've had issues with husky testing things while performing typescript ports in the past and it's best to just leave it off once the migration is done.

package.json
jeroenptrs: Not sure if there are going to be tests written in `.ts`, but probably a good idea yea.

tsconfig.json
jeroenptrs: As mentioned above, they're not complete. The `@types/ui-box` already had some issues with transferring css types over to the created Box element and `Evergreen` uses a lot of partial imports from `ui-box` in their prop-types (which the current `@types` version doesn't support).

tsconfig.json
jeroenptrs: Yes! Compiling will be fixed once I'm done with the actual migration, it's in my todo in the OP

jeroenptrs:
@mshwery with partal migration, do you mean only porting over the src and leaving tests and stories in js?

jeroenptrs:
@jaredpalmer very intrigued in TSDX, looking forward to see your changes (maybe you can merge them in my repo - I can add you as collaborator so you have free reign)

package.json
mshwery: It sounds like any types relied on in the output declaration file should be regular dependencies instead of dev deps. Ignore my previous comment.

jeroenptrs: Yep they should! You'd want that for your IDE's intellisense

mshwery:
>with partal migration, do you mean only porting over the src and leaving tests and stories in js?

@jeroenptrs I mean incremental migration. So we'd have TypeScript building our code with `allowJs: true` and would incrementally migrate individual files to .tsx.

This would have 3 benefits:
1. Decouple the package build changes from the typing effort. Reduced scope of changes, easier to review, lower risk, etc.
2. Make it easier for the rest of the community to contribute to the TypeScript migration. 
3. Allows us to be stricter about our TS settings because it's more tenable to migrate individual files in strict mode than an entire repo.

The main drawbacks I see are that:
- we have to allow js files temporarily (this isn't worse than the current state)
- we have an incomplete declaration file that would have to be supplemented with a handmade d.ts (either included with the package distro or by each consumer)
-

jeroenptrs:
@mshwery Sounds fair, and more inclusive to the community. If you guys decide on going forward in the incremental way I'd opt to close this PR until you've (potentially) done some initial setup and wrote up some guidelines (f.e. component x relies on component y, so it makes sense to migrate component y *first*, etc)

Alternatively, you could cherry pick some of the setup stuff I've done in here related to linting, testing, storybook etc to save you some time.

I'm interested in helping you guys forward in any way possible since this is easily my favourite component library (but I do prefer *Typescript all the things* 😅).

mshwery:
Awesome, this sounds good to me. I also want to be super respectful of your contribution here because this is truly a feat!! 👏

>Lastly, interested to hear @jaredpalmer's take on this since he's been following this up close. Seems to me that TSDX could help out with some of the setup and build problems that might arise in the future.

Agreed.

mshwery:
@jeroenptrs I've pulled out the minimal set of changes from your PR to get this working without changing the output build. It will allow us to incrementally migrate files to TypeScript in a easier-to-review manner.

https://github.com/segmentio/evergreen/compare/typescript?expand=1

It handles:
- [x] Support for `.tsx?` files as source code, storybook files and tests
- [x] `allowJs` so we can continue migrating the rest of the library over incrementally with stricter settings
- [x] `cd docs && yarn dev` and `cd docs && yarn build` both work
- [x] `yarn dev` works (storybook)
- [x] Output files are the same (continues to use babel)
- [x] Uses eslint typescript parser (via `xo`) since TSLint is considered "deprecated"
- [x] fixes missing peer deps in package.json
- [x] adds `yarn build-types` that will let us see what the declaration output would be
- [x] adds `yarn type-check` to use `tsc` to validate our types (since babel doesn't type check)

One caveat to this approach is that we won't be able to provide a declaration file until we've either fully moved over, or if we want to groom one by hand. I experimented with using `dts-gen` to produce a starting point. This is definitely an option.

My thinking is that we could start tagging a `v5.0.0-alpha` prerelease with all of these commits and start tackling migrating component by component (and we could cherry pick a lot of the work in this branch).

What do y'all think @jeroenptrs @jaredpalmer ?

jaredpalmer:
I would advise against this. I would change all files to .ts and .tsx and keep noImplicitAny as false but declarations set to true. Get all the tooling out of the way now. In another PR, type everything as any, number, or string so that you can turn back on noImplicitAny. Then add types as you go

jeroenptrs:
> change all files to .ts and .tsx and keep noImplicitAny as false but declarations set to true. Get all the tooling out of the way now

This is going to cause more issues than improving the workflow (because it's what I had in mind first too). Tinkering with you linter is best done in the beginning, but temporarily adding in overrides like `noImplicitAny: false` is counter-productive.

If you want to do this as a community effort, the way @mshwery has set it up seems a bit more productive. I want to stress that setting noImplicitAny to false isn't going to be beneficial.

jeroenptrs:
@mshwery I'm going to close up this PR and rebase my fork onto your typescript branch. I'll start opening PR's against that branch, 1-3 components at a time.

mshwery:
I'm game for either way – I think my main concern was trying to limit the scope of the initial branch to focus on getting TypeScript support in the build/test/lint/stories/docs. We have that now (in both of these branches).

I'm not against the idea of doing a global filename extension swap `.js` -> `.tsx?` and turning off `allowJs`

Would y'all be opposed to opening PRs into that main `/typescript` branch?

mshwery:
@jaredpalmer that sounds like a pretty small change, actually - would you want to open a PR against my branch to illustrate what you are thinking? Sounds like file renames + `declarations: true` + maybe a little extra?

mshwery:
@jeroenptrs you could probably start cherry-picking some of these component TS changes into new PRs against the `typescript` mainline branch :) I just opened one for all the `typography/*` components.

jeroenptrs:
@mshwery I checked out the `ts/typography` branch in my fork and will slowly start doing that! How should we go about merging those into your PR? Another PR 😂 ?

mshwery:
Yeah I was thinking we just spin up new branches off of the `typescript` branch and use it as an integration branch for the migration. Once we reach a level of "completeness" we are happy with we can merge the whole branch into `master`.

One thing I'm not 100% sure of is whether we should use this babel plugin that generates proptypes from typescript interfaces or if we should include propTypes by hand? On first pass it seems like the plugin doesn't quite cover the same bases (like ignoring TypeScript's `extends`).

mshwery:
@jeroenptrs it looks like Blueprint doesn't even have propTypes (they have their own notion of `validateProps`). If we can hand write the propTypes (even though it's duplicative) that might be the way to go. It might be easier too because we would just leave them as-is. I think your branch already did that, right?

jeroenptrs:
@mshwery I came to the conclusion that manual is the best way to go. Especially if you use custom types (like an aggregate of numbers, strings etc) - as you can see in my brach I had to extend the proptypes typing for some components quite a bit 😄

mshwery:
Alright @jeroenptrs , I'll refactor my `ts/typography` branch to keep the `propTypes` in there by hand.

PR #545, opened 2019-05-02
Task/include default checked

mshwery:
This PR introduces support for `defaultChecked` for people who want to use `Checkbox` or `Radio` as uncontrolled components. While this is a non-breaking change, previously `Checkbox` was written in a way that prevented you from using it as an uncontrolled component (due to `defaultProps`).

### DIFF
--- a/src/checkbox/src/Checkbox.js
+++ b/src/checkbox/src/Checkbox.js
@@ -79,6 +79,12 @@ class Checkbox extends PureComponent {
     onChange: PropTypes.func,
 
     /**
+     * When true, the checkbox is true initially.
+     * This is for uncontrolled usage.
+     */
+    defaultChecked: PropTypes.bool,
+
+    /**
      * When true, the radio is disabled.
      */
     disabled: PropTypes.bool,
@@ -102,8 +108,6 @@ class Checkbox extends PureComponent {
   }
 
   static defaultProps = {
-    checked: false,
-    indeterminate: false,
     onChange: () => {},
     appearance: 'default'
   }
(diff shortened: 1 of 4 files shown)


vadimdemedes:
What do you think about logging a warning message if both `checked` and `defaultChecked` are used at the same time?

mshwery:
@vadimdemedes does React do that if you use both? Why warn about expected React behavior?

vadimdemedes:
I don't know if React does that, but it sounds like a helpful warning to me, since `checked` is for controlled usage and `defaultChecked` is for uncontrolled usage. No reason to use both.

mshwery:
IMO that's excessive code (and probably would be annoying seeing that all the time). I'm already a bit weary of the various warnings from Evergreen and ui-box.

But happy to concede if y'all think it's really valuable for Evergreen consumers.

vadimdemedes:
Hm I disagree. React does a wonderful job on detecting wrong patterns and warning users about their mistakes. I think it's a very helpful thing to have and a wonderful developer UX.

Either way, I don't have strong feelings about it, feel free to merge as-is.

vadimdemedes:
> This PR introduces support for defaultChecked for people who want to use Checkbox or Radio as uncontrolled components.
Is this a request from within Segment or someone else asked for it in the repo? Would love to avoid uncontrolled usage in our app.

mshwery:
>previously Checkbox was written in a way that prevented you from using it as an uncontrolled component (due to defaultProps).

This is a blocker for simplifying some internal Segment features ^, but was shocked to see Evergreen doesn't support it, even as a pass-through prop.

I generally agree (prefer controlled components) but there are cases where it's a burden.

mshwery:
Going to stick with controlled usage, closing this until there's a stronger need for it.

PR #546, opened 2019-05-04
Rename `.js` -> `.tsx?` and turn off `allowJs`

mshwery:
This PR follows @jaredpalmer's recommendation to turn off `allowJs` and benefit from the compiler immediately, without adding types.

## TODO
- [x] Fix broken types with `any` (props, state, etc)
- [x] Add `declaration: true` to `tsconfig.json`

// cc @jaredpalmer @jeroenptrs


mshwery:
Turns out that this requires adding tons of `any` types 😢 even with implicit any turned off `"noImplicitAny": false`

mshwery:
Ok, `yarn type-check` passes, `yarn build-types` passes.

mshwery:
I did catch a couple bugs (missing properties being called as functions and a broken use of React context)

PR #549, opened 2019-05-08
upgrade ui-box@2.x

mshwery:
This PR upgrades us to `ui-box@2.x` which has typescript declarations built-in. I had to use glamor's `css` function directly instead of relying on `ui-box` to do it for us via a `css` prop (no longer supported).

// cc @jeroenptrs @jaredpalmer


src/typography/src/UnorderedList.tsx
```diff
@@ -26,15 +28,6 @@ export default class UnorderedList extends PureComponent<any & BoxProps> {
     size: 400
   }
 
-  static styles = {
-    is: 'ul',
```
mshwery: Box no longer likes generic `string` values... we may want to change the types to support it.

Issue #300, opened 2018-08-23
Migrate to TypeScript


Jetsly:

jeroenransijn:
No. Looking into Flow/TypeScript for the future. Probably a couple of major version out though.

pspeter3:
Would you take DTS PRs?

jeroenransijn:
@pspeter3 do you mean adding types? Not sure what DTS means.

pspeter3:
A single DTS file can provide type definitions for the entire project. It means you can continue to write in JavaScript while supporting TypeScript users.

paustint:
I would ❤️ to see typescript support!

**The options are**
1. include a DTS file along with the project, as @pspeter3 mentioned - a type definition file that is hosted as part of the repository and would need to be included alongside the JS files in the build on npm.
    - This should be kept up-to-date and should be updated along with each new feature/change.
2. Avoid including any type definitions within this project/repo and wait for the community to take ownership of this and publish the types here: https://github.com/DefinitelyTyped/DefinitelyTyped (which would then be published on npm as `@types/evergreen`)
    - The community would slowly keep this up-to-date as folks decide to contribute when changes occur.
3. Convert your library to typescript to get full compile-time type safety - then during compile and build, `.d.ts` definition files are generated automatically to be included along with the JS files generated during build.

Option 3 would be the best :) _(yes, I am biased because developing in typescript provides a much better developer experience compared to standard JS!)_
Option 2 is the most common for non-ts projects, but option 1 is preferable if the evergreen team agrees to the additional amount of ongoing work to ensure the types remain up-to-date.

zheeeng:
Plz consider reopen this issue. Evergreen's UI looks beautiful very much, the only reason blocks me using it is the missing of type definitions.

Idered:
This is heavy based on [segmentio/ui-box](https://github.com/segmentio/ui-box) so I would start with adding typescript there

balupton:
For the risk of being another +1, I just wanted to add in my support for this.

I deal with react a few months each year, and seems the best UI frameworks right now for React are:

- Segment's Evergreen
- Palantir's Blueprint (typescript)
- Ant (typescript)
- Shopify's Polaris (typescript)

After comparing their modal dialog solutions, seems order of easiest to hardest are:

1. https://evergreen.segment.com/components/dialog
2. https://ant.design/components/modal/
3. https://polaris.shopify.com/components/overlays/modal#navigation
[...]

PR #552, opened 2019-05-09
improve withTheme types

mshwery:
This PR adds better types for `withTheme`. I think we should start with defining (and expanding on) the interface first, and then we can make sure that `defaultTheme` meets the requirements / mirrors it.

This will be a good exercise to figure out where the theme is pulling values directly from the defaultTheme object instead of from the theme provider.

Overall I think theming is a bit complex (hierarchical nested config, sometimes from the context sometimes not) and would love to simplify our approach, but we can save that for another time.

Related: #300

### DIFF
--- a/src/theme/index.ts
+++ b/src/theme/index.ts
@@ -1,3 +1,3 @@
 export { default as defaultTheme } from './src/default-theme'
-export { ThemeProvider, ThemeConsumer } from './src/ThemeContext'
-export { default as withTheme } from './src/withTheme'
+export { ThemeProvider, ThemeConsumer, Theme } from './src/ThemeContext'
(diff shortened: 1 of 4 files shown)


romarioraffington:
Hey @mshwery, the definition files seem to not export ThemeContext,   ThemeContext, ThemeProvider, ThemeConsumer and useTheme. As a result ts breaks.

```js
import { ThemeContext  } from 'evergreen-ui'
// Module '"evergreen-ui"' has no exported member 'ThemeContext'
```

Thoughts an accepting a PR that add these to the definition file?

mshwery:
@romarioraffington PRs are very welcome!! 😄

PR #553, opened 2019-05-11
migrate typography components to typescript

mshwery:
This PR migrates all of the components in `typography` over to proper Props types.


src/typography/src/Heading.tsx
```diff
@@ -1,32 +1,40 @@
...
       PropTypes.bool,
       PropTypes.number,
       PropTypes.string
     ]),
-
-    /**
-     * Theme provided by ThemeProvider.
-     */
-    theme: PropTypes.object.isRequired
+    size: PropTypes.oneOf([100, 200, 300, 400, 500, 600, 700, 800, 900])
```
solon-aguiar: I wonder if instead of `size: PropTypes.oneOf([100, 200, 300, ...])` we could have `size: PropTypes.oneOf([size in HeadingSize]: number)`. Idea from [here](https://www.typescriptlang.org/docs/handbook/advanced-types.html) (I'm sorry, I couldn't find a way to link directly to it).

src/typography/src/UnorderedList.tsx
```diff
@@ -1,27 +1,33 @@
...
-     * You can override this on a individual list item.
-     */
     icon: PropTypes.string,
-
-    /**
-     * The color of the icon in each list item in the list.
-     */
-    iconColor: PropTypes.string
+    iconColor: PropTypes.string,
+    size: PropTypes.string as Validator<ListSize>
```
solon-aguiar: shouldn't this be a number instead of string?

src/typography/src/Link.tsx
```diff
@@ -1,42 +1,51 @@
...
+   * Only use if you know what you are doing.
+   */
+  className?: string
 
-    /**
-     * Specifies the URL of the linked resource. A URL might be absolute or relative.
-     */
-    href: PropTypes.string,
+  /**
+   * The color (and styling) of the Link. Can be default, blue, green or neutral.
```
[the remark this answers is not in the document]
mshwery: Because in addition to those aliases you can pass literally any color string.

src/typography/src/Heading.tsx
mshwery: Is it possible to enumerate the types as a value? I couldn't find the reference in the doc you linked.

solon-aguiar: Under `Mapped types` I saw this:

```
type Keys = 'option1' | 'option2';
type Flags = { [K in Keys]: boolean };
```

I'm not sure if this is the same case, but it looked similar to me.

src/typography/src/Link.tsx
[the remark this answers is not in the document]
solon-aguiar: Ahhh, okay. The doc confused me then. I thought it was only one of those four possibilities.

src/typography/src/Heading.tsx
mshwery: Right, but those are TypeScript types. I don't think you can use types as values in your code.

solon-aguiar: That makes sense. Thank you! I'm sorry for the back and forth. :)

PR #557, opened 2019-05-20
docs(Typography): Paragraph components scale 300 to 500

maraisr:
With this PR I aim to fix the paragraph component docs, where the scale being displayed runs through the keys `theme.typography.text`, whereas the text scale only runs 300 _to_ 500. As the paragraph style for 600 depicts 20px whereas this should be 16px.

Stretch question, should you guys expose the scale for this component outside the realm of propTypes? So if future scales get implemented for `Paragraph`, it'll be reflected in the docs?


mshwery:
@maraisr ideally we would get this from the source directly so the proptypes use the same array as the docs, but in lieu of that I think this is an improvement :)

PR #558, opened 2019-05-21
feat(popover): add prop to disable close on body click

### LINKED ISSUE #544: Disable closing Popover on outside click
Hi,

>Popovers Close On
Outside click
Escape key
The close function being called (advanced)

I'd like to disable closing `Popover` on `Outside click`, enable only `Escape key` or `The close function being called`.

Is there a way to do that with the current API? If not, could you please add this feature?

"evergreen-ui": "^4.15.0",

palmer-cl:
Gives users the ability to disable the popover closing when clicking outside the popover on the body.

Fixes #544

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -92,7 +92,7 @@ export default class Popover extends Component {
     /**
      * Boolean indicating if clicking outside the dialog should close the dialog.
      */
-    shouldNotCloseOnBodyClick: PropTypes.bool
+    shouldCloseOnExternalClick: PropTypes.bool
   }
 
   static defaultProps = {
@@ -105,7 +105,7 @@ export default class Popover extends Component {
     onOpenComplete: () => {},
     onCloseComplete: () => {},
     bringFocusInside: false,
-    shouldNotCloseOnBodyClick: false
+    shouldCloseOnExternalClick: true
   }
 
   constructor(props) {
@@ -192,7 +192,7 @@ export default class Popover extends Component {
       return
     }
 
-    if (this.props.shouldNotCloseOnBodyClick) {
+    if (this.props.shouldCloseOnExternalClick === false) {
       return
     }
 
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -92,7 +92,7 @@ export default class Popover extends Component {
     /**
      * Boolean indicating if clicking outside the dialog should close the dialog.
      */
-    shouldNotCloseOnOutsideClick: PropTypes.bool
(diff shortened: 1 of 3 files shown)


src/popover/src/Popover.js
```diff
@@ -87,7 +87,12 @@ export default class Popover extends Component {
     /**
      * When true, bring focus inside of the Popover on open.
      */
-    bringFocusInside: PropTypes.bool
+    bringFocusInside: PropTypes.bool,
+
+    /**
+     * Boolean indicating if clicking outside the dialog should close the dialog.
+     */
+    shouldNotCloseOnBodyClick: PropTypes.bool
```
mshwery: What about `shouldCloseOnExternalClick` to avoid the `shouldNot` negative?

palmer-cl: @mshwery You're welcome!

I was trying to make the default be falsy to prevent a breaking change.

If I go with `shouldCloseOnExternalClick`, I would need the default to be true, because I think the default behavior is that popover _does_ close on external click. Is my thinking correct there?

mshwery: That sounds right!

palmer-cl: Awesome. What are next steps for merging?

mshwery: @palmer-cl I can merge it in once we're all set with the naming!

palmer-cl: @mshwery re-read this comment: https://github.com/segmentio/evergreen/pull/558#discussion_r286265490

Im saying I **do not** think you want to name it what you proposed. It would be a breaking change to name it that and require default of true. Because by default, behavior should close on external click. We are adding a feature to NOT close on external click so the two options would be

1. `shouldCloseOnExternalClick` = true default (but this will be false on all current implementations)

2. `shouldNotCloseOnExternalClick` (like the user is proposing) = false default, but then passing it as true will give desired behavior.

mshwery: You can implement `shouldCloseOnExternalClick` in a non-breaking way.

`if (shouldCloseOnExternalClick === false)` + `static defaultProps = { shouldCloseOnExternalClick: true }` aka only when consumers explicitly pass it as false. That way when it's not explicitly passed at all we continue performing the current behavior, but if a consumer explicitly opts-out with `shouldCloseOnExternalClick={false}` we respect that.

palmer-cl: Roger that! I see what you're saying now. Updated in 24d05ac

palmer-cl:
@mshwery Let me know what you think.

Issue #469, opened 2019-01-04
How to remove warning:  "ui-box deprecation.."


indrabinridwan:
this warning always show on console.
```
ui-box deprecation: the “css” prop will be removed in the next major version in favour of importing glamor directly and passing it՚s generated class to the “className” prop.
render @ `box.js:175
```

mshwery:
Migrate off the `css` prop in `ui-box` or `evergreen-ui` to get rid of that warning. There might be some evergreen components that use it, too, which would have to be migrated off of the `css` prop as well.

dillonhafer:
It looks like it comes from `Pane` https://github.com/segmentio/evergreen/blob/master/src/layers/src/Pane.js#L177

indrabinridwan:
oh still confusing,  okey. thats warning will have effect to render performance?

dillonhafer:
In my experience, they only console log in development environment, not production.

mshwery:
Yeah, it should have no effect in production. They also shouldn't impact render performance, generally, even in development.

That said, we should consider migrating off of that prop in evergreen.

amalo95:
Is this fixed? should this be closed?

PR #561, opened 2019-05-24
Remove use of css prop

danieltott:
This is a fix for #469: [How to remove warning: "ui-box deprecation.."](#469)

This would allow for a pretty easy upgrade to [`ui-box` 2.0.0](https://github.com/segmentio/ui-box/releases/tag/v2.0.0), but I figured one change at a time was best.

### DIFF
--- a/src/layers/src/Pane.js
+++ b/src/layers/src/Pane.js
@@ -1,4 +1,5 @@
 import React, { PureComponent } from 'react'
+import { css as glamorCss } from 'glamor'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
 import { withTheme } from '../../theme'
@@ -178,11 +179,11 @@ class Pane extends PureComponent {
         borderLeft={_borderLeft}
         boxShadow={elevationStyle}
         background={theme.getBackground(background)}
-        css={{
+        {...glamorCss({
           ...css,
           ...hoverElevationStyle,
           ...activeElevationStyle
-        }}
+        })}
         {...props}
       />
     )
--- a/src/layers/src/Pane.js
+++ b/src/layers/src/Pane.js
@@ -183,14 +183,11 @@ class Pane extends PureComponent {
         background={theme.getBackground(background)}
         className={cx(
           className,
-          glamorCss(
-            {
-              ...css,
-              ...hoverElevationStyle,
-              ...activeElevationStyle
-            },
-            className
-          ).toString()
+          glamorCss({
+            ...css,
+            ...hoverElevationStyle,
+            ...activeElevationStyle
+          }).toString()
         )}
         {...props}
       />
--- a/src/layers/src/Pane.js
+++ b/src/layers/src/Pane.js
@@ -1,5 +1,6 @@
 import React, { PureComponent } from 'react'
 import { css as glamorCss } from 'glamor'
+import cx from 'classnames'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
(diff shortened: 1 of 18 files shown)


danieltott:
I don't have too much experience with glamor - I tried the `className` variation at first and it wasn't really working correctly (I'm sure I messed something up), so I went with this object spread and it worked great, plus it already handles dealing with conflicting/overriding `className` props without having to do any more work. It really seemed cleaner this way to me (as opposed to combining strings etc). But the glamor props do seem a little "magical", while with `className` it is easy to understand what is happening.

Does something like this look good?

```jsx
// Pane.js
<Box
  borderTop={_borderTop}
  borderRight={_borderRight}
  borderBottom={_borderBottom}
  borderLeft={_borderLeft}
  boxShadow={elevationStyle}
  background={theme.getBackground(background)}
  className={makeClassName(
    {
      ...css,
      ...hoverElevationStyle,
      ...activeElevationStyle
    },
    className
  )}
  {...props}
/>

// make-classname.js
import { css } from 'glamor'

export default function makeClassName(styles = {}, className) {
  const retClassName = css(styles)
  if (className) {
    return `${retClassName} ${className}`
  }
  return retClassName
}

```

mshwery:
@danieltott the way `ui-box` was using glamor internally was via className, so there shouldn't be any issues keeping with that functionality but moving into evergreen for now. Something like this should work:

```jsx
<Box
  className={classnames(
    this.props.className,
    someOtherClassName, 
    glamorCss(cssRules).toString()
  )}
/>
```

danieltott:
Ahh yes - I see that being used in a couple of places in Evergreen as well.

danieltott:
@mshwery I updated the PR to use `className`

I can see here the snapshot tests failing, but I can't get them to run locally, and also they're failing on a couple of components that I didn't touch...¯\_(ツ)_/¯

mshwery:
The snapshot failures actually might be related to using `className` instead of css, because now the order has changed.

I would recommend just updating the snapshots.

danieltott:
OK - I'm noticing some `css-nil` class values in there, so I'm going to look into it a bit more.

mshwery:
@danieltott I don't think the `css-nil` values are new though – that's just from the snapshot output from a set of empty styles iirc.

danieltott:
@mshwery I guess this was the one that was actually concerning:
image

But, I can't even get tests to run locally, I'm running into messages like this for every file inside /docs when I run `yarn run test`. Still looking into it so that I can resolve the snapshot discrepancies.

```
  docs/src/components/DesignersHero.js:2:28
  ✖    2:28  Unable to resolve path to module gatsby.
```

Probably not an issue with evergreen-ui, just figured I'd note it in case it's come up before.

src/layers/src/Pane.js
```diff
@@ -178,11 +181,17 @@ class Pane extends PureComponent {
...
-        }}
+        className={cx(
+          className,
+          glamorCss(
+            {
+              ...css,
+              ...hoverElevationStyle,
+              ...activeElevationStyle
+            },
+            className
```
mshwery: Why include `className` here?

danieltott: Good catch, that was left over from some experimenting I was doing at the beginning.

mshwery:
@danieltott I think you'll need to run `yarn ava --u` to update the snapshots to get the tests to pass.

For your issue, you could `cd docs && yarn install` just to bypass that issue, but I'm not sure why you are seeing it at all.

PR #564, opened 2019-05-24
Add onScroll prop

### LINKED ISSUE #563: onScroll event handler for Table.VirtualBody for infinite preloader
In our App we would like to implement infinitely loadable paginated list.
For that purpose we have to detect when table was scrolled to some defined number of rows. For that purpose there is `onScroll` callback in [react-tiny-virtual-list](https://github.com/clauderic/react-tiny-virtual-list), that is not exposed on `Table.VirtualBody`.
This little feature would save our use-case.

dy:
Fixes #563.
Should I add tests @mshwery @vadimdemedes?


dy:
Any chance we can get this merged  @mshwery @vadimdemedes?

Issue #534, opened 2019-03-29
Popover hover support


tj:
It would be sweet if you could hover the contents to show the popover, for hover menus. I was going to use Tooltip but it's missing the optional callback with `{ close }`. Just going to leave it as a click for now but happy to send a PR later when I have a chance if you're interested.

mshwery:
Sounds like a good addition @tj!

jtanya34:
Can I take up this issue?

robphoenix:
@tj Can this be closed by #565?

mshwery:
Yep! Closing.

PR #565, opened 2019-05-25
#534:open popover on hover also shown in menuhover in storybook

jtanya34:
The aim of this PR is to add an hover functionality on popover. The example is added as menuHover in menu stories.

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -25,6 +25,10 @@ export default class Popover extends Component {
      * When true, the Popover is manually shown.
      */
     isShown: PropTypes.bool,
+    /**
+     * Open the Popover based on click or hover. Default is click.
+     */
+    trigger: PropTypes.oneOf(['click', 'hover']),
 
     /**
      * The content of the Popover.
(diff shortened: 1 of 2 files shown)


src/popover/src/Popover.js
```diff
@@ -25,6 +25,7 @@ export default class Popover extends Component {
      * When true, the Popover is manually shown.
      */
     isShown: PropTypes.bool,
+    isHover: PropTypes.bool,
```
mshwery: Can you add some documentation to the prop, and maybe we could rename to `triggerOnHover`?

jtanya34:
> What would you think about providing this via an enum prop called `trigger`? That way we could easily distinguish between opening based on click or hover (or other mechanisms in the future)?
> 
> `trigger: 'click' | 'hover'` where `'click'` is the default value?

Good Idea, I made the changes accordingly.

src/popover/src/Popover.js
```diff
@@ -25,6 +25,7 @@ export default class Popover extends Component {
      * When true, the Popover is manually shown.
      */
     isShown: PropTypes.bool,
+    trigger: PropTypes.oneOf(['click', 'hover']),
```
mshwery: Could you add a jsdoc description for the prop here?

jtanya34: done!

mshwery: Oh, that reminds me – maybe we should add `trigger: 'click'` to `defaultProps`? I realize with the current implementation it's not strictly necessary, but would be good for semantics I think.

jtanya34: done

robphoenix:
🎉

PR #567, opened 2019-05-30
Added custom style to popover component to override default styles.

gjvpaet:
I added a custom style prop to popover component to override the default styles of the Positioner children. I have a use case that I need to override the z-index and I can't do that because the stateless props are overridden by the style prop.

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -45,6 +45,11 @@ export default class Popover extends Component {
     display: PropTypes.string,
 
     /**
+     * The custom style property allows to override the default property of Positioner.
+     */
+    customStyle: PropTypes.object,
+
+    /**
      * The min width of the Popover card.
      */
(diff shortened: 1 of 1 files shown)


src/popover/src/Popover.js
```diff
@@ -44,6 +44,11 @@ export default class Popover extends Component {
      */
     display: PropTypes.string,
 
+    /**
+     * The custom style property allows to override the default property of Positioner.
+     */
+    customStyle: PropTypes.object,
```
mshwery: Could we maybe name this `popoverStyle`?

src/popover/src/Popover.js
```diff
@@ -347,7 +353,7 @@ export default class Popover extends Component {
             }}
             data-state={state}
             css={css}
-            style={style}
+            style={[style, customStyle]}
```
mshwery: I don't think array notation works? Can you replace with `Object.assign({}, style, customStyle)`?

PR #569, opened 2019-06-03
add types for Alert and InlineAlert

mshwery:
This adds proper types for `Alert` and `InlineAlert`.

### DIFF
--- a/src/alert/src/Alert.tsx
+++ b/src/alert/src/Alert.tsx
@@ -1,67 +1,84 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
-import { withTheme } from '../../theme'
+import { withTheme, Theme } from '../../theme'
 import { Pane } from '../../layers'
 import { Heading, Paragraph } from '../../typography'
 import { IconButton } from '../../buttons'
 import { Icon } from '../../icon'
 
-class Alert extends PureComponent<any & React.ComponentProps<typeof Box>> {
+type Appearance = 'default' | 'card'
+type Intent = 'none' | 'success' | 'warning' | 'danger'
+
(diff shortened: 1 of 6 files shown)


src/table/src/TableVirtualBody.tsx
```diff
@@ -304,7 +304,7 @@ export default class TableVirtualBody extends PureComponent<
           scrollOffset={scrollOffset}
           scrollToAlignment={scrollToAlignment}
           renderItem={({ index, style }) => {
-            const child = children[index]
+            const child: any = children[index]
```
solon-aguiar: Any benefit in declaring as `React.ReactNode` instead of `any`?

src/alert/src/Alert.tsx
```diff
@@ -1,67 +1,84 @@
...
-     */
-    theme: PropTypes.object.isRequired
+    theme: PropTypes.object.isRequired as PropTypes.Validator<Theme>,
+    title: PropTypes.node
   }
 
-  static defaultProps = {
-    intent: 'none',
-    hasTrim: true,
+  static defaultProps: Partial<AlertProps> = {
```
Rowno: Isn't the `defaultProps` type automatically inferred from `PureComponent<AlertProps>`?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/588c5b81362cac6a460be54b3d890108c9e0c465/types/react/index.d.ts#L526

src/table/src/TableVirtualBody.tsx
mshwery: I've been using `any` with the intent that we will try to eradicate them later. This at least signals that we don't really know what the right solution is, but gotta get the compiler happy as we slowly migrate.

PR #570, opened 2019-06-03
fix(Positioner): Correctly overflow bottom's

maraisr:
With this PR I aim to resolve a random `right` being affected when we handle the overflow from an overflown `BOTTOM`.


src/positioner/test/index.js
```diff
@@ -112,3 +112,15 @@ test('It pushes the rect to the right if overflowing on the left side', t => {
     })
   )
 })
+
+test('It pushes the rect to the top if overflowing on the bottom side', t => {
```
mshwery: Perhaps this should merge with the test of the same type (I think it just didn't test for width)? https://github.com/segmentio/evergreen/blob/f7a15e97babcbcec118a2a8be6a314ae7cd9a65b/src/positioner/test/index.js#L92

maraisr: Ah good spot - thanks! Your unit test I think is still valid, I guess the unit test added was to test that the container is "locked" to the bottom of the view port, whereby the target is only _just_ off the view port. Updated the test.

mshwery:
This LGTM @maraisr but could you add a couple screenshots?

maraisr:
Thanks @mshwery! It's actually quite hard to get it into that div to show you. The `getPosition` function already handles bodies that _dont have enough room on the bottom_, I assume this conditional handles triggers that are somewhere at the top of the viewport (but inside the view), where the content of the positioner shoots out bottom of the viewport, whereby it doesnt fit _on the top_ and neither on the bottom. So that "glues" it to the bottom of the viewport plus some offset?

PR #571, opened 2019-06-03
implement types for Avatar

mshwery:
This PR adds types for the `Avatar` component.

### DIFF
--- a/src/avatar/src/Avatar.tsx
+++ b/src/avatar/src/Avatar.tsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types'
 import Box from 'ui-box'
 import { Image } from '../../image'
 import { Text } from '../../typography'
-import { withTheme } from '../../theme'
+import { withTheme, Theme } from '../../theme'
 import globalGetInitials from './utils/getInitials'
 import globalHash from './utils/hash'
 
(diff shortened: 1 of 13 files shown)


src/avatar/src/Avatar.tsx
```diff
@@ -17,65 +17,84 @@ const initialsStyleClass = css({
...
+    isSolid: PropTypes.bool,
+    name: PropTypes.string,
+    size: PropTypes.number.isRequired,
+    sizeLimitOneCharacter: PropTypes.number.isRequired,
+    src: PropTypes.string,
+    theme: PropTypes.object.isRequired as PropTypes.Validator<Theme>
   }
 
-  static defaultProps = {
+  static defaultProps: Partial<AvatarProps> = {
```
Rowno: Isn't the `defaultProps` type automatically inferred from `PureComponent<AvatarProps, AvatarState>`?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/588c5b81362cac6a460be54b3d890108c9e0c465/types/react/index.d.ts#L526

mshwery: It's supposed to be sometimes complains about the types still. It might be solved with `as const` though.

src/avatar/src/Avatar.tsx
```diff
@@ -17,65 +17,84 @@ const initialsStyleClass = css({
...
+  /**
+   * When true, force show the initials.
+   * This is useful in some cases when using Gravatar and transparent pngs.
+   */
+  forceShowInitials?: boolean
+
+  /**
+   * Function to get the initials based on the name.
+   */
+  getInitials: (name?: string | any, fallback?: string) => string
```
Rowno: Is it possible to do `string | unknown`? Because that would be more type safe.

src/avatar/src/utils/hash.ts
```diff
@@ -1,8 +1,8 @@
 // http://erlycoder.com/49/javascript-hash-functions-to-convert-string-into-integer-hash-
-export default function hashCode(s) {
+export default function hashCode(s: any): number {
```
Rowno: Is it possible to use `s: unknown` here?

src/avatar/src/Avatar.tsx
Rowno: `as const` should fix any errors with string literal types.

src/avatar/src/utils/hash.ts
mshwery: I can try that out.

PR #572, opened 2019-06-04
Add proper types for the Button components

Rowno:
This adds proper TypeScript types for all the Button components as part of the TypeScript migration.

I also went ahead and upgraded TypeScript.

### DIFF
--- a/src/buttons/src/Button.tsx
+++ b/src/buttons/src/Button.tsx
@@ -1,7 +1,6 @@
 import React, { PureComponent, ReactNode } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
-import Box from 'ui-box'
 import { Text } from '../../typography'
 import { Icon, IconName } from '../../icon'
 import { Spinner } from '../../spinner'
@@ -10,7 +9,7 @@ import { withTheme, Theme } from '../../theme'
 type Intent = 'none' | 'success' | 'warning' | 'danger'
 type Appearance = 'default' | 'minimal' | 'primary'
 
-interface ButtonProps extends React.ComponentProps<typeof Box> {
+interface ButtonProps extends React.ComponentProps<typeof Text> {
   /**
    * The intent of the button.
    */
--- a/src/buttons/src/Button.tsx
+++ b/src/buttons/src/Button.tsx
@@ -1,68 +1,84 @@
-import React, { PureComponent } from 'react'
+import React, { PureComponent, ReactNode } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 import Box from 'ui-box'
 import { Text } from '../../typography'
-import { Icon } from '../../icon'
+import { Icon, IconName } from '../../icon'
 import { Spinner } from '../../spinner'
-import { withTheme } from '../../theme'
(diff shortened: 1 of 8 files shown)


src/theme/src/ThemeContext.ts
```diff
@@ -50,6 +50,10 @@ export interface Theme {
   getTextInputClassName: (alias: 'none') => string
   getTextSizeForControlHeight: (height: number) => number
   getTextStyle: (size?: number) => any
+  getButtonClassName: (appearance: string, intent: string) => string
```
mshwery: Nit: can we alphabetize these (or automate that)?

src/buttons/src/TextDropdownButton.tsx
```diff
@@ -1,56 +1,61 @@
...
-import { Icon } from '../../icon'
+import { Icon, IconName } from '../../icon'
 import { Spinner } from '../../spinner'
-import { withTheme } from '../../theme'
+import { withTheme, Theme } from '../../theme'
 
-class TextDropdownButton extends PureComponent<
-  any & React.ComponentProps<typeof Box>
-> {
+interface TextDropdownButtonProps extends React.ComponentProps<typeof Box> {
```
mshwery: Hm, I think this should technically extend `TextProps` because it passes `...props` to `<Text />` not `<Box />`

src/buttons/src/Button.tsx
```diff
@@ -110,10 +126,11 @@ class Button extends PureComponent<any & React.ComponentProps<typeof Box>> {
     const iconSize = theme.getIconSizeForButton(height)
 
     const pr =
-      paddingRight !== undefined ? paddingRight : Math.round(height / 2) // eslint-disable-line no-negated-condition
-    const pl = paddingLeft !== undefined ? paddingLeft : Math.round(height / 2) // eslint-disable-line no-negated-condition
+      typeof paddingRight === 'number' ? paddingRight : Math.round(height / 2)
```
mshwery: Shouldn't this also accept `string`?

src/buttons/src/Button.tsx
```diff
@@ -1,68 +1,84 @@
...
-import { Icon } from '../../icon'
+import { Icon, IconName } from '../../icon'
 import { Spinner } from '../../spinner'
-import { withTheme } from '../../theme'
+import { withTheme, Theme } from '../../theme'
+
+type Intent = 'none' | 'success' | 'warning' | 'danger'
+type Appearance = 'default' | 'minimal' | 'primary'
+
+interface ButtonProps extends React.ComponentProps<typeof Box> {
```
mshwery: This might need to be extending `TextProps` (or `typeof Text`)?

src/buttons/src/TextDropdownButton.tsx
Rowno: Ah, true.

src/buttons/src/Button.tsx
Rowno: No, because of this line:
https://github.com/segmentio/evergreen/blob/db78ab2f5739aaa96e2303bbb0a47d3f64425144/src/buttons/src/Button.tsx#L139

PR #576, opened 2019-06-06
Remove `selected` restraint

dy:
We may want to store in select menu something other than just numbers/strings, for example, dates or date ranges (arrays of two dates). All works fine but this prop type check raises an error.
@mshwery

### DIFF
--- a/src/menu/src/MenuOptionsGroup.js
+++ b/src/menu/src/MenuOptionsGroup.js
@@ -39,10 +39,10 @@ class MenuOptionsGroup extends React.PureComponent {
           </Heading>
         )}
         <Pane>
-          {options.map(option => {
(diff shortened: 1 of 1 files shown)


mshwery:
I'm not 100% sure how that will function since `option.value` is also expected to be used as the `key`

dy:
@mshwery in [the docs](https://reactjs.org/docs/lists-and-keys.html) the `toString` is always called to explicitly cast type. Almost all non-object primitives have decent `toString` support.

dy:
I guess for objects without `toString` values that can be an issue, but that still extends support.
May I possibly ask to merge this and #587, if there's no unwanted effects introduced, and release a patch?

dy:
Thanks @mshwery!

PR #577, opened 2019-06-06
Add proper types for Badge and a bunch of the theme code

Rowno:
This adds proper TypeScript types for all the Badge components and a bunch of the theme code as part of the TypeScript migration.

### DIFF
--- a/src/badges/src/Badge.tsx
+++ b/src/badges/src/Badge.tsx
@@ -2,24 +2,36 @@ import React, { PureComponent } from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import { Strong } from '../../typography'
-import { withTheme } from '../../theme'
+import { withTheme, Theme } from '../../theme'
 
-class Badge extends PureComponent<any & React.ComponentProps<typeof Strong>> {
+interface BadgeProps extends React.ComponentProps<typeof Strong> {
+  /**
+   * The color used for the badge.
+   */
+  color: string
+
+  /**
+   * Whether or not to apply hover/focus/active styles
+   */
+  isInteractive: boolean
+
+  /**
+   * Whether or not a dark background with light text should be used
+   */
+  isSolid: boolean
+
+  /**
+   * Theme provided by ThemeProvider.
+   */
+  theme: Theme
+}
+
+class Badge extends PureComponent<BadgeProps> {
   static propTypes = {
-    /**
-     * The color used for the badge.
-     */
     color: PropTypes.string.isRequired,
-
-    /**
-     * Whether or not to apply hover/focus/active styles
-     */
-    isInteractive: PropTypes.bool,
-
-    /**
-     * Theme provided by ThemeProvider.
-     */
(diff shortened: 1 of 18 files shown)


src/typography/src/Text.tsx
```diff
@@ -10,17 +10,17 @@ export interface TextProps extends React.ComponentProps<typeof Box> {
   /**
    * The color (alias or valid color) applied to the text
    */
-  color?: string
+  color: string
 
   /**
    * The font family alias applied to the text
    */
-  fontFamily?: FontFamily
```
mshwery: These aren't really required though – are they?

src/typography/src/Paragraph.tsx
```diff
@@ -10,19 +10,19 @@ export interface ParagraphProps extends React.ComponentProps<typeof Box> {
...
    * The color (alias or valid color) applied to the text
    */
-  color?: string
+  color: string
 
   /**
    * Font family.
    * Can be: `ui`, `display` or `mono` or a custom font family.
    */
-  fontFamily?: FontFamily
```
mshwery: These aren't really required are they? Can't you pass `undefined` to them and still get back the right values?

src/theme/src/ThemeContext.ts
```diff
@@ -7,53 +7,32 @@ export interface Theme {
...
-  getIconSizeForIconButton: (height: number) => number
-  getLinkClassName: (color?: string) => string
-  getParagraphStyle: (size?: number) => any
-  getTagInputClassName: (alias: 'default') => string
-  getTextColor: (colorAlias?: string) => string
-  getTextDropdownButtonClassName: () => string
-  getTextInputClassName: (alias: 'none') => string
-  getTextSizeForControlHeight: (height: number) => number
-  getTextStyle: (size?: number) => any
+  getAlertProps: typeof defaultTheme.getAlertProps
```
mshwery: :awyis:

src/badges/src/Badge.tsx
```diff
@@ -2,24 +2,36 @@ import React, { PureComponent } from 'react'
...
+  color: string
 
-    /**
-     * Whether or not to apply hover/focus/active styles
-     */
-    isInteractive: PropTypes.bool,
+  /**
+   * Whether or not to apply hover/focus/active styles
+   */
+  isInteractive: boolean
```
mshwery: Technically this shouldn't be required?

src/typography/src/Text.tsx
Rowno: They are, but they have `defaultProps` so you don't need to set them.

src/typography/src/Paragraph.tsx
Rowno: Some of the theme functions don't handle undefined cases. But these have `defaultProps` so you don't need to set them.

src/badges/src/Badge.tsx
Rowno: It has a `defaultProp`.

PR #578, opened 2019-06-08
expose the`ThemeContext` to use the new context hook

2fd:
According to the documentation in the `useContext` hook is necessary to use directly the entire `context`:
https://reactjs.org/docs/hooks-reference.html#usecontext

If I try to pass the `Consumer` or the `Provider` to `useContext` the following messages are printed in the console:

```javascript
    useContext(ThemeConsumer)
```
Screen Shot 2019-06-08 at 08 13 31

```javascript
    useContext(ThemeProvider)
```

Screen Shot 2019-06-08 at 08 13 55

to prevent this I expose the `ThemeContext`


2fd:
@mshwery Done! Thanks for the review.

PR #583, opened 2019-06-13
Fix selected proptype

dy:
Selected value can be array of the option items' values.
For now it only allows strings, whereas it can do numbers too, and it does, but gives warnings in console.


solon-aguiar:
Thanks for finding and fixing this! 🎉

PR #586, opened 2019-06-13
Expose onBodyClick to popover

dy:
With controlled Popover state via `isShown` we may want to still be able to intercept bodyClick event.
Now there's no way to do that. 

Workarounds for manual bodyclick interception are horrible, since react events system disregards native bubbling, and popover gives no access to refs. @mshwery

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -198,7 +198,7 @@ export default class Popover extends Component {
       return
     }
     
-    // notify body click
+    // Notify body click
     this.props.onBodyClick(e)
 
     if (this.props.shouldCloseOnExternalClick === false) {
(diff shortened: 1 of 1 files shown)


dy:
I can, but there’s no way to ignore clicks coming from the popover itself. Besides, document native events have own bubbling, apart from react.

dy:
@mshwery can you please have a look at that again. That's still a blocker in our app - there's no way to reliably intercept clicks on body, coming not from the popover.

mshwery:
Curious if we should add an `event.preventDefault` or stop propagation instead?

dy:
You mean inside `onBodyClick` prop? Not sure if that's necessary now - that prop is perfect for organizing multiple popovers or custom popover logic.

Issue #456, opened 2018-11-25
SelectMenu close on click


andrwh:
Would love the ability to close the SelectMenu when an option is clicked

robphoenix:
Not stale. :smiley_cat: 
Would be good to hear any opinion from the maintainers, whether this is a possible consideration.

mshwery:
Sounds like a good addition! If y'all want to take a stab at a PR to implement this, I would be happy to review.  If not, we can add it to the docket.

// cc @solon-aguiar

robphoenix:
I'll take a look at opening a PR for this, though I'm new-ish to React so it may take a wee while to get my round what needs to be done where. Making no promises.

PR #588, opened 2019-06-19
Add closeOnSelect prop to SelectMenu

robphoenix:
~~This is a rough WIP to check this is the right approach, or if there is a better way.~~

In response to #456 this change adds a prop to `SelectMenu` to enable closing the select menu when an option is selected.

### DIFF
--- a/src/select-menu/src/OptionsList.js
+++ b/src/select-menu/src/OptionsList.js
@@ -36,9 +36,16 @@ export default class OptionsList extends PureComponent {
     isMultiSelect: PropTypes.bool,
 
     /**
+     * When true, menu closes on option selection.
+     */
+    closeOnSelect: PropTypes.bool,
+
+    /**
      * This holds the values of the options
      */
-    selected: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.string, PropTypes.number])),
+    selected: PropTypes.arrayOf(
+      PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+    ),
     onSelect: PropTypes.func,
     onDeselect: PropTypes.func,
     onFilterChange: PropTypes.func,
@@ -201,6 +208,9 @@ export default class OptionsList extends PureComponent {
 
   handleSelect = item => {
     this.props.onSelect(item)
+    if (!this.props.isMultiSelect && this.props.closeOnSelect) {
+      this.props.close()
+    }
   }
 
   handleDeselect = item => {
--- a/docs/src/pages/components/select-menu.mdx
+++ b/docs/src/pages/components/select-menu.mdx
@@ -18,6 +18,12 @@ and uses `react-tiny-virtual-list` for the rendering of the virtualized list of 
 The `SelectMenu` is unopinionated in how many items are selected in the list.
 Pass an array to the `selected` prop to select more items.
 
+## Close on select
+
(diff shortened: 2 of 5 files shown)


src/select-menu/src/OptionsList.js
```diff
@@ -201,6 +204,9 @@ export default class OptionsList extends PureComponent {
 
   handleSelect = item => {
     this.props.onSelect(item)
+    if (!this.props.isMultiSelect && this.props.closeOnSelect) {
```
solon-aguiar: I realized that hitting the `Enter` key has [a similar behavior](https://github.com/segmentio/evergreen/blob/master/src/select-menu/src/OptionsList.js#L190-L192) (closing the menu). It makes me wonder if we should also have the check for `multiSelect` and `closeOnSelect` in the `handleEnter` method (you can always close the menu by hitting `ESC` in that case). Doing this would be a change of contract in the interface. I'm not sure if we want to do that now.

@mshwery - would love some thoughts!

src/select-menu/src/OptionsList.js
```diff
@@ -48,7 +50,8 @@ export default class OptionsList extends PureComponent {
     filterPlaceholder: PropTypes.string,
     filterIcon: PropTypes.string,
     optionsFilter: PropTypes.func,
-    defaultSearchValue: PropTypes.string
+    defaultSearchValue: PropTypes.string,
+    closeOnSelect: PropTypes.bool
```
mshwery: While the name is pretty semantic would you mind adding some jsdoc description above the new prop?

src/select-menu/src/OptionsList.js
mshwery: I think we could save that for a separate PR if the current behavior is problematic. I think it's fine to leave alone for now TBH, though I agree with your idea!

src/select-menu/src/OptionsList.js
```diff
@@ -38,7 +38,9 @@ export default class OptionsList extends PureComponent {
     /**
      * This holds the values of the options
      */
-    selected: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.string, PropTypes.number])),
+    selected: PropTypes.arrayOf(
```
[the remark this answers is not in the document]
robphoenix: I'm using VS Code. It's picking up the prettier settings in the `package.json` and as there is no `printWidth` setting is going with the default of 80. This line is longer than that, which is why it is being formatted. 
I've left it as is as the prettier settings would need to explicitly state a larger `printWidth`. Is this ok? Would you rather I change it back?

src/select-menu/src/OptionsList.js
robphoenix: no worries, done

robphoenix:
Thanks @solon-aguiar @mshwery!

I've updated my PR & commit message, and added a note to the docs also.
I've left the formatting change for now, as it is complying with the prettier settings, lmk if you want me to change it. 👍

src/select-menu/src/OptionsList.js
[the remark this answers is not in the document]
solon-aguiar: Weird. Maybe nobody had touched the file in a while. Thanks for checking!

robphoenix:
No worries, welcome :fireworks:

PR #596, opened 2019-07-15
Add textAlign prop to TextTableCell

robphoenix:
The `isNumber` prop, when true, should align the text right. This commit adds the `textAlign` prop to the Text component, when `isNumber` is true.

### DIFF
--- a/src/table/src/TextTableCell.js
+++ b/src/table/src/TextTableCell.js
@@ -40,7 +40,7 @@ export default class TextTableCell extends PureComponent {
           size={300}
           flex="1"
           {...ellipsis}
-          {...(isNumber ? { fontFamily: 'mono' } : {})}
+          {...(isNumber ? { fontFamily: 'mono', textAlign: 'right' } : {})}
           {...textProps}
         >
           {children}


mshwery:
Oh I didn't realize this was in accordance with what the docs say! I'm sorry for not checking there.

Hm... I still think we probably should let consumers choose (or at least be able to override it). WDYT?

robphoenix:
Honestly I'm not sure.

I wonder if just doing the one thing, adding the `mono` font family, is enough to warrant using the `isNumber` prop, or even having it in the first place, and whether many folk will be adding `textAlign="right"` on number columns themselves anyway.

OTOH, though small, it is a breaking change, and may trip up some consumer apps, and, as you say, it's not hard for a consumer to define a number column the way they want it. :man_shrugging: 

Personally I'd probably remove the `isNumber` prop entirely, perhaps deprecating it in major version release.

mshwery:
>OTOH, though small, it is a breaking change, and may trip up some consumer apps, and, as you say, it's not hard for a consumer to define a number column the way they want it. 🤷‍♂
>
>Personally I'd probably remove the isNumber prop entirely, perhaps deprecating it in major version release.

@robphoenix yeah I agree with all of that

PR #598, opened 2019-07-17
Add types to majorScale & minorScale

robphoenix:
re: #300 

I wasn't sure if you wanted to keep the `TypeError` in there or not? So I left it in.


mshwery:
Thanks for contributing @robphoenix !

PR #599, opened 2019-07-17
Add types to Textarea

robphoenix:
Re: #300

### DIFF
--- a/src/textarea/src/Textarea.tsx
+++ b/src/textarea/src/Textarea.tsx
@@ -2,65 +2,78 @@ import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 import { Text } from '../../typography'
-import { withTheme } from '../../theme'
+import { withTheme, Theme } from '../../theme'
 
-class Textarea extends PureComponent<any & React.ComponentProps<typeof Text>> {
+interface TextareaProps extends React.ComponentProps<typeof Text> {
+  /**
+   * Makes the textarea element required.
+   */
+  required: boolean
+
+  /**
+   * Makes the textarea element disabled.
+   */
+  disabled: boolean
+
+  /**
+   * Sets visual styling of _only_ the text area to be "invalid".
+   * Note that this does not effect any `validationMessage`.
+   */
+  isInvalid: boolean
+
+  /**
+   * Use the native spell check functionality of the browser.
+   */
+  spellCheck: boolean
+
+  /**
+   * Allow the Grammarly browser extension to attach to the backing textarea.
+   */
+  grammarly: boolean
+
+  /**
+   * The placeholder text when there is no value present.
+   */
+  placeholder: string
+
+  /**
+   * The appearance of the TextInput.
+   */
+  appearance: 'default'
+
+  /**
+   * The width of the TextInput.
+   */
+  width: string | number
+
+  /**
+   * Theme provided by ThemeProvider.
+   */
+  theme: Theme
+
+  /**
+   * Class name passed to the textarea.
+   * Only use if you know what you are doing.
+   */
+  className: string
+}
+
+class Textarea extends PureComponent<any & TextareaProps> {
   static propTypes = {
-    /**
-     * Makes the textarea element required.
-     */
     required: PropTypes.bool,
-
-    /**
-     * Makes the textarea element disabled.
-     */
     disabled: PropTypes.bool,
-
-    /**
-     * Sets visual styling of _only_ the text area to be "invalid".
-     * Note that this does not effect any `validationMessage`.
-     */
     isInvalid: PropTypes.bool,
-
-    /**
-     * Use the native spell check functionality of the browser.
-     */
     spellCheck: PropTypes.bool,
-
-    /**
-     * Allow the Grammarly browser extension to attach to the backing textarea.
-     */
     grammarly: PropTypes.bool,
-
-    /**
-     * The placeholder text when there is no value present.
-     */
     placeholder: PropTypes.string,
-
-    /**
-     * The appearance of the TextInput.
-     */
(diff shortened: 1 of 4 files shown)


robphoenix:
Add https://github.com/segmentio/evergreen/pull/599/commits/5fae27d1ca7df1f5e34f8441c8d3e7ff73bc03e9 after seeing #577

src/textarea/src/Textarea.tsx
```diff
@@ -2,65 +2,78 @@ import React, { PureComponent } from 'react'
...
+   * Only use if you know what you are doing.
+   */
+  className: string
+}
 
-    /**
-     * Class name passed to the textarea.
-     * Only use if you know what you are doing.
-     */
+class Textarea extends PureComponent<any & TextareaProps> {
```
mshwery: I think we want to avoid `any` – is there any reason we can't remove that from the prop type?

src/theme/src/ThemeContext.ts
```diff
@@ -49,6 +49,7 @@ export interface Theme {
   getLinkClassName: (color?: string) => string
   getParagraphStyle: (size?: number) => any
   getTagInputClassName: (alias: 'default') => string
+  getTextareaClassName: typeof defaultTheme.getTextareaClassName
```
mshwery: I think this is probably fine, but any reason to deviate from the others here?

mshwery: Should we adopt this style for the rest of these functions on the theme?

robphoenix: I was pre-empting #577, expecting it to be merged first, where @Rowno adopts this style across all these functions: https://github.com/segmentio/evergreen/pull/577/files/51ddc1b740c211224cdc1ebe736fde77657d3395#diff-06dbe722e90e578810dc6405c11a6f96

src/textarea/src/Textarea.tsx
robphoenix: Removing it gives the following error:

```
Argument of type 'typeof Textarea' is not assignable to parameter of type 'ComponentType<TextareaProps>'.
  Type 'typeof Textarea' is not assignable to type 'ComponentClass<TextareaProps, any>'.
    Types of property 'propTypes' are incompatible.
      Type '{ required: Requireable<boolean>; disabled: Requireable<boolean>; isInvalid: Requireable<boolean>; spellCheck: Requireable<boolean>; grammarly: Requireable<boolean>; placeholder: Requireable<...>; appearance: Requireable<...>; width: Requireable<...>; theme: Validator<...>; className: Requireable<...>; }' is not assignable to type 'WeakValidationMap<TextareaProps>'.
        Types of property 'required' are incompatible.
          Type 'Requireable<boolean>' is not assignable to type 'Validator<boolean>'.
            Types of property '[nominalTypeHack]' are incompatible.
              Type 'boolean | null | undefined' is not assignable to type 'boolean | undefined'.
                Type 'null' is not assignable to type 'boolean | undefined'.
```

The `PropTypes.boolean` used for `required`, `disabled` etc. are of type `Requirable<boolean>`, and the `React.ComponentType` param in `withTheme` is expecting `Validator<boolean>`. This is the same for the other prop types types. We can fix it by casting all the proptypes as follows:

```
static propTypes = {
    required: PropTypes.bool as PropTypes.Validator<boolean>,
    disabled: PropTypes.bool as PropTypes.Validator<boolean>,
    isInvalid: PropTypes.bool as PropTypes.Validator<boolean>,
    spellCheck: PropTypes.bool as PropTypes.Validator<boolean>,
    grammarly: PropTypes.bool as PropTypes.Validator<boolean>,
    placeholder: PropTypes.string as PropTypes.Validator<string>,
    appearance: PropTypes.string as PropTypes.Validator<'default'>,
    width: PropTypes.oneOfType([
      PropTypes.string,
      PropTypes.number
    ]) as PropTypes.Validator<ReactText>,
    theme: PropTypes.object.isRequired as PropTypes.Validator<Theme>,
    className: PropTypes.string as PropTypes.Validator<string>
  }
```

It isn't a problem in other components, such as `Badge` & `Avatar` as the `.isRequired` on the proptype changes it from a `Requirable` to a `Validator` type.

I'm afraid I don't yet know enough about this to suggest the best fix.

mshwery: I think you can do something like this: https://github.com/segmentio/evergreen/blob/2281d9bd93cf348a60ecfe2047f1747c2d2fd516/src/typography/src/Text.tsx#L38-L39

src/theme/src/ThemeContext.ts
mshwery: Makes sense! Sounds good to me.

src/textarea/src/Textarea.tsx
robphoenix: 👍

robphoenix:
@mshwery I've updated the prop types.

src/textarea/src/Textarea.tsx
```diff
@@ -58,18 +58,21 @@ interface TextareaProps extends React.ComponentProps<typeof Text> {
...
+    disabled: PropTypes.bool as Validator<boolean>,
+    isInvalid: PropTypes.bool as Validator<boolean>,
+    spellCheck: PropTypes.bool as Validator<boolean>,
+    grammarly: PropTypes.bool as Validator<boolean>,
+    placeholder: PropTypes.string as Validator<string>,
+    appearance: PropTypes.string as Validator<'default'>,
+    width: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number
+    ]) as PropTypes.Validator<ReactText>,
```
mshwery: TIL `ReactText`. Nice!

robphoenix:
👌 thanks!

PR #600, opened 2019-07-18
Fix doc comment for isNumber prop

robphoenix:
The `isNumber` prop doesn't add `textAlign: right` to the cell.

Follow on from #596


robphoenix:
ping @mshwery

robphoenix:
Thanks!

PR #608, opened 2019-08-08
Fixes #607 - Set the default list position to inside for UnorderedList and OrderedList

### LINKED ISSUE #607: Default `list-style-position` styling not applying for lists
### Description
The default `listStylePosition: 'inside'` for the `UnorderedList` and `OrderedList` components is not applying. I verified this in Chrome and Firefox.

I'm pretty sure the issue is that since styles are listed alphabetically for these components, the `listStyle` style is implicitly setting the "listStylePosition" [which defaults to "outside"](https://www.w3schools.com/cssref/pr_list-style-position.asp) overriding what is specified for `listStylePosition`: https://github.com/segmentio/evergreen/blob/aca37cb/src/typography/src/UnorderedList.js#L36-L37

I believe this could be fixed by explicitly specifying the position within the `listStyle` directly.

MrAlexLau:
Per issue #607 : This sets the `listStylePosition` directly so that it's not clobbered by `listStyle`

### DIFF
--- a/src/typography/src/OrderedList.js
+++ b/src/typography/src/OrderedList.js
@@ -22,7 +22,6 @@ export default class OrderedList extends PureComponent {
     margin: 0,
     marginLeft: '1.1em',
     padding: 0,
-    listStylePosition: 'inside',
     listStyle: 'decimal'
   }
 
--- a/src/typography/src/UnorderedList.js
+++ b/src/typography/src/UnorderedList.js
@@ -33,7 +33,6 @@ export default class UnorderedList extends PureComponent {
     margin: 0,
     marginLeft: '1.1em',
     padding: 0,
-    listStylePosition: 'inside',
     listStyle: 'disc'
   }
 


MrAlexLau:
@mshwery - did you mean the docs which show what `listStyle` and `listStylePosition` default to? I was looking on MDN:

- [list-style-position](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position#Specifications)
- [list-style](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style#Specifications) - note that this in particular defaults to shorthand for `list-style-position: outside`

mshwery:
@MrAlexLau I meant the evergreen docsite (locally running your change) or a small proof of concept to illustrate the visual change that this makes.

Debating whether or not this would be considered a breaking change for some consumers if the visual outcome is different.

MrAlexLau:
@mshwery - oh sure I see what you mean. 

### Before
before

### After
after

There does seem to be a possibility that this would be a breaking change. An alternative to these changes would be to just remove `listStylePosition: 'inside'` altogether since it is overwritten. That way it's less confusing to those inspecting the source code.

mshwery:
@MrAlexLau let's go ahead and just remove it, since it doesn't actually do anything. We can revisit this in v5 and group it with breaking changes.

MrAlexLau:
@mshwery - sorry for the delayed reply, updated!

Issue #589, opened 2019-06-19
SelectMenu with multiselect allows selecting the same item many times with keyboard


solon-aguiar:
🐛 

- Issue and steps to reproduce.

SelectMenu: in a multi-select menu, a user can select the same element many times. `SelectMenu` has an option to allow users to select multiple elements and when users select the elements using their keyboard (e.g. just hitting the arrow down key) rather than the mouse, the same item is selected many times (one per keystroke to be precise). This is evident when the menu displays the number of selected items and that is more than the number of options.

- Versions.
Latest.

- Screenshots.
![Screen Shot 2019-06-19 at 10 12 43 AM](https://user-images.githubusercontent.com/1903101/59786188-86b2ac80-927b-11e9-94fa-5727057a1d6b.png)
(☝️ from storybook)

The console (after a modification to log the selected elements):
![Screen Shot 2019-06-19 at 10 22 16 AM](https://user-images.githubusercontent.com/1903101/59786446-22441d00-927c-11e9-8b1b-4e4e24602f76.png)

- Expected.
Each element can only be selected once.

- Actual.
Each element can be selected more than once.

- Link to your project.
https://github.com/segmentio/evergreen :)

- Browser name and version.
Chrome:  74.0.3729.169 (Official Build) (64-bit)

- Operating system.
Any

solon-aguiar:
Turns out that this is not a bug in the implementation of the component. It is just a bug in our stories and docs example. I can fix that. However, while at it, I'd like to propose a change in the way the `SelectMenu` component works for keyboard navigation.

The way that the keyboard navigation works today doesn't allow a user to select non-continuous elements. It doesn't allow for gaps in the list. For example. Let's say the options are:

```
Apple
Apricot
Banana
```

If a user, using the keyboard, wants to select just `Apple` and `Banana` (fruits that I normally put on my smoothie), there is just no way! This doesn't seem to be very useful. You can try that yourself on the storybook or in our docs example: https://evergreen.segment.com/components/select-menu.

This happens because on the `handleArrowDown` and `handleArrowUp` functions we select and de-select the components ([code](https://github.com/segmentio/evergreen/blob/master/src/select-menu/src/OptionsList.js#L168-L192)). Therefore, a user has no way to get to banana without selecting apricot. And once apricot is selected, they can only deselect by using the arrow up, which will deselect banana as well.

[...]

PR #611, opened 2019-08-24
Fix select menu keyboard navigation

solon-aguiar:
This implements the ideas described on https://github.com/segmentio/evergreen/issues/589. On MacOS `COMMAND + LEFT ARROW` or `COMMAND + RIGHT ARROW` do the navigation to the beginning and end of the search query respectively. I believe similar thing works on windows.

In summary:
- Up and down keyboard arrows for navigation between elements in the `SelectMenu`. This is a backwards incompatible change as the arrows won't select items anymore.
- Enter to select/deselect an element in the `SelectMenu`.

A video of it in action: https://recordit.co/ugChPoW5P6

### DIFF
--- a/src/select-menu/src/OptionsList.js
+++ b/src/select-menu/src/OptionsList.js
@@ -80,8 +80,8 @@ export default class OptionsList extends PureComponent {
     super(props, context)
 
     this.state = {
-      searchValue: props.defaultSearchValue,
-      selected: props.selected
+      currentIndex: 0,
+      displayedOptions: props.options
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/select-menu/src/OptionsList.js
```diff
@@ -136,72 +126,60 @@ export default class OptionsList extends PureComponent {
...
-    const { onSelect } = this.props
-    const options = this.getFilteredOptions()
-
-    let nextIndex = this.getCurrentIndex() - 1
-
-    if (nextIndex < 0) {
-      nextIndex = options.length - 1
-    }
-
-    onSelect(options[nextIndex])
```
mshwery: What happened to the `onSelect` callback?

mshwery: Ah, I think we may want to avoid breaking changes in minor/patch versions, to be strictly semver.

solon-aguiar: Yeah. Do we wait for more features before a major version bump? I think that the keyboard navigation is so bogus today that I'd push for this to be the catalyst. This change improves accessibility.

mshwery: Yeah – I think here we'd either merge into a v5 tracking branch instead of `master` if it's a known breaking change.

PR #615, opened 2019-08-27
<Dialog>: Add "press enter to submit" form behavior

nucleartide:
_Note that this pull request is a work-in-progress._

This pull request adds a "press enter to submit" form behavior to Evergreen's `<Dialog>` component.

### DIFF
--- a/src/dialog/src/Dialog.js
+++ b/src/dialog/src/Dialog.js
@@ -280,6 +280,7 @@ class Dialog extends React.Component {
       >
         {({ state, close }) => (
           <Pane
+            is="form"
             role="dialog"
             backgroundColor="white"
             elevation={4}
--- a/src/dialog/src/Dialog.js
+++ b/src/dialog/src/Dialog.js
@@ -311,6 +311,7 @@ class Dialog extends React.Component {
                     appearance="minimal"
                     icon="cross"
                     onClick={() => onCancel(close)}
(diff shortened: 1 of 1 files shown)


nucleartide:
Closing this, won't address this at the moment.

karltaylor:
I'm actually having a different issue with this at the moment.

If I have the following Dialog component, the form will not submit, however if there is only 1 TextInputField, the form will submit. Any ideas?

```
<Dialog>
  <form>
    <TextInputField />
    <TextInputField />
  </form>
</Dialog>

mshwery:
I believe this is standard html form behavior. It's part of the [html spec](https://www.w3.org/MarkUp/html-spec/html-spec_8.html#SEC8.2):

>When there is only one single-line text input field in a form, the user agent should accept Enter in that field as a request to submit the form.

mshwery:
As a workaround you can add event handlers or you can try adding a submit button/input to the form.

karltaylor:
> I believe this is standard html form behavior. It's part of the [html spec](https://www.w3.org/MarkUp/html-spec/html-spec_8.html#SEC8.2):
> 
> > When there is only one single-line text input field in a form, the user agent should accept Enter in that field as a request to submit the form.

Well shit, I've learnt something new today. Thanks @mshwery!

PR #618, opened 2019-08-28
Add optional icon to SelectMenu options

calthomson:
This PR adds an optional icon input for SelectMenu options.

Screen Shot 2019-08-28 at 2 13 58 PM 
Screen Shot 2019-08-28 at 2 13 58 PM 

These are the original designs that call for this kind of SelectMenu:
image_preview (1)

Updated the docs:
Screen Shot 2019-08-28 at 3 47 39 PM
Screen Shot 2019-08-28 at 3 47 44 PM

PR #619, opened 2019-08-29
Add placholder on FilePicker component

edumoreira1506:
The issue #531 said about need of the placeholder filepicker need to be dynamic, because the default placeholder is in english, so now the placeholder is passed from props.

### DIFF
--- a/src/file-picker/src/FilePicker.js
+++ b/src/file-picker/src/FilePicker.js
@@ -63,6 +63,10 @@ export default class FilePicker extends PureComponent {
     placeholder: PropTypes.string
   }
 
+  static defaultProps = {
+    placeholder: 'Select a file to upload…'
+  }
+
   constructor() {
     super()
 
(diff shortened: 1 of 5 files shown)


solon-aguiar:
Thank you for your contribution!

Can you please update the storybook with examples and provide some screenshots of the change?

edumoreira1506:
Sure! Here is the screenshot! But, about of the storybook, is the .storybook folder? Add a new archive or make changes in any of then file?
40cebf5f-08a9-4c78-87b7-c3f21c846021

solon-aguiar:
You can just edit the stories files and add a new example. See [this PR](https://github.com/segmentio/evergreen/pull/618) for some reference (specifically [this file](src/select-menu/stories/index.stories.js)).

I hope that helps!

edumoreira1506:
I add a example on docs :)

edumoreira1506:
I added an example on storie file, and here is one picture with example too:

src/file-picker/src/FilePicker.js
```diff
@@ -98,6 +104,13 @@ export default class FilePicker extends PureComponent {
       buttonText = 'Replace files'
     }
 
+    let placeholderText
+    if (placeholder) {
+      placeholderText = placeholder
+    } else {
+      placeholderText = 'Select a file to upload…'
+    }
```
mshwery: This could be replaced by setting a defaultProp instead.

```js
static defaultProps = {
  placeholder: 'Select a file to upload…'
  // ... other stuff
}
```

This would allow you to simply pass the prop as-is to the `<TextInput placeholder={placeholder}`

edumoreira1506: Oh sure! It's fixed now!

Issue #454, opened 2018-11-19
Link should not change font-size


ericclemmons:
As shown below, the `<Link>` element changes the `font-size` of the text, even though it's within a `<Heading>` element.

> **"But @ericclemmons, you can just use an `a` tag!"**
> -- Knee-jerk Reaction, PhD

When using a project like `mdx-js`, consistent theming relies on all `a` tags using `Link` instead, which makes this not a valid option.

> <img width="826" alt="screen shot 2018-11-19 at 1 21 30 pm" src="https://user-images.githubusercontent.com/15182/48735785-26349180-ebfe-11e8-812f-55c4a6f859c8.png">
See: https://codesandbox.io/s/32xnk54ov1

mshwery:
Evergreen doesn't currently support _cascading props_ like this. I think it would be an interesting addition, but it would also conflict with the way style props are bound.

Typically, in an example like yours, you'd have to specify the link `size` prop to match your heading's `size`.

emadabdulrahim:
@mshwery Wouldn't it be possible to set the default *Link* `fontSize="1em"` which would act the same way as `currentColor` property you guys follow in the `Icon` component?

Thoughts?

mshwery:
The thing you probably want to try here is `fontSize="inherit"` on the link component. We try to avoid inheritance/cascading styles as a rule. If just moving your component around the tree suddenly causes it to look differently that's a problem.

Since this is a much larger philosophical change that goes against the grain of how Evergreen strives for style encapsulation I am going to close the issue for now.

PR #624, opened 2019-09-10
Font size link dynamic

edumoreira1506:
Hello guys, I saw the issue #454 , now the font-size of the link is dynamic and can be sended from the parent component, like: 

```
<Link href="#" size={600}>
     Changed size font link
</Link>
```

### DIFF
--- a/src/typography/src/Link.js
+++ b/src/typography/src/Link.js
@@ -38,15 +38,21 @@ class Link extends PureComponent {
      * Class name passed to the link.
      * Only use if you know what you are doing.
      */
-    className: PropTypes.string
+    className: PropTypes.string,
+
+    /**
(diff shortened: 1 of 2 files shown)


src/typography/src/Link.js
```diff
@@ -56,6 +62,7 @@ class Link extends PureComponent {
         className={cx(className, themedClassName)}
         textDecoration="underline"
         color={null}
+        size={size}
```
mshwery: You can already pass the size prop to a link via `...props` I believe. 

What is the intended goal of this change?

edumoreira1506:
Oh, srry, I saw the issue and i though then is not possible to pass the size. I will close the PR.

Issue #521, opened 2019-03-15
Practical UI enhancements suggestions


dy:
We transitioned fully to evergreen-ui in our project, replacing material-ui and blueprint, and use it pretty heavily now.
If you let me share some practical difficulties we've faced, that would be nice to address.

* Styling toasters. Alert style does not meet app design, would be nice to be able to style them same way as tooltips: `toaster.warning('...', {appearance: "card", containerProps})`. Alternatively, 
* Positioning toasters as `toaster.warning('...', {position: Position.TOP_RIGHT})`
![image](https://user-images.githubusercontent.com/300067/54453921-b27b3f00-472e-11e9-94c6-3389cf4ba4db.png)
* Positioning for CornerDialogs (#520) or more generically, simpler dialogs for prompts/confirmations/notifications.
![image](https://user-images.githubusercontent.com/300067/54453989-d50d5800-472e-11e9-906b-1b2765a30b76.png)
* `footerProps` property for dialogs to enable buttons alignment.
![image](https://user-images.githubusercontent.com/300067/54454157-38978580-472f-11e9-8f47-da6e03e61105.png)
* `headerProps` property for dialogs to enable changing header color/font style/alignment.
![image](https://user-images.githubusercontent.com/300067/54454191-49e09200-472f-11e9-8b03-a06ce595cd3e.png)
* `hasConfirm` property to dialogs - some dialogs may not have confirmation button, like cancelling some automatic action eg. logging out / shutdown etc.
* `hasClose` property to dialogs - right now there is no way to remove close button from dialog keeping the header, ie. making dialog modal.
* `buttonProps` property for dialogs, to allow bulk-change buttons size, particularly `height`. Looks inconsistent when we have enlarged input fields but small buttons 
![image](https://user-images.githubusercontent.com/300067/54454564-1e11dc00-4730-11e9-871a-2d17310260ab.png)
* rename `statelessProps` to `containerProps` (Affects Popover, Tooltips etc.) for more intuitive/consistent naming.
* `iconLeft`, `iconRight` property for InputField (eg. we may want to add hide/disclose password, calendar icon, clear input field etc.)
![image](https://user-images.githubusercontent.com/300067/54454450-dd19c780-472f-11e9-8916-70f783784ede.png)
![image](https://user-images.githubusercontent.com/300067/54454490-f4f14b80-472f-11e9-92d5-51783816a607.png)
[...]

PR #625, opened 2019-09-11
Toast dynamic position

edumoreira1506:
Hello guys, here I am again hehe.  saw the issue #521  and I intend to do some features, I started for the dynamic position on Toaster Component. 

Before:
The toaster component always was on Top Right area:

Now:
Have 4 constants like TOP_LEFT, TOP_RIGHT, BOTTOM_RIGHT and BOTTOM_LEFT than could be used for position the toaster component.

I added 4 examples in story book, one for each position. The default position continues in top right.


mshwery:
Hi @edumoreira1506 ! Could you provide some gifs / screenshots of this in action? Thanks!

edumoreira1506:
I added some examples on storybook too!

edumoreira1506:
Peek 2019-10-24 13-43

src/toaster/stories/index.stories.js
```diff
@@ -120,6 +121,51 @@ storiesOf('toaster', module).add('examples', () => (
...
+          }
+        >
+          Right bottom
+        </Button>
+
+        <Button
+          marginRight={8}
+          onClick={() =>
+            toaster.notify('Messages now is on top left message', {
+              position: AbsolutePosition.TOP_LEFT
```
mshwery: Does it make sense for each `toaster.notify` call to require the position? I'm wondering if instead you pass a default value to Evergreen using the theme provider maybe?

edumoreira1506: I believe that pass for each `toaster.notify` it's more easy because if we pass default value to Evergreen and in any moment we need of notification in other place, it's will be not possible.

edumoreira1506:
I changed the use of constants with the same `AbsolutePosition`. I believe that now, everything is ok.

src/toaster/src/ToastManager.js
```diff
@@ -51,7 +40,8 @@ export default class ToastManager extends React.PureComponent {
     props.bindCloseAll(this.closeAll)
 
     this.state = {
-      toasts: []
+      toasts: [],
+      position: Position.TOP_RIGHT
```
mshwery: I think we want the default to still be centered at the top, e.g.

```
margin: 0 auto;
top: 0;
left: 0;
right: 0;
```

edumoreira1506:
It's ok now!

src/toaster/src/ToastManager.js
```diff
@@ -89,6 +81,12 @@ export default class ToastManager extends React.PureComponent {
     return instance
   }
 
+  checkPosition = ({ position }) => {
+    if (position) {
```
mshwery: Notably this will mean that if you set a position once, you'll have to reset the position every time you want it to be something else.

I think perhaps a nicer way to introduce this would be to only apply the position to the one instance that has the setting (and not persist the position in state). Thoughts?

astralmedia:
Am I missing something? I am following your example in the storybook, but its not working. I dug into the master branch code and I do not see anything about the position param on the settings object.

Thanks and sorry for bumping this old thread!

Issue #520, opened 2019-03-14
Position property for CornerDialog


dy:
Corner dialog is a bit different from main dialog in terms of it provides nicer default UI for prompts/confirmations. Having `position` property, similar to tooltips, would be handy.

Is there any sense in such PR?

mshwery:
Are you thinking it'd be used to change which corner the dialog is anchored to?

dy:
Yes, to make it a lightweight alternative to big dialogs.
Would be also nice to have “alignButtons” property.

dy:
I could come up with PR

PR #626, opened 2019-09-11
Dynamic position corner dialog

edumoreira1506:
More one feature from issue #521 , this time the suggest was to position the corner dialog be dynamic.

Before:
The corner dialog position was always on right bottom.

Now:
The corner dialog is totally dynamic, he could be putted in any position of screen.


src/corner-dialog/src/CornerDialog.js
```diff
@@ -143,7 +146,12 @@ export default class CornerDialog extends PureComponent {
     cancelLabel: 'Close',
     onCancel: close => close(),
     onConfirm: close => close(),
-    onCloseComplete: () => {}
+    onCloseComplete: () => {},
+    position: {
+      position: 'fixed',
+      bottom: 16,
+      right: 50
```
mshwery: curious why the change from `16` to `50` for the defaults here?

edumoreira1506: Oh, I did that just for approach the box to the bottom. Sryy, I will back to 50.

src/corner-dialog/src/CornerDialog.js
```diff
@@ -235,10 +244,8 @@ export default class CornerDialog extends PureComponent {
               width={width}
               css={animationStyles}
               data-state={state}
-              position="fixed"
-              bottom={16}
-              right={16}
               padding={32}
+              {...position}
```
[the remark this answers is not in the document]
edumoreira1506: Yeah, but there just one problem, if the user pass other properties to container and won't pass the `position`, `bottom` and `right`, the defaultProps of containerProps will be rewrite and the corner dialog won't have the reference of bottom, right and position, but yes, we can use the same variable.

edumoreira1506:
Here is one example with the default corner dialog open and the corner dialog in middle of the screen.

Screenshot from 2019-09-11 15-49-34

mshwery:
IMO it feels a bit weird to have a "CornerDialog" in the center of the screen, but maybe we can let consumers position it in the bottom-right, top-right, top-left, bottom-left. Thoughts? Otherwise using a `CornerDialog` doesnt really make much sense.

edumoreira1506:
Yes! We can use in bottom-right, top-right and other sides too. I will post some screenshots.

edumoreira1506:
Screenshot from 2019-10-24 13-37-37
Screenshot from 2019-10-24 13-38-33
Screenshot from 2019-10-24 13-38-46
Screenshot from 2019-10-24 13-38-52

edumoreira1506:
I added some examples in storybook too!

src/corner-dialog/src/CornerDialog.js
```diff
@@ -130,7 +128,12 @@ export default class CornerDialog extends PureComponent {
     /**
      * Props that are passed to the dialog container.
      */
-    containerProps: PropTypes.object
+    containerProps: PropTypes.object,
+
+    /**
+     * Props that will set position of corner dialog
+     */
+    position: PropTypes.object
```
mshwery: I think we should do like we do in other components and only take a specific set of constants instead of an object.

We will apply the right css based on that: e.g. `position={POSITIONS.TOP_LEFT}`. That way we can prevent weird scenarios like vertically and horizontally centered _corner_ dialogs, etc.

Thoughts?

src/corner-dialog/src/CornerDialog.js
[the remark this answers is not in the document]
mshwery: I would do something like `getPositionProps(this.props.position)` where position is one of `TOP_LEFT, TOP_RIGHT, BOTTOM_LEFT, BOTTOM_RIGHT`. What do you think?

src/corner-dialog/src/CornerDialog.js
edumoreira1506: I will change this!

edumoreira1506:
Now we have constants for the positions!

src/corner-dialog/src/CornerDialog.js
```diff
@@ -147,11 +148,7 @@ export default class CornerDialog extends PureComponent {
     onCancel: close => close(),
     onConfirm: close => close(),
     onCloseComplete: () => {},
-    position: {
-      position: 'fixed',
-      bottom: 50,
-      right: 50
-    }
+    position: AbsolutePosition.BOTTOM_LEFT
```
mshwery: I think we expect consumers to pass the string or constant (also a string) for `BOTTOM_LEFT`. We would then lookup the top/left/right/bottom props based on that string value, else default to `'BOTTOM_RIGHT'`.

Also the default prop should still be `BOTTOM_RIGHT`.

src/constants/src/AbsolutePosition.js
```diff
@@ -0,0 +1,18 @@
...
+    top: 0,
+    right: 0
+  },
+  BOTTOM_LEFT: {
+    bottom: 0,
+    left: 0
+  },
+  BOTTOM_RIGHT: {
+    bottom: 0,
+    right: 0
```
mshwery: I think these need to remain the same (same for the other `0` values):

```js
bottom: 50,
right: 50
```

mshwery:
Closes https://github.com/segmentio/evergreen/issues/520

src/corner-dialog/src/CornerDialog.js
```diff
@@ -134,7 +134,7 @@ export default class CornerDialog extends PureComponent {
     /**
      * Props that will set position of corner dialog
      */
-    position: PropTypes.oneOf(AbsolutePosition)
+    position: PropTypes.string
```
mshwery: This should be more like so: https://github.com/segmentio/evergreen/blob/1e2051e4ab71ae3f13331d8a841ac85bcaf6a2e6/src/popover/src/Popover.js#L13-L22

Except we only want the corners.

src/corner-dialog/src/CornerDialog.js
```diff
@@ -148,7 +148,7 @@ export default class CornerDialog extends PureComponent {
     onCancel: close => close(),
     onConfirm: close => close(),
     onCloseComplete: () => {},
-    position: AbsolutePosition.BOTTOM_LEFT
+    position: 'BOTTOM_RIGHT'
```
mshwery: We should use the same constants that we export from Evergreen – that means importing `Position` and assigning the default as `Position.BOTTOM_RIGHT` instead of a string literal here.

src/corner-dialog/src/CornerDialog.js
```diff
@@ -243,7 +242,7 @@ export default class CornerDialog extends PureComponent {
               data-state={state}
               padding={32}
               position="fixed"
-              {...position}
+              {...AbsolutePosition[position]}
```
mshwery: Nit: prefer camelCase for imports that are not components/classes.

mshwery: We also should validate that it returns something – e.g. if it's not one of the 4 supported corners we should default to `positions[Position.BOTTOM_RIGHT]`

src/corner-dialog/stories/index.stories.js
```diff
@@ -157,7 +156,7 @@ storiesOf('corner-dialog', module).add('CornerDialog', () => (
             width={492}
             confirmLabel="View Terms"
             onCloseComplete={() => setState({ isShown: false })}
-            position={AbsolutePosition.BOTTOM_LEFT}
```
mshwery: Prefer the existing constants that Evergreen exports (`Position`)

src/constants/src/AbsolutePosition.js
```diff
@@ -1,18 +1,18 @@
...
+    top: 50,
+    right: 50
   },
   BOTTOM_LEFT: {
-    bottom: 0,
-    left: 0
+    bottom: 50,
+    left: 50
   },
   BOTTOM_RIGHT: {
```
mshwery: We probably need to define these using the `Position` constants that Evergreen exports so they are always consistent.

e.g. `[Position.BOTTOM_RIGHT]: {`

src/corner-dialog/src/CornerDialog.js
edumoreira1506: We have this on default props. And with `oneOf[...]`

mshwery: Right, but you could still pass invalid values – and just get warned.

We need it to always pick a valid value – so if there's a value passed that is incorrect we should either `throw` an error or gracefully fall back to the default.

edumoreira1506:
I believe than now is everything ok, right?

mshwery:
Can you rebase this on top of latest `master`?

PR #628, opened 2019-09-11
Text input field icon

edumoreira1506:
Issue #521 

Before:
The text input field was simple and did not have icon.

Now:
The text input field can have a icon and set the position, like right or left and how many pixels far from the edge.

Have some examples on storybook.

### DIFF
--- a/src/text-input/src/TextInputField.js
+++ b/src/text-input/src/TextInputField.js
@@ -58,7 +58,7 @@ export default class TextInputField extends PureComponent {
     /**
      * Optional icon in input
      */
-    icon: PropTypes.shapce({
+    icon: PropTypes.shape({
       iconPosition: PropTypes.object,
       content: PropTypes.element
     })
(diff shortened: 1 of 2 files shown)


edumoreira1506:
Sure! Here have some examples.
Screenshot from 2019-09-11 15-53-13

mshwery:
@edumoreira1506 what happens when you click on the icon? Does it focus the input? Does it bubble the same events?

I'm wondering if it makes more sense to push the icon behavior into the `TextInput` instead?

edumoreira1506:
This PR just add the icon on text input field. When you click on the icon nothing happensg, like any other input with icon, for just the icon help to describe what the input need. Like the example:

Screenshot from 2019-10-24 13-49-21

mshwery:
IMO since it looks like it is _inside_ the input box, it should trigger focus on the input. That is to say, we should probably move this into the `TextInput` component so we get this behavior for free.

Issue #620, opened 2019-09-03
Tooltip prop request: showDelay


jasonbarry:
I'm running into a few places in my app where it would be useful to have a tooltip only display after x milliseconds. A prop like `showDelay` would make it so that the tooltip would begin to show after the passed time had elapsed (similar to `hideDelay`).

PR #630, opened 2019-09-12
Add showDelay prop to Tooltip

auderer:
This PR adds a `showDelay` prop to `Tooltip`, which takes a value in milliseconds (default 0). The tooltip will only show if the target is hovered for this length of time. If the mouse enters then leaves before the delay is up, the tooltip will not show.

I also added an additional tooltip button in Storybook to demonstrate the delay.

This feature was requested in #620. I believe this is a useful feature - it's common for apps to have text or buttons that require a long hover before showing extra info.

example tooltip delay

### DIFF
--- a/src/tooltip/src/Tooltip.js
+++ b/src/tooltip/src/Tooltip.js
@@ -39,6 +39,11 @@ export default class Tooltip extends PureComponent {
     hideDelay: PropTypes.number.isRequired,
 
     /**
+     * Time in ms before showing the Tooltip.
+     */
+    showDelay: PropTypes.number.isRequired,
+
+    /**
      * When True, manually show the Tooltip.
      */
     isShown: PropTypes.bool,
@@ -57,7 +62,8 @@ export default class Tooltip extends PureComponent {
   static defaultProps = {
     appearance: 'default',
     position: Position.BOTTOM,
-    hideDelay: 120
+    hideDelay: 120,
+    showDelay: 0
   }
 
   constructor(props, context) {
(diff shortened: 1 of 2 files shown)


src/tooltip/src/Tooltip.js
```diff
@@ -80,14 +87,28 @@ export default class Tooltip extends PureComponent {
...
 
   hide = () => {
-    if (!this.state.isShown) return
+    if (!this.state.isShown) {
+      if (!this.state.willShow) return
+      this.setState({
+        willShow: false
+      })
+      return
+    }
```
mshwery: I think we could probably simplify this? Looks like every result of this function ends with `isShown: false, willShow: false` right?

```js
// nothing to do
if (!this.state.isShown && !this.state.willShow) {
  return
}

// Otherwise, set both the current and eventual shown states to false
this.setState({
  isShown: false,
  willShow: false
})
```

You could also do this with the same effect:
```js
hide = () => {
  this.setState({
    isShown: false,
    willShow: false
  })
}
```

It'll end up calling setState in more cases, but the resulting state will not change from what the current code is doing.

auderer: That's a great point! I added a commit to simplify it 👍 

I went with the second solution of just setting state every time `hide` is called. It's not likely for `hide` to be called while `!this.state.isShown && !this.state.willShow`, since the mouse will always enter before leaving. (unless the user rapidly flails their mouse over the tooltip, but even that is fine because we're using PureComponent)

solon-aguiar:
Thank you for the change!

PR #631, opened 2019-09-16
Update downshift to 3.3.1

### LINKED ISSUE #450: Combobox does not scroll to selected item

tusbar:
This is #523, reopened.

It aliases and deprecates removed props from Downshift.

This still fixes #450 by creating a `stateReducer` that will set the `highlightedItem` to the selected item when opening the autocomplete – as suggested by @kentcdodds here: https://github.com/downshift-js/downshift/issues/645#issuecomment-451805753.

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -147,6 +147,22 @@ export default class Autocomplete extends PureComponent {
     })
   }
 
+  stateReducer = (state, changes) => {
+    const { items } = this.props
+
+    if (
+      Object.prototype.hasOwnProperty.call(changes, 'isOpen') &&
+      changes.isOpen
+    ) {
+      return {
+        ...changes,
+        highlightedIndex: items.indexOf(state.selectedItem)
+      }
+    }
+
+    return changes
+  }
+
   renderResults = ({
     width,
     inputValue,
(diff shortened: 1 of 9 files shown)


src/autocomplete/src/Autocomplete.js
```diff
@@ -48,9 +48,22 @@ export default class Autocomplete extends PureComponent {
     selectedItem: PropTypes.any,
 
     /**
-     * The selected item to be selected & shown by default on the autocomplete
+     * The selected item to be selected & shown by default on the autocomplete (deprecated)
+     * @return {Error} a deprecation warning
      */
-    defaultSelectedItem: PropTypes.any,
+    defaultSelectedItem: () =>
+      new Error(
```
mshwery: Should we adopt the same pattern seen elsewhere: https://github.com/thefrontside/deprecated-prop-type/blob/master/deprecated.js#L5

tusbar: Never seen that. 

I’m a little confused as to why you would want to validate a deprecated prop’s type.

Here using `defaultSelectedItem` will always display an error in your console – whatever type you pass to it. Then `initialSelectedItem`’s prop-type from `Downshift` will warn you if your type is invalid.

But if you want to use that, I’ll change. :)

mshwery: I think it's more about only showing the deprecation warning once per message.

tusbar:
@mshwery I’ve added a `lib/deprecated.js` utility that uses the logic already in `lib/warning.js` to display the warning only once.

Tell me what you think.

tusbar:
@mshwery to my knowledge, no other prop has been renamed and/or deleted from Downshift.

I’ve tested this on a few apps of my own. It shows the warnings properly and everything still works the same.

size-limit bumped to 210KB.

PR #635, opened 2019-09-28
Change icon prop-types from string to node

### LINKED ISSUE #535: Allow component as props for Icon
This is a minor issue but sometimes when using Buttons I pass an `Image` for `iconBefore` like so:

```
<Button
  onClick={() => {}}
  iconBefore={(
    <Image src="some-icon.png" />
  )}
>
  blah blah...
</Button>
```

It's useful because I can specify anything I want - but I get lots of propType warnings since Icon is expecting a string - maybe we can relax this condition a bit?

auderer:
This PR simply switches the prop-types for Icon (and Button icons) from `string` to `node`, which allows both strings and custom elements. Fixes #535. 

Since any valid node is correctly rendered by icon props, the PropTypes for these icons should not be restricted to strings. 

It's great to have this option for implementing custom icons with images or even FontAwesome, which already works great in spite of the PropTypes warnings.

### DIFF
--- a/src/buttons/src/Button.js
+++ b/src/buttons/src/Button.js
@@ -52,12 +52,12 @@ class Button extends PureComponent {
     isActive: PropTypes.bool,
 
     /**
-     * Sets an icon before the text. Can be any icon from Evergreen.
+     * Sets an icon before the text. Can be any icon from Evergreen or a custom element.
      */
     iconBefore: PropTypes.node,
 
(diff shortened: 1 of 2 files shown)


eoinmurray:
This is great - when would it be released?

mshwery:
It's released under v4.19.2 @eoinmurray

PR #638, opened 2019-09-30
Improve EditableCell behavior

### LINKED ISSUE #637: EditableCell behavior improvements
I love the Editable Table, but there are some behaviors that feel like bugs, especially in comparison to most spreadsheet software:

- [ ] Cell textarea opens when Shift is pressed (Expected: Textarea should not open with Shift)
- [ ] Caret position starts at beginning of textarea (Expected: Caret should start at end of text)
- [ ] When focusing a cell then pressing a letter, the letter is appended to the end of the text (Expected: pressed letter should replace the existing cell text)
- [ ] When focusing a cell then pressing a number, nothing happens (Expected: The textarea should open, replacing existing text with the pressed number)

These expectations are based on my personal UX expectations and verified against the behavior on http://sheets.google.com.

auderer:
This PR fixes #637, which states the following problems in editable table cells:

- Cell textarea opens when Shift is pressed (Expected: Textarea should not open with Shift)
- Caret position starts at beginning of textarea (Expected: Caret should start at end of text)
- When focusing a cell then pressing a letter, the letter is appended to the end of the text (Expected: pressed letter should replace the existing cell text)
- When focusing a cell then pressing a number, nothing happens (Expected: The textarea should open, replacing existing text with the pressed number)

### DIFF
--- a/src/table/src/EditableCellField.js
+++ b/src/table/src/EditableCellField.js
@@ -65,6 +65,8 @@ export default class EditableCellField extends React.PureComponent {
 
     requestAnimationFrame(() => {
       this.textareaRef.focus()
+      // Move cursor to end of cell content
+      this.textareaRef.selectionStart = this.textareaRef.value.length
     })
   }
 
--- a/src/table/src/EditableCell.js
+++ b/src/table/src/EditableCell.js
@@ -16,9 +16,9 @@ class EditableCell extends React.PureComponent {
     ...TableCell.propTypes,
 
     /*
-    * Makes the TableCell focusable.
-    * Will add tabIndex={-1 || this.props.tabIndex}.
-    */
+     * Makes the TableCell focusable.
+     * Will add tabIndex={-1 || this.props.tabIndex}.
+     */
     isSelectable: PropTypes.bool.isRequired,
 
     /**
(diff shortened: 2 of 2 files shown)


mshwery:
@calthomson @hellooimkat thoughts on these changes?

I think they may be significant enough to classify them as "breaking" changes (requires a major version bump of `evergreen-ui`), but since y'all have heavily used editable cells I'm curious if we had the current behavior for a reason?

calthomson:
@mshwery Thanks for looping us in - these changes look alright to me, but I'm testing them out in the Tracking Plan editor now to be sure.

src/table/src/EditableCellField.js
```diff
@@ -65,6 +65,8 @@ export default class EditableCellField extends React.PureComponent {
 
     requestAnimationFrame(() => {
       this.textareaRef.focus()
+      // Move cursor to end of cell content
+      this.textareaRef.selectionStart = this.textareaRef.value.length
```
mshwery: You know what would be even better? Inserting the cursor exactly where they clicked :)

SwenSchaeferjohann:
Any reference to why `editableCell` is not part of the documentation?

Issue #639, opened 2019-10-02
Popover statelessProps props


rafaelcamaram:
I was taking a look at the `PopOver` component and I noticed that when I pass a value for the `statelessProps` props - with a `style` object - it'll start to override and lost all the previous `statelessProps` props.

Do you know if there is a way to keep these old properties that I'm not passing? Or is it necessary to create a PR for it?

PR #640, opened 2019-10-02
Fixing stateless props override on Popover component

rafaelcamaram:
Fixing issue reported on: https://github.com/segmentio/evergreen/issues/639

I was taking a look at the `PopOver` component and I noticed that when I pass a value for the `statelessProps` props - with a style object - it'll start to override and lost all the previous style properties.

It's because it was just spreading the `statelessProps` into the component and overriding all the default styling values.

So my change was basically to merge the `statelessProps.style` with the current `style` into one single `style` props.

So, I can do something like this without losing any default style props.
```js
<Popover
          ...
          statelessProps={{
            style: {
              zIndex: 1000
            }
          }}
         ...
/>
```

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -87,7 +87,7 @@ export default class Popover extends Component {
      * Function that will be called when the exit transition is complete.
      */
     onCloseComplete: PropTypes.func.isRequired,
-    
+
     /**
      * Function that will be called when the body is clicked.
      */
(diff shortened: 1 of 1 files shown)


rafaelcamaram:
@jeroenransijn Can you review it and verify if it makes sense? I'm using this component on my production project and we're needing it as soon as possible

rafaelcamaram:
@mshwery Thank you for approving this PR. Do you know when it will be available on NPM (deployed)?

mshwery:
Now! https://github.com/segmentio/evergreen/releases/tag/v4.19.2

PR #641, opened 2019-10-04
Use theme

Pasalietis:
Just add useTheme hook


mshwery:
@Pasalietis won't we need to bump our minimum supported version of react?
https://github.com/segmentio/evergreen/blob/master/package.json#L63-L64

PR #643, opened 2019-10-04
Adding overlayProps to the Dialog component in order to allow some cu…

rafaelcamaram:
It's a similar change I've on the merged PR (https://github.com/segmentio/evergreen/pull/640) but on the `Dialog` component at this time.

Currently the `Overlay` component - inside the `Dialog` - only have the following `containerProps`:

```
containerProps={{
   display: 'flex',
   alignItems: 'flex-start',
   justifyContent: 'center',
}}
```

So I added a new props called `overlayProps` in order to allow us to make more customisations with it.

So, in this way, we're going to keep the default values to `containerProps` and override or add more  when it's passed.

### DIFF
--- a/src/dialog/src/Dialog.js
+++ b/src/dialog/src/Dialog.js
@@ -38,14 +38,10 @@ const closeAnimation = css.keyframes('closeAnimation', {
 
 const animationStyles = {
   '&[data-state="entering"], &[data-state="entered"]': {
-    animation: `${openAnimation} ${ANIMATION_DURATION}ms ${
-      animationEasing.deceleration
-    } both`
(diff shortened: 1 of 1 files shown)


rafaelcamaram:
@mshwery Matt, if it's possible, please take a look on it :)

rafaelcamaram:
@mshwery Can you deploy it for me? We're needing it :)

drummerjolev:
Has this been deployed? I need to modify the `zIndex` on a Dialog and this would be very useful.

mshwery:
Looks like this should be available as of v4.19.3 and above.

PR #644, opened 2019-10-04
Disable default scrollIntoView behavior

tusbar:
#631 introduces a small issue with `scrollIntoView` where the page would scroll to display the items list. This is already handled by the `Popover` being displayed above or below the input, depending on the available space.

Before:

existing

After:

new

This simply disables the `scrollIntoView` default behavior.


tusbar:
Can we merge this? This issue is actually pretty annoying.

PR #646, opened 2019-10-09
Types for components

Pasalietis:
This is types for components only. So there is missing types for theme and etc

 Also list of components with props = Record<string, any>

- [ ] InlineAlert
- [ ] AutocompleteItem
- [ ] Badge
- [ ] Pill
- [ ] BackButton
- [ ] TextDropdownButton
- [ ] Combobox
- [ ] StackingOrder
- [ ] Intent
- [ ] Position
- [ ] CornerDialog
- [ ] Dialog
- [ ] FilePicker
- [x] IconNames
- [ ] Image
- [ ] Overlay
- [ ] Portal
- [x] minorScale
- [x] majorScale
- [ ] Select
- [ ] SelectField
- [ ] OptionShapePropType
- [x] OptionsList
- [ ] SelectedPropType
- [x] SelectMenuContent
- [ ] Switch
- [x] extractStyles
- [ ] StackingContext
- [x] Table
- [x] TableHead
- [x] TableHeaderCell
- [x] TextTableHeaderCell
- [x] SearchTableHeaderCell
- [x] TableBody
- [x] TableRow
- [x] TableCell
- [x] TextTableCell
- [ ] TagInput
- [ ] ThemeContext
- [ ] ThemeProvider
- [ ] ThemeConsumer
- [ ] withTheme
- [ ] defaultTheme
- [ ] Textarea
- [ ] toaster
- [ ] UnorderedList
- [ ] Ul
- [ ] OrderedList
- [ ] Ol
- [ ] ListItem
- [ ] Li
- [ ] Heading
- [ ] Code
- [ ] Pre
- [ ] Link
- [ ] Small
- [ ] Strong

Todo:
- [ ] move types: https://github.com/segmentio/evergreen/pull/646#pullrequestreview-301547974
- [ ] wait for https://github.com/segmentio/evergreen/pull/649 will be accepted


index.d.ts
```diff
@@ -0,0 +1,611 @@
+/* tslint:disable:interface-name max-classes-per-file no-empty-interface */
+
+declare module 'evergreen-ui' {
+  import * as React from 'react'
```
mshwery: Can this be `import React from 'react'`?

index.d.ts
```diff
@@ -0,0 +1,611 @@
...
+  import { EnhancerProps } from 'ui-box/dist/types/enhancers'
+
+  interface UIBoxProps extends Partial<EnhancerProps> {
+    innerRef?: React.Ref<HTMLElement>;
+    children?: React.ReactNode;
+    height?: string | number;
+    width?: string | number;
+    css?: object;
+  }
```
mshwery: I think we might be able to get rid of this by using `BoxProps<T>` for specific components that accept box props in addition to their own props API.

index.d.ts
```diff
@@ -0,0 +1,611 @@
...
+
+  type PositionTypes = 'top' | 'top-left' | 'top-right' | 'bottom' | 'bottom-left' | 'bottom-right' | 'left' | 'right';
+  type IntentTypes = 'none' | 'success' | 'warning' | 'danger';
+  type IconNameTypes =
```
mshwery: We can outsource these! :)

```ts
import { IconName } from '@blueprintjs/icons'

export type IconName = IconName
```

index.d.ts
```diff
@@ -0,0 +1,611 @@
...
+    | 'person'
+    | 'import'
+    | 'predictive-analysis'
+    | 'refresh'
+    | 'export';
+
+  type RemovedAttributes = 'color' | 'title' ;
+  type ElementProps = Omit<React.HtmlHTMLAttributes<HTMLElement>, RemovedAttributes>;
+  type ElementInputProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, RemovedAttributes>;
+  type ElementButtonProps = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, RemovedAttributes>;
```
mshwery: Why do we need to omit these props?

Also can we write this as `Omit<React.HTMLAttributes<HTMLElement>, RemovedAttributes>` – just not sure what `HtmlHTMLAttributes` gets us (it adds an optional `manifest` attribute...)

mshwery: Curious if it would be better for us to omit these properties only on the components that expose those same prop names?

index.d.ts
```diff
@@ -0,0 +1,611 @@
...
+
+  export interface AlertProps extends PaneProps, ElementProps {
+    intent: IntentTypes;
+    title?: React.ReactNode;
+    hasTrim?: boolean;
+    hasIcon?: boolean;
+    isRemoveable?: boolean;
+    onRemove?: () => void;
```
mshwery:
```suggestion
    onRemove?: (event: React.MouseEvent<HTMLButtonElement> | React.TouchEvent<HTMLButtonElement>) => void;
```

index.d.ts
```diff
@@ -0,0 +1,611 @@
...
+    appearance?: 'default' | 'card';
```
mshwery: Maybe we should type this: `type Appearance = 'default' | 'card'`

index.d.ts
Pasalietis: Nope.
TS1259: Module '"./evergreen/node_modules/@types/react/index"' can only be default-imported using the 'esModuleInterop' flag

index.d.ts
Pasalietis: Do you mean:
```TS
  export interface CardProps extends PaneProps {
  }

  export class Card extends React.PureComponent<CardProps & BoxProps<'div'>> {
  }
```
 Because this way innerRef is broken and passes `'div' | null` to callback

index.d.ts
Pasalietis: This Props and types was a work around for various React/ui-box props which conflicts with component props. If the use `BoxProps<T>` then ref still broken, but conflicts will be avoided and props a no longer nessary.
Unless you can provide better sniped then i can figured out

index.d.ts
mshwery: We should fix `innerRef` in ui-box 😅 

https://github.com/segmentio/ui-box/pull/41

mshwery: @Pasalietis there's a new release for ui-box@2.1.2 that has proper types for `innerRef`

Pasalietis: @mshwery can you look again if now is correct approach for using `UiBox<T>` ?

index.d.ts
```diff
@@ -0,0 +1,790 @@
...
+    popoverMaxHeight?: number;
+    selectedItem?: any;
+    buttonProps?: ButtonProps;
+    onChange: (selectedItem: any) => void;
+  }
+
+  export class Autocomplete extends React.PureComponent<AutocompleteProps> {
+  }
+
+  export interface AvatarProps extends BoxProps<'div'> {
```
mshwery: This looks right as a starting point. I think it'll produce the wrong type for when you pass a custom `is` value, but for that to work we'd need to do something like `React.ComponentProps<typeof Box>` but I've tried that and it won't play nicely with the prop generic inference we have going on with `is`.

Pasalietis:
> Can you make sure we include the file path for these types in the `package.json`?

do you mean just add
`"types": "index.d.ts"` ?

pearlyplan:
@mshwery @Pasalietis anything I can do to help this along?  I've been improving my local typings  and will contribute my edits as soon as this is up!

Pasalietis:
> @mshwery @Pasalietis anything I can do to help this along? I've been improving my local typings and will contribute my edits as soon as this is up!

I need help with setting types in package.json

package.json
```diff
@@ -17,6 +17,7 @@
   "license": "MIT",
   "main": "commonjs/index.js",
   "module": "esm/index.js",
+  "types": "index.d.ts",
```
mshwery: @Pasalietis added this here :)

package.json
```diff
@@ -44,6 +45,7 @@
   "dependencies": {
     "@babel/runtime": "^7.1.2",
     "@blueprintjs/icons": "^3.2.0",
+    "@types/react": "^16.9.5",
```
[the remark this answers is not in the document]
Pasalietis: ¯\_(ツ)_/¯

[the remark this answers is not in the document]
mshwery: Looks like the TypeScript docs suggest `dependencies` though I can see that getting a bit tricky: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

We can start with it here and see if any people run into issues!

pearlyplan:
Thanks so much for pushing this through everyone, makes the library much more usable!

seanemmer:
Thanks so much for pushing this through everyone, makes the library much more usable!

PR #647, opened 2019-10-10
fix:closeOnSelect must be destructed from props

Nikhil-Pavan-Sai:
This closeOnSelect must be declared separately in const{options}


mshwery:
🤔 https://github.com/segmentio/evergreen/pull/616

PR #649, opened 2019-10-10
Updating lib ui-box@^2.1.1

Pasalietis:
Breaking Changes:
css is no longer an accepted prop for any component (except `<Pane />`) – it won't interpret that prop and will pass it through to the underlying component.


Pasalietis:
Is this size limit is hard prerequisite or it can be raised?

src/tooltip/src/Tooltip.js
```diff
@@ -199,13 +200,13 @@ export default class Tooltip extends PureComponent {
...
       >
-        {({ css, style, state, getRef }) => (
+        {({ className, style, state, getRef }) => (
           <TooltipStateless
             id={this.state.id}
             appearance={appearance}
             innerRef={ref => getRef(ref)}
             data-state={state}
-            css={css}
+            className={cx(statelessClassName, className)}
```
mshwery: Can we push this prop change into `TooltipStateless` instead? That way we can maintain the `css` prop api here for this version without breaking changes.

src/corner-dialog/src/CornerDialog.js
```diff
@@ -233,7 +234,7 @@ export default class CornerDialog extends PureComponent {
               backgroundColor="white"
               elevation={4}
               width={width}
-              css={animationStyles}
```
mshwery: Seems like we could avoid this change if we pushed it into `Pane` and maybe `Text`. Anywhere that applies `css` to a `Box` component _directly_.

src/tooltip/src/Tooltip.js
Pasalietis: I updated this file to be more closely to https://github.com/segmentio/evergreen/pull/549

src/corner-dialog/src/CornerDialog.js
Pasalietis: Doesn't evergreen should follow ui-box way to drop support of css prop ?

mshwery: Yeah. I think we should mark it as deprecated so devs know to move off of it. Then in the next major version we can remove it completely.

src/switch/src/Switch.js
```diff
@@ -7,12 +8,12 @@ const animationEasing = {
   spring: `cubic-bezier(0.175, 0.885, 0.320, 1.175)`
 }
 
-const handleStyle = {
+const handleStyleClass = {
```
tusbar: Shouldn’t this call `css({ … })`?

src/popover/src/Popover.js
```diff
@@ -372,11 +374,11 @@ export default class Popover extends Component {
               getRef(ref)
             }}
             data-state={state}
-            css={css}
             display={display}
             minWidth={minWidth}
             minHeight={minHeight}
             {...statelessProps}
+            className={cx(statelessProps.className, gcss(css))}
```
mshwery: This might produce `css-nil` class? Do we want `css ? gcss(css) : undefined`?

mshwery:
>Is this size limit is hard prerequisite or it can be raised?

@Pasalietis we can bump it up a tiny bit if needed. Ideally we find other ways to cut it back but with this type of change it makes sense. What's the size limit change?

Pasalietis:
> 
> 
> > Is this size limit is hard prerequisite or it can be raised?
> 
> @Pasalietis we can bump it up a tiny bit if needed. Ideally we find other ways to cut it back but with this type of change it makes sense. What's the size limit change?

```bash
Package size limit has exceeded by 123 B
Package size: 210.12 KB
Size limit:   210 KB
```

src/popover/src/Popover.js
Pasalietis: I look at it

src/switch/src/Switch.js
Pasalietis: fixed

src/popover/src/Popover.js
Pasalietis: fixed

mshwery:
@Pasalietis go ahead and bump the size limit by 1-2 kb :)

tusbar:
Why not update to 3.0.0 while we’re at it?

Pasalietis:
> 
> 
> Why not update to 3.0.0 while we’re at it?

because 3.0.0 have more breaking changing

mshwery:
Yeah I think if we moved to ui-box@^3.0.0 we'll need to major version bump evergreen as well since the classnames will change.

mshwery:
Closes #469

docs/src/components/DocsMDXProvider.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { MDXProvider } from '@mdx-js/tag'
+import { MDXProvider } from '../../node_modules/@mdx-js/react'
```
mshwery: This should just be `@mdx-js/react`

mshwery:
so close @Pasalietis !! How can I help you get this over the line?

docs/src/components/DocsMDXProvider.js
Pasalietis: fixed

Pasalietis:
> so close @Pasalietis !! How can I help you get this over the line?

i guess it's done now.

PR #660, opened 2019-10-30
add more types and docs!

mshwery:
This PR introduces more types and some documentation

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -589,6 +589,23 @@ declare module 'evergreen-ui' {
   export class IconButton extends React.PureComponent<IconButtonProps> {
   }
 
+  export interface InlineAlertProps extends PaneProps {
+    intent?: IntentTypes
+
+    /**
+     * When true, show a icon on the left matching the type,
+     */
+    hasIcon?: boolean
+
+    /**
+     * The size of the Text.
+     */
+    size?: keyof Typography['text']
+  }
+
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -4,13 +4,14 @@ declare module 'evergreen-ui' {
   import { IconName } from '@blueprintjs/icons'
   import * as React from 'react'
   import { extractStyles as boxExtractStyles } from 'ui-box'
-  import { BoxProps } from 'ui-box/dist/types/box-types'
+  import { BoxProps, Is } from 'ui-box/dist/types/box-types'
```
Rowno: Maybe `ui-box` should properly export these types so you can just do:
```
import { BoxProps, Is } from 'ui-box'
```

index.d.ts
```diff
@@ -120,12 +522,37 @@ declare module 'evergreen-ui' {
   export class Card extends React.PureComponent<CardProps> {
   }
 
+  export interface CodeProps extends TextProps<'code'> {
+
+  }
```
Rowno: Might be nicer to use `type CodeProps = TextProps<'code'>`?

index.d.ts
mshwery: Yes! Would be better.

index.d.ts
mshwery: Ok!

PR #665, opened 2019-11-01
update types based on breakages in the app

mshwery:
I tried switching our app to use the evergreen types and these were the necessary changes to match.

Unfortunately we can't use `BoxProps<T>` because of how the conditional type inference works on  passthrough props. :( Doing so essentially breaks using the `is` prop.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,9 @@ declare module 'evergreen-ui' {
   import { IconName } from '@blueprintjs/icons'
   import * as React from 'react'
   import { extractStyles as boxExtractStyles } from 'ui-box'
+  import Box from 'ui-box'
   import { BoxProps, Is } from 'ui-box/dist/types/box-types'
+  import { DownshiftProps, GetInputPropsOptions } from 'downshift'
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1406,16 +1410,16 @@ declare module 'evergreen-ui' {
   export class SegmentedControl extends React.PureComponent<SegmentedControlProps> {
   }
 
-  export interface SelectProps extends Omit<BoxProps<'div'>, 'onChange'> {
+  export interface SelectProps extends Omit<React.ComponentPropsWithoutRef<typeof Box>, 'onChange'> {
     /**
      * The initial value of an uncontrolled select
      */
-    defaultValue?: string
+    defaultValue?: string | number | string[]
```
drew-thompson: Could this also be a `number[]`?

index.d.ts
```diff
@@ -1406,16 +1410,16 @@ declare module 'evergreen-ui' {
...
 
     /**
      * The value of the select.
      */
-    value?: string
+    value?: string | number | string[]
```
drew-thompson: Same here with the `number[]` question.

index.d.ts
mshwery: I grabbed these types from React itself, so Imma trust their types :)

PR #667, opened 2019-11-02
Fix custom className not applied to Tooltip components

### LINKED ISSUE #666: Tooltip not applying custom "className" property
## Description

[`<Tooltip />`](https://evergreen.segment.com/components/tooltip/) allows user to specify properties passed through to the tooltip via the `statelessProps` prop.

However since [v4.21.0](https://github.com/segmentio/evergreen/releases/tag/v4.21.0), the component doesn't seem to be handling the `className` property correctly (i.e. `className` property passed to `props.statelessProps` isn't being applied to the Tooltip component)

## Cause

This bug seems to be introduced in https://github.com/segmentio/evergreen/commit/be35cad9fe447cf38c2d0b1a8196e838c9ee06eb#diff-484772c4ad03c20c8d908b4c643e0b9aR215 where the component references `statelessProps.classNames` instead of `statelessProps.className` (the former is an invalid property).

## Proposed Solution

Modify [src/tooltip/src/Tooltip.js](https://github.com/segmentio/evergreen/blob/v4.21.1/src/tooltip/src/Tooltip.js#L215) to use `statelessProps.className` instead

tohjustin:
This PR fixes #666, which modifies [src/tooltip/src/Tooltip.js](https://github.com/segmentio/evergreen/blob/v4.21.1/src/tooltip/src/Tooltip.js#L215) to use `statelessProps.className` instead of `statelessProps.classNames`.

### DIFF
--- a/src/tooltip/src/Tooltip.js
+++ b/src/tooltip/src/Tooltip.js
@@ -212,7 +212,7 @@ export default class Tooltip extends PureComponent {
             onMouseLeave={this.handleMouseLeaveTarget}
             {...statelessProps}
             className={cx(
-              statelessProps.classNames,
+              statelessProps.className,
               css ? gcss(css).toString() : undefined
             )}
           >


tohjustin:
Thanks for the quick review! 🙏

PR #670, opened 2019-11-06
fix missing animations due to premature glamor classname-ifying

### LINKED ISSUE #669: Side Sheet animation broken
It looks like the Side Sheet animation broke in the latest version.  The close button still animates but the sheet itself does not.  This can be seen on the docs site:

https://evergreen.segment.com/components/side-sheet

mshwery:
When we migrated off of `ui-box`'s `css` prop, we made the mistake of merging styles into a classname too early in a few cases (where more styles need to be incorporated). This leads to style overrides that don't quite match up with expectations it seems. Plumbing the `css` prop down to components like Pane and Text allows us the ability to more safely merge those styles.

Fixes #669

### DIFF
--- a/src/typography/src/Text.js
+++ b/src/typography/src/Text.js
@@ -1,3 +1,5 @@
+import cx from 'classnames'
+import { css as gcss } from 'glamor'
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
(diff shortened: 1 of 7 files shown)


src/typography/src/Text.js
```diff
@@ -51,6 +62,7 @@ class Text extends PureComponent {
         fontFamily={theme.getFontFamily(fontFamily)}
         marginTop={finalMarginTop}
         {...textStyle}
+        className={cx(className, css ? gcss(css).toString() : undefined)}
```
mshwery: Because we were supporting pass-through props to `Box` we now need to grab `css` ourselves and convert it to a classname as this point.

Technically there could be other places that are impacted as well, but I hopefully covered them by converting `Text` (`Pane` is already converted to support this scenario).

src/typography/src/Text.js
```diff
@@ -1,3 +1,5 @@
+import cx from 'classnames'
+import { css as gcss } from 'glamor'
```
vadimdemedes: Can you use a more descriptive name, e.g. `glamorCss`?

PR #680, opened 2019-11-19
Adding custom CSS (open/close) animations to the Dialog component

rafaelcamaram:
# Dialog - Adding custom CSS (open/close) animations 🎉

With this PR I'm aiming to add to the Dialog component the support to add custom animations. So, as you can see on the storybook, we can open/close it in different ways.

I hope it gets approved. We're needing it on our current project.

Thank you :)

**@Edit**
GIF

_Code:_
```jsx
<Dialog
        isShown={state.isShown}
        animationDuration={400}
        openAnimation={{
          from: {
            transform: 'translateX(1000px)',
            opacity: 0
          },
          to: {
            transform: 'translateX(0)',
            opacity: 1
          }
        }}
        closeAnimation={{
          from: {
            transform: 'translateX(0)',
            opacity: 1
          },
          to: {
            transform: 'translateX(1000px)',
            opacity: 0
          }
        }}
      >
        You're using custom CSS animations.
      </Dialog>
```

### DIFF
--- a/src/dialog/src/Dialog.js
+++ b/src/dialog/src/Dialog.js
@@ -12,40 +12,32 @@ const animationEasing = {
   acceleration: `cubic-bezier(0.4, 0.0, 1, 1)`
 }
 
-const ANIMATION_DURATION = 200
+const DEFAULT_ANIMATION_DURATION = 400
 
-const openAnimation = css.keyframes('openAnimation', {
-  from: {
-    transform: 'scale(0.8)',
-    opacity: 0
-  },
-  to: {
-    transform: 'scale(1)',
-    opacity: 1
-  }
-})
+const generateOpenAnimation = animation => {
+  return css.keyframes('openAnimation', animation)
+}
 
-const closeAnimation = css.keyframes('closeAnimation', {
-  from: {
-    transform: 'scale(1)',
-    opacity: 1
-  },
-  to: {
-    transform: 'scale(0.8)',
-    opacity: 0
-  }
-})
+const generateCloseAnimation = animation => {
+  return css.keyframes('closeAnimation', animation)
+}
 
-const animationStyles = {
-  '&[data-state="entering"], &[data-state="entered"]': {
-    animation: `${openAnimation} ${ANIMATION_DURATION}ms ${
-      animationEasing.deceleration
-    } both`
-  },
-  '&[data-state="exiting"]': {
-    animation: `${closeAnimation} ${ANIMATION_DURATION}ms ${
-      animationEasing.acceleration
-    } both`
+const animationStyles = (
+  openAnimation,
+  closeAnimation,
+  animationDuration = DEFAULT_ANIMATION_DURATION
+) => {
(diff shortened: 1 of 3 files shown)


mshwery:
Can you include some gifs of it in action?

rafaelcamaram:
@mshwery Sure! I just edited the first comment with it. It's important to notice that it can be customizable as you want..

src/dialog/src/Dialog.js
```diff
@@ -299,7 +331,11 @@ class Dialog extends React.Component {
             marginY={topOffsetWithUnit}
             display="flex"
             flexDirection="column"
-            css={animationStyles}
+            css={animationStyles(
```
mshwery: I wonder if we could make use of glamor's `compose` instead? That way consumers can literally put whatever styles and animations they want – not restricted to keyframes.

rafaelcamaram: @mshwery I'm not sure if I understood what you meant; since the consumers are able to put whatever styles they want on the `containerProps`; even overwrite the `CSS` property.

Can you please explain to me a little bit more your thoughts?

mshwery: Great point – so this is mostly just a convenience method to expose glamor's `keyframes` in a simplified way to consumers?

Why not override with the `css` prop like you suggested?

rafaelcamaram: If we would do this overwriting the CSS property with the `containerProps`, we would need to implement all the following variables/functions _(`animationStyles`, `generateOpenAnimation` and `generateCloseAnimation`)_ on the client side. So I believe that, in this way (I implemented on this PR), we're going to have a cleaner API.

rafaelcamaram: @mshwery Can you please give me your thoughts?

mshwery: I think your API for this component is pretty sound, and it seems like a reasonable benefit/simplification.

While I think it's a good addition, I'm concerned that this is an ad-hoc feature that only benefits this component. Something like this should probably be more "baked-in" to the design of the library in a consistent way across all/many components – any that have animations or css to override...

I'm also not sure that the library needs to solve this immediately, because as you said – you can achieve this through props today, although it's a bit more verbose. One could write a wrapping component that provides this interface, if you find yourself doing it often.

I'll have to share with the rest of the team and would love feedback from others in the community. :) 

Can we put this on hold until we get that feedback? Also, I would be open to exploring a more holistic approach that solves the problem across multiple components.

rafaelcamaram: @mshwery I agree with you. This change needs to be done in a general way; so all the components can take advantage of it. Otherwise, we're going to have components with different behavior from each other.

Let's put it on hold, as you said. If you need anything in this meanwhile, please let me know. I'll try to develop something more generic for it

mshwery:
Going to close this for now – we can revisit when we have a more holistic approach. 👍 Thanks!

PR #683, opened 2019-11-28
Update index.d.ts

tomitrescak:
Menu group title can also be string.

# 🎉 Thanks for taking the time to contribute to 🌲Evergreen! 🎉

It is highly appreciated that you take the time to help improve Evergreen.
We appreciate it if you would take the time to document your Pull Request.

Sadly, if we don't receive enough information, or the Pull Request doesn't
align well with our roadmap, we might respectfully
thank you for your time, and close the issue.

## Respect earns Respect 👏

Please respect our Code of Conduct, in short:

- Using welcoming and inclusive language.
- Being respectful of differing viewpoints and experiences.
- Gracefully accepting constructive criticism.
- Focusing on what is best for the community.
- Showing empathy towards other community members.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1104,7 +1104,7 @@ declare module 'evergreen-ui' {
   }
 
   export interface MenuGroupProps extends Omit<PaneProps, 'title'> {
-    title?: JSX.Element | string
+    title?: React.ReactNode
     children: React.ReactNode[] | React.ReactNode
   }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -1104,7 +1104,7 @@ declare module 'evergreen-ui' {
   }
 
   export interface MenuGroupProps extends Omit<PaneProps, 'title'> {
-    title?: JSX.Element
+    title?: JSX.Element | string
     children: React.ReactNode[] | React.ReactNode
   }
 


index.d.ts
```diff
@@ -1104,7 +1104,7 @@ declare module 'evergreen-ui' {
   }
 
   export interface MenuGroupProps extends Omit<PaneProps, 'title'> {
-    title?: JSX.Element
+    title?: JSX.Element | string
```
maraisr: I'd probably say that this is not quite right. The more "correct" way would be to turn this title, into a `ReactElement | ReactText` (or simply `ReactChild`).

Ultimately it depends on how it's used. Is it simply `<p>{title}</p>`, or is there a `crateElement(title, null, null)`?

tomitrescak:
Sooo ... how do you just pass string in it? With your proposal it’s not possible. Please let me know. It’s throwing errors now.

index.d.ts
mshwery: I think `React.ReactNode` is the most accurate type for `title` given how it's used: https://github.com/segmentio/evergreen/blob/master/src/menu/src/MenuGroup.js#L25

mshwery:
@tomitrescak in the meantime (while waiting for this type fix) you can adjust the types in your own repo locally by overriding them, or you can temporarily use `as JSX.Element` (or `as any` if you must)

tomitrescak:
@mshwery that's what I did ;) thanks.

index.d.ts
mshwery: @tomitrescak did you try `React.ReactNode` ?

tomitrescak: Ok, I updated the PR

PR #686, opened 2019-11-29
Allow passing a ref (innerRef) to a Checkbox

tusbar:
I believe it is time to think about implementing ref forwarding – so that passing a ref always does what you expect.

I have started using https://github.com/react-hook-form/react-hook-form – and it would be nice if this library would support it as the API is really great.

### DIFF
--- a/src/checkbox/src/Checkbox.js
+++ b/src/checkbox/src/Checkbox.js
@@ -74,6 +74,11 @@ class Checkbox extends PureComponent {
     indeterminate: PropTypes.bool,
 
     /**
+     * Function that returns the ref of the checkbox.
+     */
+    innerRef: PropTypes.func,
+
+    /**
      * Function called when state changes.
      */
     onChange: PropTypes.func,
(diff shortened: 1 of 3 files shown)


src/checkbox/src/Checkbox.js
```diff
@@ -104,13 +109,15 @@ class Checkbox extends PureComponent {
...
+    innerRef: () => {},
     onChange: () => {},
     appearance: 'default'
   }
 
-  setIndeterminate = el => {
+  handleInnerRef = el => {
     if (!el) return
     el.indeterminate = this.props.indeterminate
+    this.props.innerRef(el)
```
mshwery: wouldn't you want to bubble the `innerRef` even when `!el`?

```suggestion
    if (el) {
      el.indeterminate = this.props.indeterminate
    }
    this.props.innerRef(el)
```

mshwery:
Also I agree that we should start adding support for ref forwarding.

src/checkbox/src/Checkbox.js
tusbar: Yeah, that makes sense.

tusbar:
Fixed.

src/checkbox/src/Checkbox.js
```diff
@@ -73,6 +73,11 @@ class Checkbox extends PureComponent {
      */
     indeterminate: PropTypes.bool,
 
+    /**
+     * Function that returns the ref of the checkbox.
+     */
+    innerRef: PropTypes.func,
```
mshwery: Can you update the `index.d.ts` typedef too?

tusbar: 👍

Issue #597, opened 2019-07-16
FilePicker onBlur throws an error


dy:
```js
import React from 'react'
import { render } from 'react-dom'
import { FilePicker } from 'evergreen-ui'

render(<FilePicker/>, document.querySelector('#app'))
```

If we click on filepicker text input and then blur, it throws an error.
That happens in all the examples in docs https://evergreen.segment.com/components/filepicker

ronocod:
Looks like it's related to https://github.com/segmentio/evergreen/pull/533

Chrischuck:
If nobody's working on this, I'd like to take this issue

mshwery:
@Chrischuck go for it! 🙌

PR #693, opened 2019-12-31
Fixes input onBlur error in FilePicker

Chrischuck:
Fixes [#597 ](https://github.com/segmentio/evergreen/issues/597)

This error occurs because you're not allowed to set `event.target`'s `files` property to a regular array, it must be a `FileList`.

Unfortunately you can't create a `FileList` from an array so to work around that I used the ref to pull the input's `FileList` directly.

This seems to fix the issue + maintain the previous functionality of the `onBlur` handler

### DIFF
--- a/src/file-picker/src/FilePicker.js
+++ b/src/file-picker/src/FilePicker.js
@@ -174,7 +174,9 @@ export default class FilePicker extends PureComponent {
   }
 
   handleBlur = e => {
-    if (e && e.target) e.target.files = this.state.files
+    // Setting e.target.files to an array fails. It must be a FileList
+    if (e && e.target) e.target.files = this.fileInput.files
+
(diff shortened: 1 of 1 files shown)


src/file-picker/src/FilePicker.js
```diff
@@ -174,7 +174,9 @@ export default class FilePicker extends PureComponent {
   }
 
   handleBlur = e => {
-    if (e && e.target) e.target.files = this.state.files
+    // Setting e.target.files to an array fails. It must be a FileList
+    if (e && e.target) e.target.files = this.fileInput.files
```
mshwery: We'll need to ensure we check that `this.fileInput` is not null.

Chrischuck: done

PR #694, opened 2019-12-31
[CCPA] Update consent manager, and add in-regions for React consent manager

notfelineit:
This PR updates the Consent Manager used on evergreen docs to respect CA consent.
In terms of respecting CCPA, added features are as follows:

- [x]  If in CA, default is to not load advertising, load everything else
- [x]  If in EU, default is to not load anything
- [x]  For both CA and EU, banner should be shown
- [x]  For everyone else, analytics.js should be automatically loaded
- [x]  For everyone (EU/CA/Everywhere else), preferences can be updated, and should be reflected

The banner appears at the top of the page:
Screen Shot 2019-12-31 at 12 13 42 AM

and is closable, while the `Website Data Collection Preferences` button re-opens the consent window:
evergreen


package.json
```diff
@@ -46,13 +46,16 @@
...
+    "@segment/consent-manager": "^4.2.2",
+    "@segment/in-regions": "^1.2.0",
     "@types/react": "^16.9.5",
     "arrify": "^1.0.1",
     "classnames": "^2.2.6",
     "dom-helpers": "^3.2.1",
     "downshift": "^3.3.1",
     "fuzzaldrin-plus": "^0.6.0",
     "glamor": "^2.20.40",
+    "lodash": "^4.17.11",
```
mshwery: What's the reason for these package.json changes?

docs/src/components/ConsentManager.js
```diff
@@ -1,9 +1,36 @@
...
-import { ConsentManager, openConsentManager } from '@segment/consent-manager'
-import inEU from '@segment/in-eu'
+import inRegions from '@segment/in-regions'
+import { ConsentManager, CloseBehavior } from '@segment/consent-manager'
 import segmentWriteKey from '../../segmentWriteKey'
 
 export default () => {
+  const inCA = inRegions(['CA'])
+  const inEU = inRegions(['EU'])
+  const shouldRequireConsent = inRegions(['CA', 'EU'])
```
mshwery: Non-blocking: maybe `inRegions` should export a list of valid regions rather than relying on magic strings.

docs/src/components/ConsentManager.js
```diff
@@ -1,9 +1,36 @@
...
+  const caDefaultPreferences = {
+    advertising: false,
+    marketingAndAnalytics: true,
+    functional: true
+  }
+
+  const euDefaultPreferences = {
```
mshwery: I'm wondering if it makes more sense for the consent manager itself to own the defaults for given regions? Or maybe you can pass a `isCCPA` that handles the default preferences?

mshwery: I say this because it seems a bit cumbersome to put the onus on consumers to know which default preferences to use for which regions.

notfelineit: That's a really good idea!!  I wanted to be careful about that from a legal standpoint, and thought putting ownership on consumers would release us from any liability around guaranteeing CCPA compliance.

Maybe we can do that in a next version if we get it approved by legal/ leave an asterisk?

docs/src/components/ConsentManager.js
notfelineit: Totally! I'd like to do that in a next version, maybe in 2020 😉

package.json
notfelineit: Good q..I'll remove this.

PR #699, opened 2020-01-07
Makes docs app use same react instance as evergreen-ui

Chrischuck:
This change (or something to the effect of this change) is required in order for future usage of hooks.

The problem is that when you run the `docs`, it's using the locally built `evergreen-ui` which has one version of React while the `docs` have another.

The error can be explained here https://github.com/facebook/react/issues/13991.

I'm not sure if this is the best course of action in order to make hooks work in the `docs` app, but after converting the `Spinner` component to use hooks (worked in storybook but not `docs`), this ended up working.

### DIFF
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -23,6 +23,7 @@ exports.onCreateWebpackConfig = ({ actions }) => {
       // Force Gatsby to look for dependencies within the local node_modules from docs.
       modules: ['node_modules'],
       alias: {
+        react: path.resolve(__dirname, '../node_modules/react'),
         'evergreen-ui': path.resolve(__dirname, '../src/index.js'),
         components: path.resolve(__dirname, './src/components')
       }


Chrischuck:
@mshwery that didn't seem to work for me unfortunately.

Even if I boosted the docs to use the same version as `evergreen` itself it wouldn't work. My guess is because even though they're the same version, simply being different instances is enough to break it.

Chrischuck:
@mshwery do you know when this'll get merged in? I think this is the only thing blocking hooks from working!

docs/gatsby-node.js
```diff
@@ -23,6 +23,7 @@ exports.onCreateWebpackConfig = ({ actions }) => {
       // Force Gatsby to look for dependencies within the local node_modules from docs.
       modules: ['node_modules'],
       alias: {
+        react: path.resolve(__dirname, '../node_modules/react'),
```
mshwery: What about upgrading the docs version of React to match? Isn't that problem. I'm not sure about resolving up a directory.

Chrischuck: The issue is this: https://github.com/facebook/react/issues/13991 

There's a couple resolutions in the comments, another option would be to import evergreen from `npm` instead of from `src`

mshwery:
We can give this a shot :)

Chrischuck:
@mshwery sweet! Hopefully nothing adverse happens, but I believe this would technically bump the doc's React version since it'll be using evergreen's version

Issue #610, opened 2019-08-22
Improve the documentation for the custom theme section


rumsky:
Improve the documentation for the custom theme section

Chrischuck:
I can work on this :)

akleiner2:
Think this should've been addressed by #702. Let us know if you have any other requests / issues. Thanks!

PR #702, opened 2020-01-09
Update theme docs

Chrischuck:
For issue #610 

dfsafsadff

Needs PR #699 For the `useTheme` docs 

I'm not exactly sure the full scope/capabilities of theming atm. If there's a feature I'm missing feel free to let me know.

And regarding the copy, feel free to make suggestions + edits :)

I'll probably end up filling out the Theme Object attributes in another pr.


mshwery:
Part of the reason it's undocumented is because it's incomplete :) 

Theming works OK but you actually have to mutate the `defaultTheme` in some cases – the ThemeProvider doesn't always do what it's supposed to do unfortunately.

Thanks for taking a stab at this! We'll take a look and provide some feedback. 👀

docs/src/components/DocsMDXProvider.js
```diff
@@ -29,7 +29,21 @@ Components.strong = props => <strong className="strong" {...props} />
...
+
+  static defaultProps = {
+    noInline: false
+  }
+
+  constructor(props) {
+    super(props)
+
+    // This feels hacky...
+    Components.code = newProps => (
```
mshwery: Can you document the purpose of it?

docs/src/pages/get-started/theming.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+In order to style your app, you must wrap it in the `ThemeProvider` component.
+This component uses the `React.Context` API under the hood.
+
+In order to use the `ThemeProvider` component, you must pass it a value prop that is a `Theme` object.
+
+## Simple theming
+
+The `ThemeProvider` Component, you can either override the default styles(shown below) or use your own theme completely.
+You can add anything you want to the theme object, as long as you have required properties.
```
mshwery: I think we may want to caution users that theming is not 100% complete and some things may not work as expected – they should also be mindful of making sure the `theme` they provide to the ThemeProvider has all of the necessary theme functions included.

The way you show will work fairly well (`{ ...defaultTheme, ...myCustomTheme }`) but I think there are some unique cases that are looking specifically at `defaultTheme` instead of the context 😬

Chrischuck: Thanks for clarifying, I wasn't sure if theming was fully complete yet, will definitely add that little note

docs/src/components/DocsMDXProvider.js
Chrischuck: I needed to pass the `noInline` prop to the `react-live` component in order to show full examples in the theme docs rather than just a single component.

Do you know of a better way to do this? or should adding a comment here suffice

mshwery: Curious if we could instead do this on a component-by-component basis? That would avoid needing to propagate it from `DocsMDXProvider` to the `.code` component right?

Chrischuck: How would it work on a component-by-component basis?

I'm thinking alternatively we could just have the `ThemeLayout` have its own MDXProvider  since I don't think there'll be another case where the `noInline` prop will be needed for `react-live`

mshwery:
I couldn't get this running locally oddly.

Chrischuck:
> I couldn't get this running locally oddly.

weird.. any errors? what happened?

Chrischuck:
On a side note, do you think we could do something like Material UI does by having a `createMuiTheme` so the end user wouldn't have to create the entire theme object? I'm trying to document each theme object attribute and I don't think I want to spend 6 hours doing it 😂

mshwery:
@Chrischuck I think there was a WIP branch for theming that essentially did that – we'll have to dig it up and see if it's still working :)

Chrischuck:
@mshwery if it still works and gets merged in, I'll be glad to update the docs again

PR #719, opened 2020-01-21
Upgrade dev deps and fix linting errors

mshwery:
This is a fairly small 😄 update that mostly focused on devDeps. I touched a few files to make the linter happy.


package.json
```diff
@@ -165,12 +174,10 @@
   "lint-staged": {
     "*.js": [
       "xo --fix",
-      "prettier --write",
-      "git add"
+      "prettier --write"
```
Rowno: Won't removing the `git add` prevent the `prettier` changes from being committed?

mshwery: Nope! The latest version of `lint-staged` does that automatically be default. It actually told me to remove this line.

Rowno: Ok

Issue #714, opened 2020-01-20
UMD support


g1eny0ung:
Are there any plans to support the UMD module? If any, I would glad to help.

mshwery:
We currently create both esm (import/export) and commonjs (require) style modules which works quite well for a broad majority of consumers (e.g. TypeScript / babel users typically rely on the esm modules).

What did you have in mind for supporting UMD? PRs welcome! :)

g1eny0ung:
My original idea was to integrate evergreen into [cljsjs](http://cljsjs.github.io/), which is a package set with google closure externs support, use in clojurescript. It easy to implement this with a UMD module.

I will make a PR to support UMD and try not to affect the current production build.~

g1eny0ung:
@mshwery Hi, I already make a PR.

I use rollup to emit the UMD bundles. Include development and production versions.

In order to be able to pack, I create `index.umd.js`, it removes the `ssr` imports. But I'm not sure Is this a good way because the UMD module also supports node.js environment. Currently, I think it's not very important because we use the main entry to import the commonjs module in node.js.

`{  "main": "commonjs/index.js" }`

This PR also relates to issue #700. I also prefer to use a module bundler instead of through babel directly.

PR #720, opened 2020-01-22
Add UMD support

g1eny0ung:
Refer to #714


src/index.umd.js
```diff
@@ -0,0 +1,84 @@
+export { Alert, InlineAlert } from './alert'
```
mshwery: Can we reuse the existing `index.js` export file? I realize you want to omit ssr, but if that's the only thing maybe we export everything from `index.js` and stub the ssr export with no-op functions for the umd bundle?

rollup.config.js
```diff
@@ -0,0 +1,85 @@
+import babel from 'rollup-plugin-babel'
+import commonjs from '@rollup/plugin-commonjs'
+import replace from '@rollup/plugin-replace'
+import resolve from '@rollup/plugin-node-resolve'
+import { terser } from 'rollup-plugin-terser'
+import pkg from './package.json'
+
+const { peerDependencies } = pkg
+const external = Object.keys(peerDependencies)
```
mshwery: Can you add some documentation comments to this file to explain why we do things like this?

rollup.config.js
```diff
@@ -0,0 +1,85 @@
...
+    'number',
+    'object',
+    'oneOf',
+    'oneOfType',
+    'string'
+  ]
+}
+
+export default [
+  // UMD Development
```
mshwery: How does rollup know that this is development and the 2nd item in the exported array is production?

rollup.config.js
g1eny0ung: Yep. I will do it.

rollup.config.js
g1eny0ung: Rollup doesn't know about it. The first I specified is development because I don't use something optimization. And in the 2nd item, I use `terser()` to minify the generated.

src/index.umd.js
g1eny0ung: You are right. I will try it. 👌

g1eny0ung: At last I create `index.umd.js` in the `src/ssr` folder, make `extractStyles` and `autoHydrate` be an empty function, then replace the `./ssr` import string with `./ssr/index.umd.js` in the UMD build.

I think this can solve the reuse problem. If we have any changes in the future, we can keep the main entry unchanged.

src/ssr/index.umd.js
```diff
@@ -0,0 +1,2 @@
+export function extractStyles() {}
+export function autoHydrate() {}
```
mshwery: Should we throw a `console.warn` in here that these functions are not available in umd bundles for the browser?

rollup.config.js
```diff
@@ -5,9 +5,13 @@ import resolve from '@rollup/plugin-node-resolve'
 import { terser } from 'rollup-plugin-terser'
 import pkg from './package.json'
 
+// Use peerDependencies as rollup external
+// https://rollupjs.org/guide/en/#peer-dependencies
 const { peerDependencies } = pkg
 const external = Object.keys(peerDependencies)
 
+// Explicitly specify unresolvable named exports.
+// https://github.com/rollup/plugins/tree/master/packages/commonjs#namedexports
```
mshwery: Why are these specifically causing problems that require being named exports? Wondering what you ran into to know this?

src/ssr/index.umd.js
g1eny0ung: I think so. I will add soon.

rollup.config.js
g1eny0ung: For example, like `react-is`, it's main entry contains:

```js
'use strict';

if (process.env.NODE_ENV === 'production') {
  module.exports = require('./cjs/react-is.production.min.js');
} else {
  module.exports = require('./cjs/react-is.development.js');
}
```

From the perspective of rollup, this module doesn't export `isForwardRef`(use in evergreen) explicitly, although it works, we should let rollup know how to resolve the exports. So we need to specify the namedExports.

g1eny0ung:
> Can we hook up `size-limit` to check the umd bundle(s) too?

Let me try.

src/ssr/index.umd.js
g1eny0ung: Added.~

mshwery:
@g1eny0ung I think our size limit for the existing commonjs bundle is ~210kb.

g1eny0ung:
@mshwery passed!

image

package.json
```diff
@@ -193,6 +203,10 @@
     {
       "path": "commonjs/index.js",
       "limit": "212 KB"
+    },
+    {
+      "path": "umd/evergreen.min.js",
```
mshwery: Need to check: are we using the minified commonjs output for size limit checks? If not we may need to figure out why this is so big u minified...

g1eny0ung: No.

Check the build script: `"build-commonjs": "BABEL_ENV=commonjs babel src --out-dir commonjs --ignore '**/stories','**/test' --source-maps inline"`

We didn't use any compress options.

For me I prefer to use: `BABEL_ENV=commonjs babel src --out-dir commonjs --ignore '**/stories','**/test' --minified --no-comments`

Remove source map and comments, then compress it. Same with `build-esm` script.

Then the `commonjs` dir size will reduce from 1.8 MB to 580 KB after I test.

I also test with the `size-limit`. It seems whether I compress it or not, the gzipped sizes are all `210.3 KB`.

mshwery: I think size-limit will automatically apply the gzip. Maybe we should use the unminified version for umd size-limit checks?

g1eny0ung: Even if we compress more, the umd output still can't be reduced to 220kb......because it bundles all dependencies to one file......not like commonjs, only babel transforming.

g1eny0ung: And also, I think we still need to update the `build-commonjs` and `build-esm`. Like I said above: Remove source map and comments, then compress it.

These will reduce the npm package's size.

mshwery: Ah true, I wasn't thinking of that aspect of umd. Ok, I think we can ship this as-is and improve upon it if we need based on actual usage feedback.

g1eny0ung:
> Thank you for knocking this out! Amazing to add more support :)

You' re welcome. I like the style of evergrenn-ui very much. Expect it to be used by more people. 😆

g1eny0ung:
> Thank you for knocking this out! Amazing to add more support :)

You' re welcome. I like the style of evergrenn-ui very much. Expect it to be used by more people. 😆

PR #723, opened 2020-01-29
Converts avatar component to hooks

Chrischuck:
Pretty straight forward port from `PureComponent` to `hooks`

Just make sure it works in storybook + docs. 

**Before: ~19.1kb**
before

**After: ~16.4kb**
Screen Shot 2020-01-28 at 22 46 29

### DIFF
--- a/src/avatar/src/Avatar.js
+++ b/src/avatar/src/Avatar.js
@@ -1,10 +1,10 @@
 import { css } from 'glamor'
-import React, { PureComponent } from 'react'
+import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
 import { Image } from '../../image'
 import { Text } from '../../typography'
-import { withTheme } from '../../theme'
+import { useTheme } from '../../theme'
 import globalGetInitials from './utils/getInitials'
 import globalHash from './utils/hash'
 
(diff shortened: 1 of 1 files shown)


src/avatar/src/Avatar.js
```diff
@@ -21,167 +21,145 @@ const initialsStyleClass = css({
...
+  color: 'automatic',
+  size: 24,
+  isSolid: false,
+  getInitials: globalGetInitials,
+  forceShowInitials: false,
+  sizeLimitOneCharacter: 20
 }
 
-export default withTheme(Avatar)
+export default Avatar
```
mshwery: Same comment as in https://github.com/segmentio/evergreen/pull/724#discussion_r372533724 – for parity we should probably support `forwardRef` + `React.memo`?

src/avatar/src/Avatar.js
```diff
@@ -21,167 +21,145 @@ const initialsStyleClass = css({
...
-    forceShowInitials: false,
-    sizeLimitOneCharacter: 20
-  }
+  return theme.getAvatarProps({ isSolid, color })
+}
 
-  constructor(props, context) {
-    super(props, context)
-    this.state = { imageHasFailedLoading: false }
+function Avatar({
```
mshwery: For `defaultProps` should those just be assigned in the method signature? e.g. `forceShowInitials = false`

src/avatar/src/Avatar.js
```diff
@@ -21,167 +21,137 @@ const initialsStyleClass = css({
...
+  forceShowInitials: PropTypes.bool,
+
+  /**
+   * When the size is smaller than this number, use a single initial for the avatar.
+   */
+  sizeLimitOneCharacter: PropTypes.number
 }
 
-export default withTheme(Avatar)
+export default memo(forwardRef(Avatar))
```
mshwery: Might be wrong but I think we'll also want to set the `.displayName` of the component (or the forwardRef component... or the memoized component)?

Chrischuck:
> https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

Seems like a good idea to throw it into v5, curiously, how would rolling out other components work. Since the non hook components would remain the same in v5 but once those components get converted that would be a breaking change in a minor release if that makes sense.

mshwery:
>Since the non hook components would remain the same in v5 but once those components get converted that would be a breaking change in a minor release if that makes sense.

Yeah I think there are a couple ways we could approach this. 
1. Strict mode: We could target these PRs to a `v5` branch (probably need to start fresh since the current one is quite stale) and once we convert _everything_ over... then we finally release v5 to npm
2. Incremental mode: Ship these PRs as v5 release candidates so we can start opting into them without broadly releasing v5. This will allow us to continue to make breaking changes on v5 until we've finished all the incremental PRs. (Though maybe we still want to target a new v5 branch for that instead of `master`?)

Chrischuck:
> > Since the non hook components would remain the same in v5 but once those components get converted that would be a breaking change in a minor release if that makes sense.
> 
> Yeah I think there are a couple ways we could approach this.
> 
> 1. Strict mode: We could target these PRs to a `v5` branch (probably need to start fresh since the current one is quite stale) and once we convert _everything_ over... then we finally release v5 to npm
> 2. Incremental mode: Ship these PRs as v5 release candidates so we can start opting into them without broadly releasing v5. This will allow us to continue to make breaking changes on v5 until we've finished all the incremental PRs. (Though maybe we still want to target a new v5 branch for that instead of `master`?)

Personally, I think incremental mode is the way to go so there's at least some level of real world testing. What're your thoughts?

Chrischuck:
@mshwery this one good to go?

src/avatar/src/Avatar.js
```diff
@@ -156,7 +77,8 @@ class Avatar extends PureComponent {
         justifyContent="center"
         backgroundColor={colorProps.backgroundColor}
         title={name}
-        {...props}
+        innerRef={ref}
```
mshwery, line 80: We'll need to update this to `ref` once https://github.com/segmentio/evergreen/pull/843 lands :)

mshwery:
Thanks for this @Chrischuck ! 🙌

PR #724, opened 2020-01-29
Migrates Spinner component to hooks

Chrischuck:
Pretty straight forward port from `PureComponent` to `hooks`

Just make sure it works in storybook + docs. 

Not much savings here
**Before ~10kb**
Screen Shot 2020-01-28 at 22 54 26

**After ~9kb**
Screen Shot 2020-01-28 at 22 54 26

### DIFF
--- a/src/spinner/src/Spinner.js
+++ b/src/spinner/src/Spinner.js
@@ -2,7 +2,7 @@ import { css } from 'glamor'
 import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
-import { withTheme } from '../../theme'
+import { useTheme } from '../../theme'
 
 const loadingKeyframes = css.keyframes('loading', {
   '0%': {
@@ -38,7 +38,8 @@ const innerClass = color =>
     fill: 'transparent'
   }).toString()
 
-const Spinner = ({ delay, theme, size, ...props }) => {
+const Spinner = ({ delay, size, ...props }) => {
+  const theme = useTheme()
   const [isVisible, setIsVisible] = useState(delay === 0)
   const [delayTimer, setDelayTimer] = useState(null)
 
@@ -94,12 +95,7 @@ Spinner.propTypes = {
   /**
    * The size of the spinner.
    */
-  size: PropTypes.number.isRequired,
-
-  /**
-   * Theme provided by ThemeProvider.
-   */
-  theme: PropTypes.object.isRequired
+  size: PropTypes.number.isRequired
 }
 
 Spinner.defaultProps = {
@@ -107,4 +103,4 @@ Spinner.defaultProps = {
   delay: 0
 }
 
-export default withTheme(Spinner)
+export default Spinner
--- a/src/spinner/src/Spinner.js
+++ b/src/spinner/src/Spinner.js
@@ -2,7 +2,7 @@ import { css } from 'glamor'
 import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
-import { withTheme } from '../../theme'
+import { useTheme } from '../../theme'
 
 const loadingKeyframes = css.keyframes('loading', {
   '0%': {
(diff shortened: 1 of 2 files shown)


src/spinner/src/Spinner.js
```diff
@@ -38,84 +38,69 @@ const innerClass = color =>
...
-    super()
-
-    this.state = {
-      isVisible: delay === 0
+  useEffect(() => {
+    if (delay > 0) {
+      const newDelayTimer = setTimeout(() => {
+        setIsVisible(true)
+      }, delay)
+      setDelayTimer(newDelayTimer)
```
mshwery: I dont think we even need to use state for this – we know the `id` and can return that as part of the effect cleanup return function.

src/spinner/src/Spinner.js
```diff
@@ -38,84 +38,69 @@ const innerClass = color =>
...
-  componentWillUnmount() {
-    clearTimeout(this.delayTimer)
-  }
+Spinner.defaultProps = {
+  size: 40,
+  delay: 0
 }
 
-export default withTheme(Spinner)
+export default Spinner
```
mshwery: I think to have parity with `PureComponent` we need to support `forwardRef` and `React.memo` right?

Chrischuck: Cool, not exactly sure what goes into making it work the same as `PureComponent`, I get the `memo` but im not sure how `forwardRef` plays into it. Will take a look tonight!

mshwery: forwardRef will make sure that consumers can still pass `ref` like they could with PureComponent.

We'll have to hook it up to either `innerRef` from ui-box or ref depending on what the inner component is.

Chrischuck: Solid, thanks for the clarification!

src/spinner/src/Spinner.js
```diff
@@ -98,9 +97,4 @@ Spinner.propTypes = {
   size: PropTypes.number.isRequired
 }
 
-Spinner.defaultProps = {
-  size: 40,
-  delay: 0
-}
-
-export default Spinner
+export default memo(forwardRef(Spinner))
```
mshwery: I think we want/need `.displayName` for the final component?

src/spinner/src/Spinner.js
```diff
@@ -38,84 +38,63 @@ const innerClass = color =>
...
     }
-  }
 
-  render() {
-    if (!this.state.isVisible) {
-      return null
+    return function() {
+      clearTimeout(delayTimer)
     }
+  }, [])
```
IdealSystemsMCP: I got 2 small suggestions:
- I believe we should rerun the effect when the delay prop changes (I know it will only have an effect before the spinner is visible, so maybe it's overkill but I think the eslint-plugin-react-hooks would have warned for this, so we should add it)
- you can return  the cleanup function inside the if statement (less and more obvious code)

```javascript
useEffect(() => {
  if (delay > 0) {
    let timeout = setTimeout(() => {
      setIsVisible(true)
    }, delay)

    return () => clearTimeout(timeout)
  }
}, [delay])
```

mshwery: I think the current implementation mimics what we have today (only `componentDidMount` actually does anything with `delay` – changing it to another value will not change the behavior.

IdealSystemsMCP: As you wish, but don't you think we should add the [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) to at least warn when something like this (see: react-hooks/exhaustive-deps)  happens?

mshwery:
@Chrischuck I retargeted v5 as the target branch. Can you do a rebase to pull in latest v5 here? We can get this merged in. :)

src/spinner/src/Spinner.js
```diff
@@ -38,84 +38,63 @@ const innerClass = color =>
...
+        />
+      </Box>
+    </Box>
+  )
+}
 
-  componentWillUnmount() {
-    clearTimeout(this.delayTimer)
-  }
+Spinner.propTypes = {
```
mshwery, line 83: Apparently we have to add propTypes to the final resulting exported component instead.

PR #725, opened 2020-01-30
Migrate Alert component to hooks

Chrischuck:
Straight forward change for this hook. Hopefully these simple components will be easy to get out of the way

### DIFF
--- a/src/alert/src/Alert.js
+++ b/src/alert/src/Alert.js
@@ -52,7 +52,7 @@ const Alert = memo(
             display="flex"
             alignItems="center"
           >
-            <Icon size={14} {...theme.getIconForIntent(intent)} />
+            {theme.getIconForIntent(intent, { size: 14 })}
           </Pane>
         )}
         <Pane display="flex" width="100%">
--- a/src/alert/src/Alert.js
+++ b/src/alert/src/Alert.js
@@ -1,163 +1,154 @@
-import React, { PureComponent } from 'react'
+import React, { memo, forwardRef } from 'react'
 import PropTypes from 'prop-types'
 import { spacing, dimensions, position, layout } from 'ui-box'
-import { withTheme } from '../../theme'
+import { useTheme } from '../../theme'
 import { Pane } from '../../layers'
 import { Heading, Paragraph } from '../../typography'
 import { IconButton } from '../../buttons'
 
-class Alert extends PureComponent {
-  static propTypes = {
-    /**
-     * Composes some Box APIs.
-     */
-    ...spacing.propTypes,
-    ...position.propTypes,
-    ...layout.propTypes,
-    ...dimensions.propTypes,
+function Alert(
+  {
+    title,
+    children,
+    onRemove,
+    intent = 'none',
+    hasTrim = true,
+    hasIcon = true,
+    appearance = 'default',
+    isRemoveable = false,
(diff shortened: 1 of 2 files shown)


src/alert/src/Alert.js
```diff
@@ -1,170 +1,153 @@
...
-    hasIcon: true,
-    isRemoveable: false,
-    appearance: 'default'
-  }
-
-  getIconForIntent = intent => {
-    const { theme } = this.props
-
-    return <Icon size={14} {...theme.getIconForIntent(intent)} />
-  }
```
Chrischuck: I'm curious why this separate function was needed/used? Will removing cause anything funky?

src/alert/src/InlineAlert.js
```diff
@@ -1,77 +1,69 @@
...
-    )
-  }
+  /**
+   * Theme provided by ThemeProvider.
+   */
+  theme: PropTypes.object.isRequired
 }
 
-export default withTheme(InlineAlert)
+export default memo(forwardRef(InlineAlert))
```
mshwery: Do we need to set the `.displayName` of the component? Not sure if we need it but it might be nice.

src/alert/src/Alert.js
mshwery: Good question – not really needed anymore.

src/alert/src/InlineAlert.js
Chrischuck: Not sure if what I did is what you're looking for, lmk if it is/isn't and I'll make those changes to the other PR's as well :)

Chrischuck:
> As for the `displayName` I think you'll be able to look via the React Dev Tools to see how the component hierarchy reads – if you see anonymous components we might want to move the component name there, but at this point it's purely a Dev Tools / DX concern.

Screen Shot 2020-01-30 at 21 19 58
I dont _think_ there's any anonymous components (picture from before the last commit I pushed.

I'm not too familiar with the `displayName` property, learn something new everyday! But I apologize in advance!

mshwery:
OH man, now we just gotta get `Pane` converted to `useTheme` and the component hierarchy will be _so much cleaner_

mshwery:
@Chrischuck what happens in the React Dev Tools tree if you remove the `.displayName` ? I'm curious if it's actually needed or if React is smart enough to pull the name from the original component when using `React.memo` and `forwardRef`?

Chrischuck:
> OH man, now we just gotta get `Pane` converted to `useTheme` and the component hierarchy will be _so much cleaner_

That can be arranged 👀

Chrischuck:
> @Chrischuck what happens in the React Dev Tools tree if you remove the `.displayName` ? I'm curious if it's actually needed or if React is smart enough to pull the name from the original component when using `React.memo` and `forwardRef`?

The picture I posted was when I removed explicitly setting the `displayName`. It's also the same if I set `displayName` on the wrapped component as well.

Doing this:
`Alert.displayName = 'foofoo'`
Was the only way to get the name to change, so I'm assuming React's HOC's smartly set the `displayName`

mshwery:
Looks like there are some conflicts, but should be good to retarget this against v5 and merge in after resolving those conflicts.

mshwery:
Closing in favor of #823

PR #727, opened 2020-01-31
Converts Pane component to hooks

Chrischuck:
Ask and you shall receive 👀

Converts Pane component to hooks

### DIFF
--- a/src/layers/src/Pane.js
+++ b/src/layers/src/Pane.js
@@ -11,121 +11,117 @@ const StringAndBoolPropType = PropTypes.oneOfType([
 ])
 
 const Pane = memo(
-  forwardRef(
-    (
-      {
-        background,
+  forwardRef((props, ref) => {
+    const {
+      background,
 
-        elevation,
-        hoverElevation,
-        activeElevation,
+      elevation,
+      hoverElevation,
+      activeElevation,
 
-        border,
-        borderTop,
-        borderRight,
-        borderBottom,
-        borderLeft,
+      border,
+      borderTop,
+      borderRight,
+      borderBottom,
+      borderLeft,
 
-        css = {},
-        ...props
-      },
-      ref
-    ) => {
-      const theme = useTheme()
+      css = {},
+      ...restProps
+    } = props
+    const theme = useTheme()
 
-      function getHoverElevationStyle(hoverElevation, css) {
-        if (!Number.isInteger(hoverElevation)) return {}
+    function getHoverElevationStyle(hoverElevation, css) {
+      if (!Number.isInteger(hoverElevation)) return {}
 
-        return {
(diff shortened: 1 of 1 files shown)


mshwery:
@Chrischuck I retargeted v5 as the target branch. Can you do a rebase to pull in latest v5 here? We can get this merged in. :)

src/layers/src/Pane.js
```diff
@@ -136,64 +84,105 @@ class Pane extends PureComponent {
...
+      borderLeft={_borderLeft}
+      boxShadow={elevationStyle}
+      background={theme.getBackground(background)}
+      {...props}
+      className={className}
+    />
+  )
+}
+
+Pane.propTypes = {
```
mshwery, line 127: Apparently we have to add propTypes to the final resulting exported component instead.

Chrischuck: So like

```
const MemoizedPane = memo(forwardRef(Pane))

MemoizedPane.propTypes = {...}
```
?

mshwery: exactly!

Chrischuck: Gettung some proptype list issues when i tried this for the spinner? Do I need to do it on the component itself too?
Screen Shot 2020-04-20 at 21 57 26

mshwery: Can we do something like this instead:

```js
const Spinner = memo(forwardRef((props, ref) => {
  ...
})

Spinner.propTypes = {
  ...
}

export default Spinner
```

Chrischuck: @mshwery Done, my only gripe is it looks a little messy at the declaration point, but not much we can do I guess. Unless maybe:

```
const Spinner = memo(forwardRef((props, ref) => {
  const {
     ...
  } = props
})

Spinner.propTypes = {
  ...
}

export default Spinner
```

instead of restructuring in the parameters?

mshwery: Yeah would be +1 to destructuring the props in the function body instead of in the function declaration.

src/layers/src/Pane.js
```diff
@@ -153,23 +26,77 @@ class Pane extends PureComponent {
...
+        transitionDuration: '150ms',
+        transitionProperty: 'box-shadow, transform',
+        transitionTimingFunction: `cubic-bezier(0.0, 0.0, 0.2, 1)`,
+        ':hover': {
+          ...(css[':hover'] || {}),
+          transform: 'translateY(-2px)',
+          boxShadow: theme.getElevation(hoverElevation)
+        }
+      }
+    }
```
mshwery, line 46: We have opportunity to refactor these into custom hooks, but that can come later.

PR #729, opened 2020-02-02
Adds eslint + prettier npm scripts

Chrischuck:
Adds scripts so you can run lint + pretty easily.

Not sure if other's like having this but I find it useful.

On a related note there's a few lint warnings and prettier fixes that I can clean up in another pr.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,12 @@
 package.json
+*.log
+node_modules
+.DS_Store
+/commonjs/
+/esm/
+/.out/
+/docs/public/
+.cache
+/npm-shrinkwrap.json
+/package-lock.json
+/.idea/
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "postpublish": "yarn deploy-docs",
     "pretty": "prettier --write \"./**/*.{js,json,md}\"",
     "lint": "xo --fix",
-    "pretty-lint": "npm run lint && npm run pretty"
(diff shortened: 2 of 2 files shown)


.prettierignore
```diff
@@ -1 +1,12 @@
...
+*.log
+node_modules
+.DS_Store
+/commonjs/
+/esm/
+/.out/
+/docs/public/
+.cache
+/npm-shrinkwrap.json
+/package-lock.json
```
mshwery: We shouldn't have package-lock or shrinkwrap files in evergreen.

package.json
```diff
@@ -37,7 +37,10 @@
     "clean": "git clean -Xdf",
     "release": "np",
     "deploy-docs": "cd docs && yarn install && yarn deploy",
-    "postpublish": "yarn deploy-docs"
+    "postpublish": "yarn deploy-docs",
+    "pretty": "prettier --write \"./**/*.{js,json,md}\"",
```
mshwery: This automatically gets run on commit. Do we need a script for running it manually?

Chrischuck: Yeah, wasn't too sure if these were needed or not, I like having them to run manually because sometimes stuff slips through the cracks.

package.json
```diff
@@ -37,7 +37,10 @@
...
+    "lint": "xo --fix",
+    "pretty-lint": "npm run lint && npm run pretty"
```
mshwery: These should be `yarn lint && yarn pretty`

Issue #728, opened 2020-02-01
Menu Item when using React Router Link: PropType error and aggressive styling 


cstrat:
One issue and one annoyance.

1 - `Warning: Failed prop type: Invalid prop 'is' supplied to 'MenuItem'`
2 - Doing this with a Menu.Item renders the menu with an underline and coloured like an anchor. Really it should still be styled as a menu-item and not a link. I can override it manually but thats a pain.

Code:

```<Menu>
  <Menu.Item
    is={Link}
    to="/settings/"
    icon="settings"
  >
    Settings
  </Menu.Item>
</Menu>
```

Image:

![image](https://user-images.githubusercontent.com/4209076/73590889-1b80c680-453c-11ea-8477-5e1d4e8006c5.png)

Chrischuck:
For your first issue, looks like the `Menu.Item` component is missing `PropTypes.elementType` in its propTypes validation. 

For the second issue, I would suggest to pass your own `className` to `Menu.Item`. I think it's best not to try an have the library override another library's behavior and leave that up to the developer to change. 

Thoughts @mshwery ?

cstrat:
Yeah the second point I am not wasn't sure of however in other parts of Evergreen it is overridden. Check the docs under 'Button'.
For consistency I expected Menu.Item to behave the same.

![image](https://user-images.githubusercontent.com/4209076/73624761-2cdde600-4696-11ea-9655-371061cb5718.png)

mshwery:
@cstrat @Chrischuck I think it's a reasonable expectation, given that menu is likely to include anchor tags. `Button` [has code that expects anchor tags may be used](https://github.com/segmentio/evergreen/blob/78cf402525b59fcac49e53236f382dba0c60256f/src/themer/src/createButtonAppearance.js#L10). 

Can you use `ui-box` css props to override the style: e.g. `textDecoration="none"` in the meantime?

Chrischuck:
Nice, I can fix that tonight

cstrat:
@mshwery @Chrischuck thanks guys, that is great.

I wanted to try look into how I could try doing a PR myself but I've never used typescript so I'd end up making a mess of it all 😃 

I am overriding the styles at the moment, cheers for your help 👏

Chrischuck:
@cstrat You shouldn't have to mess with `ts` all the src code is in `js`! It's a little tricky to figure out how the theming/styling works under the hood though.

cstrat:
[...]

PR #731, opened 2020-02-03
Adds PropTypes.elementType to is MenuItem prop

Chrischuck:
Fix for part 1 of #728

### DIFF
--- a/src/menu/src/MenuItem.js
+++ b/src/menu/src/MenuItem.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import Box from 'ui-box'
 import { Pane } from '../../layers'
 import { Text } from '../../typography'
 import { Icon } from '../../icon'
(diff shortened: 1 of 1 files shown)


src/menu/src/MenuItem.js
```diff
@@ -13,7 +13,11 @@ class MenuItem extends React.PureComponent {
      * Element type to use for the menu item.
      * For example: `<MenuItem is={ReactRouterLink}>...</MenuItem>`
      */
-    is: PropTypes.oneOfType([PropTypes.string, PropTypes.func]).isRequired,
+    is: PropTypes.oneOfType([
```
Chrischuck: Possibly do `Pane.propTypes.is`?

mshwery: Yeah, that'd be a little bit more accurate.

Chrischuck: Unfortunately I don't think it's possible to access `Pane.propTypes.is` (says Pane.propTypes is undefined). It might be possible to do once Pane is a hook however.

mshwery: You could use `Box.propTypes` like Pane does.

PR #733, opened 2020-02-05
fix(multi-select-menu):   #589 SelectMenu with multiselect allows selecting …

AmanAgarwal041:
…the same item many times with keyboard

### DIFF
--- a/src/select-menu/src/OptionsList.js
+++ b/src/select-menu/src/OptionsList.js
@@ -129,9 +129,10 @@ export default class OptionsList extends PureComponent {
 
     // Preserve backwards compatibility with allowing custom filters, which accept array of strings
     if (typeof optionsFilter === 'function') {
-      return optionsFilter(options.map(item => item.label), searchValue).map(
-        name => options.find(item => item.label === name)
-      )
+      return optionsFilter(
+        options.map(item => item.label),
+        searchValue
+      ).map(name => options.find(item => item.label === name))
     }
 
     return fuzzyFilter(options, searchValue, { key: 'label' })
@@ -164,6 +165,10 @@ export default class OptionsList extends PureComponent {
     if (e.keyCode === 13) {
       this.handleEnter()
     }
+
+    if (e.keyCode === 27) {
+      this.props.close()
+    }
   }
 
   handleArrowUp = () => {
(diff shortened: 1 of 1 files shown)


AmanAgarwal041:
@solon-aguiar

src/select-menu/src/OptionsList.js
```diff
@@ -164,6 +165,10 @@ export default class OptionsList extends PureComponent {
     if (e.keyCode === 13) {
       this.handleEnter()
     }
+
+    if (e.keyCode === 27) {
```
mshwery: Instead of `keyCode` we should use `key`: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

```suggestion
    if (e.key === 'Escape') {
```

src/select-menu/src/OptionsList.js
```diff
@@ -189,14 +198,20 @@ export default class OptionsList extends PureComponent {
       nextIndex = 0
     }
 
+    if (this.isSelected(options[nextIndex])) {
+      return
+    }
+
     onSelect(options[nextIndex])
```
mshwery: Maybe instead of returning early we make the semantics clear by only calling `onSelect` when it's not selected already.

```js
if (!this.isSelected(options[nextIndex])) {
  onSelect(options[nextIndex])
}
```

Also, could you elaborate on when arrow down/up triggers selecting the item that's already selected?

AmanAgarwal041:
Okay. I will update the files and the pull request.

AmanAgarwal041:
@mshwery I have updated the changes and given support for key as well as keyCode.

src/select-menu/src/OptionsList.js
```diff
@@ -153,16 +154,38 @@ export default class OptionsList extends PureComponent {
...
-      this.handleArrowUp()
-    }
-
-    if (e.keyCode === 40) {
-      this.handleArrowDown()
-    }
-
-    if (e.keyCode === 13) {
-      this.handleEnter()
+    if (e.key !== undefined) {
```
mshwery: We don't need to support this case – `event.key` has wide browser support all the way back to IE9.

AmanAgarwal041: Okay. Will remove the code.

AmanAgarwal041: removed keyCode piece

PR #742, opened 2020-02-09
Adds Icon code splitting

Chrischuck:
Takes from what Rowno built a while back. The only changes are rebasing off current master and to use hooks since it'd be going to v5. It still ends up breaking `size-limit` and I'd assume that's the only reason it wasn't merged a while ago.

Maybe we should go the way of material and make icons a separate package? or just increase our size limit.


mshwery:
2 questions:

1. does this have to be a breaking change or can it be introduced in a way that won't impact consumers?
2. i think we are fine with bumping the overall filesize slightly given that it can now be tree-shaken. How much larger is it?

Chrischuck:
@mshwery 
1. I believe this should be able to rollout fine since the old icon component still exists and thus won't break existing things since it's all new stuff.
2. It's ~115kb over the limit

Chrischuck:
@mshwery 

Bumped `size-limit` to 350kb.

Really weird why build is still failing though, seems very unrelated to these changes?
Screen Shot 2020-02-11 at 9 40 50 PM

mshwery:
cc @colinlohner for visibility

Chrischuck:
@akleiner2 Would we be able to get this guy in? Would love to be able to code split icons :)

package.json
```diff
@@ -32,6 +32,7 @@
     "dev": "start-storybook -p 6006",
     "create-package": "./tools/create-package.js",
     "create-package:components": "./tools/create-package-components.js",
+    "generate-icons": "./tools/generate-icons.js",
```
mshwery, line 35: Should we run this during the build?

akleiner2: +1

package.json
```diff
@@ -202,16 +204,16 @@
   "size-limit": [
     {
       "path": "commonjs/index.js",
-      "limit": "212 KB"
+      "limit": "350 KB"
```
mshwery, line 207: I'm wondering if we can omit the original package from the deps after generating icons? 

The main question here is whether we can actually get tree-shaking if we still are importing the blueprint icons in `Icon`.

mshwery:
@Chrischuck Love this work. Thank you for taking up the helm on this.

I see a couple problems though:
1. Because Evergreen internally uses `Icon` (which imports `@blueprintjs/icons`) if your build includes any of those Evergreen components, you are still pulling in all of `@blueprintjs/icons` – status quo with today, but now we've increased the bundle size with duplication from the original package.
2. We are very close to being able to remove `@blueprintjs/icons` from the final bundle(s) completely. I wonder if we can export the IconNames (that's the only thing remaining really).

If we want to release in v4 minor, we can maintain backwards compatibility by exporting a big switch statement in `icon/src/Icon.js` instead – this file could be generated from `generate-icons`. It will still not be tree-shakeable, but as I mentioned in #1 above, that is the status quo for today. It has the added benefit of us not increasing the bundle size at all because we would effectively be replacing `@blueprintjs/icons` entirely with a tree-shakable replica.

mshwery:
For context, here's the `npx size-limit --why`

Screen Shot 2020-04-15 at 11 01 20 AM

mshwery:
I've got some promising results locally – just need to autogen typedefs and check the docs!
Screen Shot 2020-04-15 at 9 33 52 PM

package.json
Chrischuck: @mshwery @akleiner2 run during build and  don't push generated files?

Chrischuck:
> @Chrischuck Love this work. Thank you for taking up the helm on this.
> 
> I see a couple problems though:
> 
> 1. Because Evergreen internally uses `Icon` (which imports `@blueprintjs/icons`) if your build includes any of those Evergreen components, you are still pulling in all of `@blueprintjs/icons` – status quo with today, but now we've increased the bundle size with duplication from the original package.
> 2. We are very close to being able to remove `@blueprintjs/icons` from the final bundle(s) completely. I wonder if we can export the IconNames (that's the only thing remaining really).
> 
> If we want to release in v4 minor, we can maintain backwards compatibility by exporting a big switch statement in `icon/src/Icon.js` instead – this file could be generated from `generate-icons`. It will still not be tree-shakeable, but as I mentioned in #1 above, that is the status quo for today. It has the added benefit of us not increasing the bundle size at all because we would effectively be replacing `@blueprintjs/icons` entirely with a tree-shakable replica.

Thanks for doing a lot of exploratory work @mshwery, gonna see if I can get this done in the next couple of days 🤘

mshwery:
@Chrischuck do you mind if I push directly to this branch?

Chrischuck:
> @Chrischuck do you mind if I push directly to this branch?

All you man!! Go for it

mshwery:
Closing now that #779 landed. Thanks for your amazing push on this branch! Couldn't have shipped without you.

PR #746, opened 2020-02-20
[Dialog] Add support for custom headers and footers

colinking:
The `Dialog` component wraps the `children` with two `Pane`'s -- one that you can supply custom props to, and another that you cannot. Because of that second pane, you can't supply a custom header and footer via the body and still have those header/footer stick to the top/bottom of the Dialog (CMIIW!).

We can't just remove that `Pane`, since it would break the rendering behavior of children in a `Dialog`. As a compromise, this PR introduces a `header` and `footer` component that can be used to override the default header and footer.

| Before | After |
| --- | --- |
| Scrolling doesn't work: image |  Scrolling works: image |

### DIFF
--- a/src/dialog/src/Dialog.js
+++ b/src/dialog/src/Dialog.js
@@ -79,9 +79,21 @@ class Dialog extends React.Component {
     hasHeader: PropTypes.bool,
 
     /**
+     * A custom header to render at the top of the Dialog. If set, this
+     * takes precedence over the default header.
+     */
+    header: PropTypes.element,
+
+    /**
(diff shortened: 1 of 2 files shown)


colinking:
Updated this PR to be a non-breaking change -- it now introduces a header/footer prop.

akleiner2:
@colinking -- can you share a screenshot of what the props table for `Dialog` looks like now? I changed a little bit how we resolve PropTypes in the documentation, so just want to make sure that it's accounted for

colinking:
@akleiner2 here you go! lgtm

image

PR #755, opened 2020-03-12
fix checkbox types

mshwery:
This fixes a bug in our type definitions for the Checkbox component. When we added `innerRef` we failed to omit the BoxProps.innerRef type (it is different in this case).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -576,7 +576,7 @@ declare module 'evergreen-ui' {
   export class Card extends React.PureComponent<CardProps> {
   }
 
-  export interface CheckboxProps extends BoxProps<'input'> {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -576,7 +576,7 @@ declare module 'evergreen-ui' {
   export class Card extends React.PureComponent<CardProps> {
   }
 
-  export interface CheckboxProps extends BoxProps<'input'> {
+  export interface CheckboxProps extends Omit<BoxProps<'input'>, 'innerRef'> {
```
mshwery: I think this should do the trick. It was actually getting caught/highlighted by VS Code directly.

PR #769, opened 2020-03-31
[Docs] Map over and display specific `union` and `enum` types

akleiner2:
As I was glancing over the docs, I noticed that there's a lot of `enum` and `union` types that were masking the specific values that the props actually take. This diff brings forward the specific union and enum types so that the information is displayed up-front.

For example -- now in the `Popover` docs:
image


akleiner2:
Thanks both!

PR #770, opened 2020-03-31
Sync Popover internal open/close state to prop "isShown"

andycarrell:
## Motivation
We are using an evergreen popover to create an address search component. We want the ability to control when the popover is open / closed - for example when a user selects an option, we close the dropdown. We can achieve this via the `isShown` prop, _but_ in doing so we lose 'click outside' and 'close on escape' functionality - because these are only applied when the component is internally controlled.

In the render method popover currently checks whether `isShown` is a boolean - indicating the component is controlled. This change follows the same pattern, syncing the internal state with the prop.

## Changes
Adds check on component did update - if popover is controlled manually, call open / close to ensure event listeners are correctly managed

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -133,6 +133,19 @@ export default class Popover extends Component {
     document.body.removeEventListener('keydown', this.onEsc, false)
   }
 
+  componentDidUpdate(prevProps) {
+    const { isShown } = this.props
+    // If `isShown` is a boolean, popover is controlled manually.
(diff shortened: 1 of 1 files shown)


andycarrell:
@mshwery - are we able to get an update on this? Hope it doesn't go stale if it could be merged 🙏

colinlohner:
I think making this controlled would be a nice improvement in the future. I think this adds some good functionality to the component as is though! Thank you @andycarrell

PR #771, opened 2020-04-01
[Docs] Recursively un-nest `arrayOf`, `shape`, and `union` types

akleiner2:
This builds off of #769 and uncovers a bit more specific type information, especially with regards to `arrayOf` and `shape` types.

For example: 
image


akleiner2:
Thank you both!

Issue #780, opened 2020-04-16
FormField error icon


ghost:
The error icon in `FormField` displays an error icon regardless of what is passed to `validationMessage` unless you specify null. Instead it should render if `validationMessage` isn't falsy. In other words, if an empty string for example is supplied, it should render nothing.

Chrischuck:
I've come across this bug(or is this desired?) as well, happy to make a fix for it @mshwery @akleiner2

jibberishclient:
On that same note it would be nice if you made FormField wrappers for the other form fields too not just text field. I'm using formik and the single component form field is a live saver, otherwise I'd need to destructure some of the props to the input and others to FormField that wraps it.

jibberishclient:
On second thought, I think validationMesssge should be completely ignored unless isInvalid is true

mshwery:
>validationMesssge should be completely ignored unless isInvalid is true

I agree! cc @akleiner2 @colinlohner

colinlohner:
That makes sense to me

jackcwu:
Me and @tcho0501 are on this

akleiner2:
Closing out due to inactivity - please feel free to re-open / take a stab at this if there's availability!

PR #782, opened 2020-04-17
Prevents error message render for empty values

Chrischuck:
fix issue #780 

## Before
Screen Shot 2020-04-16 at 18 02 29

## After
Screen Shot 2020-04-16 at 18 04 09

### DIFF
--- a/src/form-field/src/FormField.js
+++ b/src/form-field/src/FormField.js
@@ -66,6 +66,22 @@ export default class FormField extends PureComponent {
     }
   }
 
+  renderValidationMessage = () => {
+    const { validationMessage } = this.props
+
+    if (!validationMessage) {
+      return null
+    }
+
+    return typeof validationMessage === 'string' ? (
+      <FormFieldValidationMessage marginTop={8}>
+        {validationMessage}
+      </FormFieldValidationMessage>
+    ) : (
+      validationMessage
+    )
+  }
+
   render() {
     const {
       hint,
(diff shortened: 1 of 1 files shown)


akleiner2:
Seems sane to me! My only comment is that `renderX` methods are a usually code smell to move `X` into its own component, but non-blocking :)

Chrischuck:
> Seems sane to me! My only comment is that `renderX` methods are a usually code smell to move `X` into its own component, but non-blocking :)

Hmm, fair point, in that case do you think  `FormFieldValidationMessage` should handle returning `null` if it's an empty string? I'm kind of leaning towards that idea actually since that would mean `FormFieldValidationMessage` handles all string types

jibberishclient:
Is there any update on this? Additionally I think there should be other composed form fields for selects, radios, and text areas just as there is for TextFieldInput such that it will make it easier for form libraries like formik and redux-form, etc. instead of having to destructure props to both the input itself and the wrapping form field.

Secondly on that note, I think the visibility of the validationMessage should rely entirely on isValid. In terms of UX, you're only going to want to show an error message if the input is also styled with a danger/warning intent. Thus it makes sense to have validationMessage render conditionally based on isValid regardless of its value such that the both of them are "in sync"

mshwery:
> Hmm, fair point, in that case do you think `FormFieldValidationMessage` should handle returning `null` if it's an empty string? I'm kind of leaning towards that idea actually since that would mean `FormFieldValidationMessage` handles all string types

I think that'd be fine as long as we can guarantee that having that component will never add an empty styled div. We also want to make sure it's obvious to wherever the component is used.

jibberishclient:
Will this be available in yarn? Or do we have to wait until a new release?

PR #785, opened 2020-04-17
add codemod for icon swapping

mshwery:
This PR introduces a codemod for transition from `<Icon icon="foo" />` to tree-shakeable imports like `<FooIcon />`

It relies on jest for tests, so we may consider switching from ava to jest.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 package.json
+/codemods/
--- a/package.json
+++ b/package.json
@@ -23,11 +23,12 @@
     "commonjs",
     "esm",
     "umd",
-    "index.d.ts"
+    "index.d.ts",
+    "codemods"
   ],
   "sideEffects": false,
   "scripts": {
-    "test": "xo && ava",
+    "test": "xo && jest --bail && ava",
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -58,7 +59,7 @@
     "fuzzaldrin-plus": "^0.6.0",
     "glamor": "^2.20.40",
     "lodash.debounce": "^4.0.8",
-    "lodash.mapvalues": "^4.6.0",
```
mshwery, line 61: unused

.prettierignore
```diff
@@ -1 +1,2 @@
 package.json
+/codemods/
```
akleiner2, line 2: Any reason we _don't_ want to have prettier run on the codemods dir?

package.json
akleiner2: 👍

.prettierignore
mshwery: Yes – well mainly the test fixtures – I couldn't get the test runner to apply our prettier/babel config so the output matched properly. Any tips here?

PR #786, opened 2020-04-17
Add resize prop

jibberishclient:
I think TextAreas should have the ability to resize, in most cases it will be full width with the ability to resize vertically. I think this should be accepted as a prop rather than making the user specify it in the styles themselves.

### DIFF
--- a/src/textarea/src/Textarea.js
+++ b/src/textarea/src/Textarea.js
@@ -10,6 +10,12 @@ class Textarea extends PureComponent {
      * Composes the Text component as the base.
      */
     ...Text.propTypes,
+    
+    /**
+     * Allows resize style of the textarea
+     */
+    
+    resize: PropTypes.string,
 
     /**
      * Makes the textarea element required.
@@ -70,7 +76,8 @@ class Textarea extends PureComponent {
     disabled: false,
     isInvalid: false,
     spellCheck: true,
-    grammarly: false
+    grammarly: false,
+    resize: "vertical"
   }
 
   static styles = {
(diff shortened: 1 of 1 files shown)


jibberishclient:
@Chrischuck Kinda new to PRs, can you take a look?

src/textarea/src/Textarea.js
```diff
@@ -116,6 +123,9 @@ class Textarea extends PureComponent {
         {...(disabled ? { color: 'muted' } : {})}
         {...Textarea.styles}
         {...props}
+        style={{
+          resize: resize      
+        }}
```
mshwery, line 128: Ideally this should be handled by `ui-box` not Evergreen. That way it's handled anytime you use a `textarea` not just Evergreen's `<Textarea />` component.

jibberishclient: As far as I could tell, ui-box doesn't include a resize prop

mshwery: You are correct! I'm suggesting we add it there.

jibberishclient: Ya for sure, that makes sense

mshwery:
Closing this as we should target `ui-box` for supporting `resize` as a css prop.

mshwery:
@jibberishclient you should be to upgrade ui-box directly if you want now! 
https://github.com/segmentio/ui-box/releases/tag/v3.3.0

Will work on upgrading in `evergreen-ui` eventually :)

PR #788, opened 2020-04-18
Removes fixed height on badge

### LINKED ISSUE #776: Badge Superscript/Subscript
The badge doesn't appear to handle badges correctly, when dealing with subscripts and superscripts? Can you guys look into this? I'd definite like to use superscripts and subscripts in my badges.

image

Chrischuck:
Fixes #776 

Allows badge height to grow based on text inside so subscripts and super scripts don't look ugly.

Screen Shot 2020-04-18 at 16 12 06

Only caveat is if you try to mix badges with + without subscripts, it'll look funky due to the height difference.

### DIFF
--- a/src/badges/src/Badge.js
+++ b/src/badges/src/Badge.js
@@ -33,7 +33,7 @@ class Badge extends PureComponent {
   static styles = {
     display: 'inline-block',
     boxSizing: 'border-box',
-    height: 16,
+    minHeight: 16,
     paddingTop: 0,
     paddingRight: 6,
     paddingBottom: 0,


src/badges/src/Badge.js
```diff
@@ -33,7 +33,7 @@ class Badge extends PureComponent {
   static styles = {
     display: 'inline-block',
     boxSizing: 'border-box',
-    height: 16,
+    minHeight: 16,
```
mshwery, line 36: What do you think of using `lineHeight` or flex to solve for this?

Chrischuck: Hmm, messing with line hight (setting it to 1.1) worked great for where there was a sup/sub tag...but then it was off centered without those tags. I'm actually in favor of closing this PR and having the user pass down the line height style for when they want to use a sup/sub tag. What do you think?

Issue #690, opened 2019-12-07
Combobox deselcts on mobile device


clarkkozak:
- Issue and steps to reproduce.

- I'm using a Galaxy S7 and Free Adblocker Browser 72 and the following code example:
`
<Combobox
  items={ ['Autobiography', 'Anime', 'Biography', 'Classic', 'Comics', 'Crime', 'DIY (Do It Yourself)', 'Dystopia', 'Fantasy', 'Folktale', 'Historical', 'Horror', 'Humor', 'Legend', 'Memoir', 'Mystery', 'Mythology ', 'Personal Development', 'Pyschology', 'Political Science', 'Romance', 'Satire', 'Science fiction', 'Science', 'Short', 'Spirituality / Religion', 'Textbook / Reference Book', 'Thriller', 'Western', 'Other']}
  onChange={selected => console.log(selected)}
  placeholder="Genres"
  autocompleteProps={{
    // Used for the title in the autocomplete.
    title: 'Genres'
  }}
/>
`

I'm unsure if it happens on any other mobile device since I can only test on my own, yet I've replicated it on the Combobox example at https://evergreen.segment.com/components/combobox/

- Expected.
When selecting a value on a large combobox, I can scroll through the options on a mobile device

- Actual.
When scrolling through the options, it will select when the finger leaves a select box

mshwery:
Nice find! Is this specific to Evergreen or is it happening in `downshift` too (which is what Evergreen is wrapping here)?

juxley:
Hi @mshwery ! Thank you for your reply and patience.

I took a deeper look into this, and it seems to be specific to Evergreen. Here's an example of `downshift` working properly; although slightly different.

https://codesandbox.io/s/github/kentcdodds/downshift-examples/tree/master/?module=%2Fsrc%2Fordered-examples%2F02-complete-autocomplete.js&moduleview=1

juxley:
Hello again!

Here's where I think it goes wrong:

https://github.com/segmentio/evergreen/blob/e4e350239c782843d6c2ae7d68119a28a384bf8e/src/autocomplete/src/Autocomplete.js#L224

I tried playing with the events to see if anything would change yet I couldn't figure it out.

mshwery:
@juxley seems like we may need to remove `onMouseUp` and `onTouchEnd` to replace them with `onClick`?

mshwery:
Or maybe we should remove the `onMouseUp` and `onTouchEnd` altogether?

[...]

PR #793, opened 2020-04-21
fix touch events for autocomplete

mshwery:
This fixes autocomplete touch events (so you can scroll on touch devices without closing/selecting something).

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -176,9 +176,9 @@ export default class Autocomplete extends PureComponent {
     width,
     inputValue,
     highlightedIndex,
-    selectItemAtIndex,
     selectedItem,
-    getItemProps
+    getItemProps,
+    getMenuProps
   }) => {
     const {
       title,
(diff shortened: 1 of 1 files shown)


src/autocomplete/src/Autocomplete.js
```diff
@@ -176,9 +176,9 @@ export default class Autocomplete extends PureComponent {
     width,
     inputValue,
     highlightedIndex,
-    selectItemAtIndex,
```
akleiner2, line 179: is this just used internally? not super familiar with the downshift api (and can't super easily find v3 docs)

mshwery: This is exposed externally (as are most downshift props), but it gets captured by `...restDownshiftProps` when we receive props on `<Autocoomplete />` and pass those `...restDownshiftProps` to children as a function.

All that to say that this PR doesn't change the component API we expose, but we don't need to pluck it out explicitly anymore.

mshwery:
Fixes #690

PR #801, opened 2020-04-24
Fix types for `<TextareaField />`

akleiner2:
Forgot to include this as a part of #800 😬. Will follow-up and release a patch immediately.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1980,6 +1980,45 @@ export interface TextareaProps extends TextProps {
 export class Textarea extends React.PureComponent<TextareaProps> {
 }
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -9,7 +9,15 @@ import { DownshiftProps } from 'downshift'
 
 export { configureSafeHref, setClassNamePrefix } from 'ui-box'
 
-type PositionTypes = 'top' | 'top-left' | 'top-right' | 'bottom' | 'bottom-left' | 'bottom-right' | 'left' | 'right'
+type PositionTypes =
```
akleiner2, line 12: Whoa. Looks like Prettier went crazy on this file. Will revert.

PR #805, opened 2020-04-25
Adds more control features for corner dialog

Chrischuck:
Working with this component outside of simple notifications has been a bit of a painpoint in my projects.

One issue: if you have no title and no close icon, the `h4` and `div` are still rendered:

Screen Shot 2020-04-25 at 11 23 15

The following PR adds the ability to:

1. set props on the header component
2. set props on the content component
3. remove header completely

### DIFF
--- a/src/corner-dialog/src/CornerDialog.js
+++ b/src/corner-dialog/src/CornerDialog.js
@@ -39,9 +39,7 @@ const closeAnimation = css.keyframes('closeAnimation', {
 
 const animationStyles = {
   '&[data-state="entering"], &[data-state="entered"]': {
-    animation: `${openAnimation} ${ANIMATION_DURATION}ms ${
-      animationEasing.spring
-    } both`
+    animation: `${openAnimation} ${ANIMATION_DURATION}ms ${animationEasing.spring} both`
   },
   '&[data-state="exiting"]': {
     animation: `${closeAnimation} 120ms ${animationEasing.acceleration} both`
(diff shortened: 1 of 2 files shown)


src/corner-dialog/src/CornerDialog.js
```diff
@@ -39,9 +39,7 @@ const closeAnimation = css.keyframes('closeAnimation', {
 
 const animationStyles = {
   '&[data-state="entering"], &[data-state="entered"]': {
-    animation: `${openAnimation} ${ANIMATION_DURATION}ms ${
-      animationEasing.spring
-    } both`
+    animation: `${openAnimation} ${ANIMATION_DURATION}ms ${animationEasing.spring} both`
```
Chrischuck, line 42: Prettier auto-formatted this for me, I can change it back if we don't want any unrelated diffs :)

src/corner-dialog/src/CornerDialog.js
```diff
@@ -134,6 +137,16 @@ export default class CornerDialog extends PureComponent {
...
 
+    /**
+     * Props that are passed to the dialog content body.
+     */
+    contentProps: PropTypes.object,
+
+    /**
+     * Props that are passed to the dialog header.
+     */
+    headerProps: PropTypes.object,
```
mshwery, line 148: We may want to avoid creating more of these types of pass-through props to spread on children because it creates a tightly-coupled dependency on how consumers are using it – there's no way of knowing when you are making a breaking change that might impact those props.

In general I'd prefer greater control over the wrapper component's API with _less customization_ and when more customization is needed, we should export the child components, something like:

```js
import { CornerDialog } from 'evergreen-ui'

const MyCornerDialog = ({ title }) => (
  <CornerDialog isShown>
    <CornerDialog.Header {...headerStyles}>{title}</CornerDialog.Header>
    The content
  </CornerDialog>
)
```

Chrischuck: I see, makes sense!

Can do!

Chrischuck: @mshwery Would we be able to change this pr to add only the `hasHeader` prop so it matches parity w/ the modals?

Issue #807, opened 2020-04-27
Production build is broken


gohyifan:
Hi team evergreen, my production build is showing this error in the console but the development build works fine in localhost.

![error.png](https://i.imgur.com/mnaWFVI.png)

Steps to reproduce the bug:
1. Clone my repo here: https://github.com/gohyifan/openflow
2. Run `yarn install` and `yarn start`
3. See that it works fine in localhost.
4. Visit [production](https://openflow.herokuapp.com/) and see that it's broken.

Environment:
OS: macOS Catalina 10.15
Browser: Chrome
Evergreen version: 4.27.2
React boilerplate: create-react-app

mshwery:
@gohyifan if you use v4.25.0 or earlier, does the same thing happen?

mshwery:
I was able to pinpoint this to something between v4.25.0-v4.26.0-0 but it doesn't affect all production builds. For instance, we're using v4.27.2 in production and the optimized webpack build is fine.

mshwery:
Resolved by https://github.com/segmentio/evergreen/pull/809

mshwery:
I clone your repo and upgraded to `v4.27.3` and was able `yarn build && serve -s build`

gohyifan:
`v4.27.3` worked, thanks @mshwery !

PR #808, opened 2020-04-27
Fix production build

mshwery:
Using #807 locally to test the production build of evergreen (via `yarn link`) I was able to narrow in on a circular dependency warnings for Icon and `defaultTheme`.

- [x] moved codemods deps into a separate package
- [x] upgraded a bunch of deps
- [x] added hooks linting
- [x] reverted theme helper changes that led to circular dependency


mshwery:
Closing in favor of a much smaller diff: #809

PR #809, opened 2020-04-27
revert circular dep

mshwery:
Using #807 locally to test the production build of evergreen (via yarn link) I was able to narrow in on a circular dependency warnings for Icon and defaultTheme.


mshwery:
It's annoying because now Alert/InlineAlert break tree-shaking, but enough of v4 already breaks it that this wont matter much. We'll have to come up with a non-theme way to get the right icons.

PR #817, opened 2020-05-07
[v5] move codemod deps

mshwery:
## Overview 

I moved all of the codemod deps out of the main evergreen-ui package.json and into their own so only if you want to use the codemod do you run the installer. 

Following up on: https://github.com/segmentio/evergreen/pull/785


mshwery:
Not worried about size-limit for v5, looks like it is failing from pulling in `master` actually (not this PR).

PR #819, opened 2020-05-07
[TagInput] Expose `tagSubmitKey` prop 

akleiner2:
## Overview 
This PR exposes a new `tagSubmitKey` prop which takes in two values either `"enter"` (default), or `"space"`. The motivation behind this is to allow consumers to choose how they let users submit tags as they're typing (with the hunch that `space` has some utility here).
 
N/A. Interaction change, not visual.

## Testing

### DIFF
--- a/src/tag-input/src/TagInput.js
+++ b/src/tag-input/src/TagInput.js
@@ -15,8 +15,8 @@ import Tag from './Tag'
 let inputId = 1
 
 const GET_KEY_FOR_TAG_DELIMITER = {
-  Enter: 'Enter',
-  Space: ' '
+  enter: 'Enter',
+  space: ' '
 }
 
 class TagInput extends React.Component {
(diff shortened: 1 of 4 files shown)


src/tag-input/src/TagInput.js
```diff
@@ -158,7 +166,10 @@ class TagInput extends React.Component {
   handleKeyDown = event => {
     const { selectionEnd, value } = event.target
 
-    if (event.key === 'Enter') {
+    const key = GET_KEY_FOR_TAG_DELIMITER[this.props.tagSubmitKey]
+
+    if (event.key === key) {
+      console.log(event.key)
```
mshwery, line 172: remove `console.log`

akleiner2: oof -- good catch! we should lint for these things 😬

PR #829, opened 2020-05-18
Convert <Radio />, <Checkbox />, and <Card /> to FC + Hooks

akleiner2:
## Overview 
Converts to aforementioned components to use forwardRef + memo + React.FC + useTheme. 

Also, makes the following API modifications: 
* Removes `innerRef` support on `<Checkbox />` (this can be implemented in consumer-space) 
* Removes the additional `onChange` parameter in `<Radio />`
 
image

image

## Testing

### DIFF
--- a/src/checkbox/src/Checkbox.js
+++ b/src/checkbox/src/Checkbox.js
@@ -49,10 +49,10 @@ const Checkbox = memo(
     } = props
 
     useEffect(() => {
-      if (ref?.current) {
+      if (ref && ref.current) {
         ref.current.indeterminate = indeterminate
       }
-    }, [ref, ref?.current, indeterminate])
+    }, [ref, indeterminate])
 
     const theme = useTheme()
     const themedClassName = theme.getCheckboxClassName(appearance)
--- a/src/checkbox/src/Checkbox.js
+++ b/src/checkbox/src/Checkbox.js
@@ -1,4 +1,4 @@
-import React, { memo, forwardRef } from 'react'
+import React, { memo, forwardRef, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import Box, { spacing, position, layout, dimensions } from 'ui-box'
 import { Text } from '../../typography'
@@ -45,12 +45,16 @@ const Checkbox = memo(
       onChange,
       value,
       indeterminate,
-      innerRef,
       ...rest
     } = props
 
-    const theme = useTheme()
+    useEffect(() => {
+      if (ref?.current) {
+        ref.current.indeterminate = indeterminate
+      }
+    }, [ref, ref?.current, indeterminate])
 
+    const theme = useTheme()
     const themedClassName = theme.getCheckboxClassName(appearance)
 
     return (
@@ -73,7 +77,7 @@ const Checkbox = memo(
           onChange={onChange}
           disabled={disabled}
           aria-invalid={isInvalid}
-          ref={ref}
+          innerRef={ref}
         />
         <Box
           boxSizing="border-box"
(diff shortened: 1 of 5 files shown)


src/checkbox/src/Checkbox.js
```diff
@@ -162,7 +73,7 @@ class Checkbox extends PureComponent {
           onChange={onChange}
           disabled={disabled}
           aria-invalid={isInvalid}
-          innerRef={this.handleInnerRef}
+          ref={ref}
```
akleiner2, line 76: Now that we have `forwardRef` here, we drop support for the `handleInnerRef` -- is there a better way to go about this @mshwery

src/radio/src/Radio.js
```diff
@@ -119,18 +29,24 @@ class Radio extends PureComponent {
...
       appearance,
-      ...props
-    } = this.props
+      ...rest
+    } = props
+    const theme = useTheme()
     const themedClassName = theme.getRadioClassName(appearance)
 
+    const handleChange = useCallback(
+      event => onChange(event, event.target.checked),
```
mshwery, line 38: Curious if we should stop shimming the event handler?

Why don't we let consumers pass in a normal `onChange` handler?

src/checkbox/src/Checkbox.js
mshwery: Rule of thumb:

1. `Box` does not support `ref` yet. If adding a ref to a `Box` component you need to pass it to `innerRef={ref}`
2. If adding a ref to any internal Evergreen component we should use `ref` (they are all either `PureComponent` or `forwardRef` components).

mshwery: As for your question there are two ways to keep the behavior intact:

Keep the `handleInnerRef` shim:
```
  handleInnerRef = el => {	
    if (el) {	
      el.indeterminate = indeterminate	
    }

    // manually update the ref (depending on whether it's a legacy ref or `createRef`
    if (ref) { ... }
  }
```

Use `useEffect` to trigger an update to set the `inputInstance.indeterminate` value:
```js
useEffect((
  if (ref?.current && props.indeterminate) {
    ref.current.indeterminate = true
  }
), [ref, ref?.current, props.indeterminate])
```

src/radio/src/Radio.js
akleiner2: Unsure -- I was surprised to see this API different than the one on `<Checkbox />` for example. Since this is going to v5, I'm fine just removing this and letting this and `<Checkbox />` have the same behavior.

mshwery: 👍 would be happy with normal event handlers everywhere, for consistency

akleiner2: +1

src/checkbox/src/Checkbox.js
```diff
@@ -136,10 +45,16 @@ class Checkbox extends PureComponent {
...
       value,
       indeterminate,
-      innerRef,
-      ...props
-    } = this.props
+      ...rest
+    } = props
+
+    useEffect(() => {
+      if (ref?.current) {
```
mshwery, line 52: Oh shoot... this won't compile? I was stuck in TypeScript mode. We will have to use `ref && ref.current` instead.

src/checkbox/stories/index.stories.js
```diff
@@ -9,6 +9,25 @@ const innerRef = el => {
   }
 }
 
+function IndeterminateWithRefExample() {
+  useEffect(() => {
+    if (ref?.current) {
+      innerRef(ref.current)
+    }
+  }, [ref, ref?.current])
```
mshwery, line 17: We should update this to drop the `?.`

src/checkbox/src/Checkbox.js
akleiner2: Ah yeah. Might be worth upgrading Babel to support optional chaining?

PR #830, opened 2020-05-18
Convert <Dialog /> and <CornerDialog /> to FC + memo + forwardRef

akleiner2:
## Overview 
Do as the titles do. Also heavily memoizes handlers with `useCallback`.

image

image

## Testing

### DIFF
--- a/src/corner-dialog/src/CornerDialog.js
+++ b/src/corner-dialog/src/CornerDialog.js
@@ -1,10 +1,4 @@
-import React, {
-  memo,
-  forwardRef,
-  useState,
-  useEffect,
-  useCallback
-} from 'react'
+import React, { memo, useState, useEffect, useCallback } from 'react'
 import { css } from 'glamor'
 import PropTypes from 'prop-types'
 import Transition from 'react-transition-group/Transition'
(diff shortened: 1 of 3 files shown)


src/corner-dialog/src/CornerDialog.js
```diff
@@ -222,28 +67,75 @@ export default class CornerDialog extends PureComponent {
...
-          onExited={this.handleExited}
+          onExited={handleExited}
           onEntered={onOpenComplete}
         >
           {state => (
             <Card
               role="dialog"
               backgroundColor="white"
               elevation={4}
+              ref={ref}
```
mshwery, line 138: Actually not sure we need a ref here... What would you be attempting to do with a ref for a dialog?

src/dialog/src/Dialog.js
```diff
@@ -398,6 +208,7 @@ class Dialog extends React.Component {
             flexDirection="column"
             css={animationStyles}
             data-state={state}
+            ref={ref}
```
mshwery, line 211: Same Q here.

src/corner-dialog/src/CornerDialog.js
akleiner2: Good q! I think these actually could do without `forwardRef` support, since we're not spreading props to the underlying element. Will update.

PR #831, opened 2020-05-18
Convert <Image /> to React.FC + forwardRef + memo

akleiner2:
## Overview 
Tiny one -- low hanging fruit.
 
N/A

## Testing

### DIFF
--- a/src/image/src/Image.js
+++ b/src/image/src/Image.js
@@ -12,3 +12,5 @@ Image.propTypes = {
   ...Box.propTypes,
   src: PropTypes.string
 }
+
+export default Image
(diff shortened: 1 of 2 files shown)


src/image/src/Image.js
```diff
@@ -1,14 +1,16 @@
...
+    return <Box is="img" {...props} innerRef={ref} />
+  })
+)
 
-  render() {
-    return <Box is="img" {...this.props} />
-  }
+Image.propTypes = {
+  ...Box.propTypes,
+  src: PropTypes.string
```
mshwery, line 13: Should we require `alt` tag as a prop?

akleiner2: Think so! Can probably surface API changes later :)

PR #833, opened 2020-05-21
Fix Combobox autocompleteProps typing

leesiongchan:
Since some of the props are being used by the combobox, is that safe for us to omit them from  `autocompleteProps`? Now I have a problem with `autocompleteProps.items` and `autocompleteProps.children` because they are required fields.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -652,7 +652,7 @@ export interface ComboboxProps extends React.ComponentPropsWithoutRef<typeof Box
   /**
    * Properties forwarded to the autocomplete component. Use with caution.
    */
-  autocompleteProps?: AutocompleteProps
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -652,7 +652,7 @@ export interface ComboboxProps extends React.ComponentPropsWithoutRef<typeof Box
   /**
    * Properties forwarded to the autocomplete component. Use with caution.
    */
-  autocompleteProps?: AutocompleteProps
+  autocompleteProps?: Omit<AutocompleteProps, 'items' | 'selectedItem' | 'onChange' | 'itemToString' | 'children'>
```
mshwery, line 655: Can you describe why we would omit these types from `autocompleteProps`? All of these values are accepted and would in fact take precedent over the top-level props that match.

leesiongchan: Because `items` is required, so I to use both `autocompleteProps`'s `items` and `ComboboxProps`'s `items`

PR #835, opened 2020-05-21
Fix autocomplete and combobox typing to use generic

leesiongchan:
Use generic for `item` instead of `any` type in Autocomplete & Combobox

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -458,13 +458,13 @@ export class AutocompleteItem extends React.PureComponent<AutocompleteItemProps>
 }
 
 // https://github.com/downshift-js/downshift
-export interface AutocompleteProps extends Omit<DownshiftProps<any>, 'children'> {
+export interface AutocompleteProps<T> extends Omit<DownshiftProps<T>, 'children'> {
   // @deprecated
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -458,13 +458,13 @@ export class AutocompleteItem extends React.PureComponent<AutocompleteItemProps>
 }
 
 // https://github.com/downshift-js/downshift
-export interface AutocompleteProps extends Omit<DownshiftProps<any>, 'children'> {
+export interface AutocompleteProps<T> extends Omit<DownshiftProps<T>, 'children'> {
   // @deprecated
-  defaultSelectedItem?: string
```
akleiner2, line 463: To not lose any of the existing type safety here, do you think we could do something like 

`export interface AutocompleteProps<T extends string = string>` ? That way, the fallback type (if not specified) will be `string`, which preserves the existing behavior. 

More broadly, I've been trying to think about good ways for consumers to use components that consume the generics as-is. There is support for generics in JSX, but the syntax is pretty gnarly. Maybe a class method that generates a strongly-typed instance consuming the generic type? 

Curious if you have any thoughts here!

PR #836, opened 2020-05-21
Convert <Switch />, <SelectMenu /> to hooks + memo + forwardRef

akleiner2:
## Overview 
Relatively mechanical conversion. Comments inline. 
 
From Storybook: 
image

image

## Testing

### DIFF
--- a/src/select-menu/src/SelectMenu.js
+++ b/src/select-menu/src/SelectMenu.js
@@ -1,4 +1,4 @@
-import React, { memo } from 'react'
+import React, { memo, forwardRef } from 'react'
 import PropTypes from 'prop-types'
 import arrify from 'arrify'
 import { Popover } from '../../popover'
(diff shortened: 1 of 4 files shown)


src/select-menu/src/SelectMenu.js
```diff
@@ -8,158 +8,44 @@ import SelectMenuContent from './SelectMenuContent'
...
-      return {
-        detailView: detailView({ close })
-      }
-    }
+  return {}
+}
 
-    if (detailView) {
-      return { detailView }
+const getEmptyView = (close, emptyView) => {
```
akleiner2, line 25: Could probably memoize / inline this and `getDetailView`

src/select-menu/src/SelectMenu.js
```diff
@@ -192,23 +78,139 @@ export default class SelectMenu extends PureComponent {
...
-            {...this.getDetailView(close, detailView)}
-            {...this.getEmptyView(close, emptyView)}
+            {...getDetailView(close, detailView)}
+            {...getEmptyView(close, emptyView)}
             closeOnSelect={closeOnSelect}
           />
         )}
-        {...props}
+        {...rest}
+        ref={ref}
```
akleiner2, line 93: `<Popover />` will be a really tricky one to convert, since so much stuff is reliant upon it. Right now, it actually swallows refs, so this is a no-op. This just preserves the same prop behavior as before.

src/switch/src/Switch.js
```diff
@@ -172,10 +69,27 @@ class Switch extends PureComponent {
...
+      value => {
+        if (checkedProps) {
+          onChange(value)
+        } else {
+          setIsChecked(checked => !checked)
+          onChange(value)
+        }
+      },
+      [onChange]
+    )
```
mshwery, line 90: This is a bit weird that it is both... controlled and uncontrolled? (e.g. passing in `checked` but also.. state is internally handled?)

src/select-menu/src/SelectMenu.js
mshwery: Do we want to avoid ref forwarding in this component then?

src/switch/src/Switch.js
akleiner2: Agreed! -- preserving old behavior 🤷‍♂️

mshwery: Should we break from it? (can do in a follow up PR, but add to the v5 roadmap tracker)

src/select-menu/src/SelectMenu.js
akleiner2: Probably. I wanted to decouple API changes from the conversion piece just to make it a little easier to read

akleiner2: But definitely worth a follow-up. Might make an `API Changes` section in the v5 tracker

src/switch/src/Switch.js
akleiner2: Definitely should. Imo no sense in having an uncontrolled component, especially iven that other inputs are controlled 🤷‍♂️

PR #837, opened 2020-05-22
Convert <Combobox />, <RadioGroup />, <(Un)orderedList /> to fc + memo + forwardRef

akleiner2:
## Overview 
Converts the aforementioned to fc + memo + forwardRef. Also does some light cleanup: 
* Adds + refines types for `<OrderedList />` + `<Ol />`
* Aliases `Ul` and `UnorderedList` as referencing the same component props (no more unknown!)
* Removes the ability for `<RadioGroup />` to hold its own state (no more debates around controlled vs. uncontrolled)> 
 
**`<Ul />`**: 
image

**`<Combobox />`**: 
image

**`<RadioGroup />`**: 
image

**`<Ol />`**: 
image

## Testing

### DIFF
--- a/src/radio/src/RadioGroup.js
+++ b/src/radio/src/RadioGroup.js
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react'
+import React, { memo, forwardRef, useRef, useCallback } from 'react'
 import PropTypes from 'prop-types'
 import { spacing, position, layout, dimensions } from 'ui-box'
 import { Pane } from '../../layers'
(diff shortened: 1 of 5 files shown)


src/radio/src/RadioGroup.js
```diff
@@ -99,14 +17,24 @@ export default class RadioGroup extends PureComponent {
...
       onChange,
       isRequired,
-      ...props
-    } = this.props
+      ...rest
+    } = props
+
+    const name = useRef()
+    name.current = `RadioGroup-${radioCount}`
+    radioCount += 1
```
mshwery, line 25: Wont this produce new names on re-renders?

akleiner2: Ah yeah -- I'll move to a ref. Good spot.

akleiner2: yeah -- good spot. should just update / set the ref on mount

mshwery, line 25:
```suggestion
    const [name] = useState(`RadioGroup-${radioCount++}`)
```

PR #838, opened 2020-05-22
Align Checkbox and Switch props a little more closely

akleiner2:
## Overview 
This PR removes the optionality for `onChange` in `<Switch />` (forcing developers to wire up state for it themselves), and cleans up an unused `innerRef` in `<Checkbox />` (now that we're slowly supporting ref-forwarding everywhere in EG.
 
## Testing

### DIFF
--- a/src/switch/src/Switch.js
+++ b/src/switch/src/Switch.js
@@ -1,4 +1,4 @@
-import React, { memo, forwardRef, useCallback } from 'react'
+import React, { memo, forwardRef } from 'react'
 import { css } from 'glamor'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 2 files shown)


src/switch/src/Switch.js
```diff
@@ -63,33 +63,19 @@ const Switch = memo(
...
-        if (checkedProps) {
-          onChange(value)
-        } else {
-          setIsChecked(checked => !checked)
-          onChange(value)
-        }
-      },
-      [onChange]
-    )
+    const handleChange = useCallback(event => onChange(event), [onChange])
```
mshwery, line 77: This can go away too

akleiner2: True! Good spot

PR #839, opened 2020-05-22
Convert <Menu /> and its subcomponents to fc + hooks

akleiner2:
## Overview 
Converts and lightly polishes up the APIs for `<Menu />` and its subcomponents. I'm still not _entirely_ sure the best way to type this, but I think it's a start and is somewhat close. 
 
image

## Testing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1094,9 +1094,9 @@ declare const MenuItem: React.FC<MenuItemProps>
 declare const MenuDivider: React.FC<{}>
 declare const MenuGroup: React.FC<MenuGroupProps>
 declare const MenuOption: React.FC<MenuOptionProps>
-declare const MenuOptionsGroup: React.FC<MenuOptionsGroupProps<T>>
+declare class MenuOptionsGroup<T> extends React.FC<MenuOptionsGroupProps<T>> {}
 
-export class Menu extends React.PureComponent<MenuProps> {
+export declare class Menu extends React.FC<MenuProps> {
   // @ts-ignore
   public static Item = MenuItem
   // @ts-ignore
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -1090,22 +1090,23 @@ export interface MenuOptionsGroupProps<T> {
   options: Array<{ value: T, label: string }>
 }
 
-export class Menu extends React.PureComponent<MenuProps> {
+declare const MenuItem: React.FC<MenuItemProps>
+declare const MenuDivider: React.FC<{}>
+declare const MenuGroup: React.FC<MenuGroupProps>
+declare const MenuOption: React.FC<MenuOptionProps>
+declare class MenuOptionsGroup<T> extends React.FC<MenuOptionsGroupProps<T>> {}
```
akleiner2, line 1097: Couldn't figure out a quick way to get static members to respect static members.

index.d.ts
```diff
@@ -1090,22 +1090,23 @@ export interface MenuOptionsGroupProps<T> {
...
+
+export declare class Menu extends React.FC<MenuProps> {
   // @ts-ignore
-  public static Item = class MenuItem extends React.PureComponent<MenuItemProps> {
-  }
```
akleiner2, line 1096: Is there a better way to declare additional properties / members on a function, and have them be respected by JSX?

akleiner2: Haven't yet given it a spin -- the TS compiler didn't complain (hovering over `Menu` shows that it has the right types), but lemme give it a spin

src/menu/src/Menu.js
```diff
@@ -7,112 +7,114 @@ import MenuGroup from './MenuGroup'
...
-  render() {
-    const { children } = this.props
-    return (
-      <Pane is="nav" innerRef={this.onMenuRef} role="menu" outline="none">
-        {children}
-      </Pane>
-    )
-  }
+    }
+  }, [menuRef])
```
mshwery, line 95: This will only trigger a re-run when the actual ref object changes, not when `menuRef.current` changes. Is that intended?

PR #845, opened 2020-05-26
update types to avoid React.ComponentProps with Box

mshwery:
## Overview 
This PR removes our use of `typeof Box` which breaks React typings because Box has custom types to support generics (React types don't support generic inference).
 
NA

## Testing


mshwery:
I think this may break any consumers that pass the `is` prop though. Might need to make it more generic like `Box<any>` :(

mshwery:
I'd like to get typing tests in place. // cc @colinlohner

PR #849, opened 2020-05-29
Convert portal to fc

Chrischuck:
I believe this will carry over the previous behavior. Biggest thing was there was a lot of logic in the constructor and this component has to return null if it's rendered on the server for SSR.

Logic basically follows:
1. Initialize portal container + el in `useState` which should carry over the same behavior as having it in the constructor (returning null if no dom is accessible) 
2. When component mounts (client side behavior only), append portal + el to the dom. At this point, portal + el should have been initialized
3. sanity check if the dom is there before returning react portal element
4. clean up in useEffect but ripping portal from the dom

Haven't tested this in SSR, storybook works though :)

### DIFF
--- a/src/portal/src/Portal.js
+++ b/src/portal/src/Portal.js
@@ -1,38 +1,50 @@
-import { Component } from 'react'
+import { useState, useEffect, memo } from 'react'
 import canUseDom from 'dom-helpers/util/inDOM'
 import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
 
-let portalContainer
+const initializePortal = () => {
+  if (!canUseDom) {
+    return null
+  }
 
-export default class Portal extends Component {
-  constructor() {
-    super()
+  const portalContainer = document.createElement('div')
+  portalContainer.setAttribute('evergreen-portal-container', '')
+  return portalContainer
+}
 
-    // This fixes SSR
-    if (!canUseDom) return
+const initializeEl = () => {
+  if (!canUseDom) {
+    return null
+  }
 
-    if (!portalContainer) {
-      portalContainer = document.createElement('div')
-      portalContainer.setAttribute('evergreen-portal-container', '')
+  return document.createElement('div')
+}
+
+const Portal = memo(({ children }) => {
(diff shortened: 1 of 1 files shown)


src/portal/src/Portal.js
```diff
@@ -1,38 +1,50 @@
...
+  }
+
+  return document.createElement('div')
+}
 
-    // This fixes SSR
-    if (!canUseDom) return
+const Portal = memo(({ children }) => {
+  const [portalContainer] = useState(initializePortal)
+  const [el] = useState(initializeEl)
```
mshwery, line 26: Wouldn't this require `initializePortal()` and `initializeEl()` to be called?

mshwery: Also, won't this cause `initializePortal()` to get called every time there's a Portal component mounted? We don't want multiple divs for the portal, just one.

Chrischuck: For your first comment, yeah I think to have the same exact behavior as the constructor we need to call the functions. I didn't know passing a function would lazily set the state until now!

Chrischuck: Screen Shot 2020-05-29 at 08 42 28

that looks better :)

src/portal/src/Portal.js
```diff
@@ -1,38 +1,56 @@
...
-    if (!portalContainer) {
-      portalContainer = document.createElement('div')
-      portalContainer.setAttribute('evergreen-portal-container', '')
+const Portal = memo(({ children }) => {
+  const [portalContainer] = useState(initializePortal())
+  const [el] = useState(initializeEl())
+
+  useEffect(() => {
+    if (portalContainer) {
       document.body.appendChild(portalContainer)
```
mshwery, line 36: We only want to add the container to the DOM once. This will add it every time a new Portal is mounted.

Chrischuck: hmm, what do you think the best way to solve this would be? Simple DOM check to see if it exists before appending?

mshwery: Move this to the `initializeContainer` function only when `portalContainer` isn't already defined.

PR #862, opened 2020-06-25
Convert the Table components to FC/Hooks

colinlohner:
## Overview 
Table and all sub components converted to use hooks and memo.
 
## Screenshots
localhost_6006__path=_story_image--image(Laptop with HiDPI screen)
localhost_6006__path=_story_image--image(Laptop with HiDPI screen) (1)
localhost_6006__path=_story_image--image(Laptop with HiDPI screen) (2)

## Testing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1349,8 +1349,7 @@ export interface SearchTableHeaderCellProps extends Omit<TableHeaderCellProps, '
   icon?: React.ReactElement | null | false
 }
 
-export class SearchTableHeaderCell extends React.PureComponent<SearchTableHeaderCellProps> {
-}
+export declare const SearchTableHeaderCell: ForwardRefComponent<SearchTableHeaderCellProps>
 
 export interface SegmentedControlProps extends Omit<React.ComponentPropsWithoutRef<typeof Box>, 'defaultValue' | 'onChange'> {
   options: Array<{ label: string, value: NonNullable<SegmentedControlProps['value']> }>
@@ -1631,8 +1630,7 @@ export declare const Switch: ForwardRefComponent<SwitchProps>
 export interface TableBodyProps extends PaneProps {
 }
 
-export class TableBody extends React.PureComponent<TableBodyProps> {
-}
+export declare const TableBody: React.FC<TableBodyProps>
 
 export interface TableCellProps extends PaneProps {
   /**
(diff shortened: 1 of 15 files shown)


index.d.ts
```diff
@@ -1700,15 +1697,15 @@ interface TableEditableCellProps extends Omit<TextTableCellProps, 'placeholder'
...
+export class TableHeaderCell extends React.FC<TableHeaderCellProps> {
 }
 
 export interface TableHeadProps extends PaneProps {
   height?: number | string
   accountForScrollbar?: boolean
 }
 
-export class TableHead extends React.PureComponent<TableHeadProps> {
+export class TableHead extends React.FC<TableHeadProps> {
```
mshwery, line 1708: If any of these are using memo or forwardRef you should use `ForwardRefComponent` custom type.

These aren't classes, so you'll want something like this:
https://github.com/segmentio/evergreen/blob/v5/index.d.ts#L1118

mshwery: If you really just want React.FC you dont need `class Foo extends`

src/table/src/EditableCell.js
```diff
@@ -1,190 +1,168 @@
...
+    disabled,
+    placeholder,
+    isSelectable = true,
+    textProps = {},
+    autoFocus = false,
+    ...rest
+  } = props
+
+  let cursor = 'text'
+  let mainRef
```
mshwery, line 23: Would put this in state instead.

src/table/src/EditableCell.js
```diff
@@ -1,190 +1,168 @@
...
-  handleFieldCancel = () => {
-    this.setState({ isEditing: false })
+  const handleFieldCancel = () => {
+    setIsEditing(false)
   }
 
-  handleClick = () => {
-    this.mainRef.focus()
+  const handleClick = () => {
+    mainRef.focus()
```
mshwery, line 77: This needs a `if (mainRef) {`

src/table/src/SelectMenuCell.js
```diff
@@ -1,197 +1,166 @@
...
+  const onResizeHandler = () => {
+    updateOnResize()
   }
 
-  constructor(props) {
-    super(props)
-    this.onResize = debounce(this.onResize, 200)
-  }
+  const onResize = debounce(onResizeHandler, 200)
+  let mainRef
```
mshwery, line 32: Same here, `useState` for any refs

colinlohner:
Gonna split this up into separate PRs for the different sub components of Table

PR #863, opened 2020-06-25
Convert Segmented Control to Hooks/FC

colinlohner:
## Overview 
Updated the Segmented Control component to use memo and hooks
 
## Screenshots
localhost_6006__path=_story_segmented-control--segmentedcontrol(Laptop with HiDPI screen)

## Testing
- [x] Works as a controlled component
- [x] Works as an uncontrolled component

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1361,8 +1361,7 @@ export interface SegmentedControlProps extends Omit<React.ComponentPropsWithoutR
   height?: number
 }
 
-export class SegmentedControl extends React.FC<SegmentedControlProps> {
-}
+export declare const SegmentedControl: ForwardRefComponent<SegmentedControlProps>
 
 export interface SelectProps extends Omit<React.ComponentPropsWithoutRef<typeof Box>, 'onChange'> {
   /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -1361,7 +1361,7 @@ export interface SegmentedControlProps extends Omit<React.ComponentPropsWithoutR
   height?: number
 }
 
-export class SegmentedControl extends React.PureComponent<SegmentedControlProps> {
+export class SegmentedControl extends React.FC<SegmentedControlProps> {
 }
 
 export interface SelectProps extends Omit<React.ComponentPropsWithoutRef<typeof Box>, 'onChange'> {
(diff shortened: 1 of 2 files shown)


src/segmented-control/src/SegmentedControl.js
```diff
@@ -1,144 +1,137 @@
...
-            disabled={disabled}
-          />
-        ))}
-      </Box>
-    )
+    safeInvoke(onChange, newValue)
   }
+
+  return (
+    <Box display="flex" marginRight={-1} height={height} {...rest}>
```
mshwery, line 54: Do we want/need to support ref forwarding on the outer Box component?

src/segmented-control/src/SegmentedControl.js
```diff
@@ -1,144 +1,137 @@
...
+    height = 32,
+    options,
+    onChange,
+    defaultValue,
+    disabled,
+    ...rest
+  } = props
+
+  const groupName = `SegmentedControl-${radioCount}`
+  radioCount += 1
```
mshwery, line 22: This will regen the groupName and increment the counter on every render. Check out this for a good example of solving for this so it only ever increments on _mount_:

https://sourcegraph.segment.com/github.com/segmentio/evergreen@v5/-/blob/src/text-input/src/TextInputField.js?subtree=true#L11:66

index.d.ts
```diff
@@ -1361,7 +1361,7 @@ export interface SegmentedControlProps extends Omit<React.ComponentPropsWithoutR
   height?: number
 }
 
-export class SegmentedControl extends React.PureComponent<SegmentedControlProps> {
+export class SegmentedControl extends React.FC<SegmentedControlProps> {
```
mshwery, line 1364: I don't think we want these to be `class Foo extends`...?

akleiner2, line 1364: `export declare const SegmentedControl: ForwardRefComponent<>`?

akleiner2: Oh I see there's no ref support, this should just be `export const SegmentedControl: React.FC<>` maybe?

src/segmented-control/src/SegmentedControl.js
colinlohner: Might not be a bad idea to support. I'll add it in

PR #864, opened 2020-06-25
Convert Positioner component to Hooks/FC

colinlohner:
## Overview 
Convert the Positioner component to use forwardRef, memo, and hooks 

## Screenshots
Screen Recording 2020-06-25 at 2 40 57 PM
Screen Recording 2020-06-25 at 2 39 42 PM

## Testing
- [x] Popovers work as expected
- [x] Tooltips work as expected

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1182,7 +1182,7 @@ export interface PositionerProps {
   onOpenComplete?: () => void
 }
 
-export class Positioner extends React.PureComponent<PositionerProps> {
+export declare const Positioner: ForwardRefComponent<PositionerProps> {
 }
 
 export interface RadioProps extends Omit<BoxProps<'input'>, 'onChange'> {
--- a/index.d.ts
+++ b/index.d.ts
@@ -1182,8 +1182,7 @@ export interface PositionerProps {
   onOpenComplete?: () => void
 }
 
-export declare const Positioner: ForwardRefComponent<PositionerProps> {
-}
+export declare const Positioner: ForwardRefComponent<PositionerProps>
 
 export interface RadioProps extends Omit<BoxProps<'input'>, 'onChange'> {
   /**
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1182,7 +1182,7 @@ export interface PositionerProps {
   onOpenComplete?: () => void
 }
 
-export class Positioner extends React.PureComponent<PositionerProps> {
+export declare const Positioner: ForwardRefComponent<PositionerProps> {
 }
```
mshwery, line 1186:
```suggestion
export declare const Positioner: ForwardRefComponent<PositionerProps>
```

src/positioner/src/Positioner.js
```diff
@@ -34,236 +29,230 @@ const getCSS = ({ initialScale, animationDuration }) => ({
...
+    position = Position.BOTTOM,
+    bodyOffset = 6,
+    targetOffset = 6,
+    innerRef,
+    onOpenComplete = () => {},
+    onCloseComplete = () => {}
+  } = props
+
+  const [top, setTop] = useState(0)
+  const [left, setLeft] = useState(0)
```
mshwery, line 48: Should these be `null` to match the `initialState` from before?

src/positioner/src/Positioner.js
```diff
@@ -34,236 +29,230 @@ const getCSS = ({ initialScale, animationDuration }) => ({
...
+        cancelAnimationFrame(latestAnimationFrame)
+      }
+    }
+  }, [left, top, height, width, transformOrigin, positionerRef])
 
-  getTargetRef = ref => {
-    this.targetRef = ref
+  const getRef = (newRef) => {
+    setPositionerRef(newRef)
+    safeInvoke(innerRef, newRef)
```
mshwery, line 74: Can we drop the `innerRef` support?

src/positioner/src/Positioner.js
colinlohner: It could be, though it doesn't seem to have an effect (null vs 0). I like using 0, as its more descriptive of the values that will be set. Thoughts?

src/positioner/src/Positioner.js
colinlohner: sgtm

PR #875, opened 2020-07-12
switch to `emotion`

mshwery:
## Overview
This PR migrates off of glamor over to emotion, because the APIs are similar enough and glamor is unmaintained.

NA

## Testing


package.json
```diff
@@ -214,7 +215,7 @@
   "size-limit": [
     {
       "path": "commonjs/index.js",
-      "limit": "241 KB"
+      "limit": "260 KB"
```
akleiner2, line 218: Dang -- how come the bundle size bump?

src/dialog/src/Dialog.js
```diff
@@ -14,7 +14,7 @@ const animationEasing = {
 
 const ANIMATION_DURATION = 200
 
-const openAnimation = css.keyframes('openAnimation', {
+const openAnimation = keyframes({
```
akleiner2, line 17: Assuming these two calls are equivalent?

mshwery: Roughly, minus the first arg (optional name) which was removed in this PR.

package.json
mshwery: I think that `emotion` is actually smaller than glamor, but then adding `emotion-server` for SSR support bloats the commonjs bundle. Fortunately the esm bundle shouldn't be impacted since that will get tree-shaken.

mshwery:
Ugh, SSR works during `next` (dev mode) but fails in the production build. I get a flash of unstyled content. Resolving this is not straightforward as the all solutions outlined in the docs suggest we'd have to break how users use SSR with Evergreen today in non-trivial ways.

Looking at other options.

PR #877, opened 2020-07-13
V5 virtual body fc

colinlohner:
## Overview 
 Update virtual body to FC/Hooks

N/a

## Testing
- [x] Existing functionality remains

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1824,8 +1824,7 @@ export class Table extends React.PureComponent<TableProps> {
   public static Body = TableBody
 
   // @ts-ignore
-  public static VirtualBody = class VirtualBody extends React.PureComponent<TableVirtualBodyProps> {
-  }
+  public static VirtualBody: React.FC<TableVirtualBodyProps>
 
   // @ts-ignore
   public static Head = TableHead
--- a/index.d.ts
+++ b/index.d.ts
@@ -1823,7 +1823,6 @@ export class Table extends React.PureComponent<TableProps> {
   // @ts-ignore
   public static Body = TableBody
 
-  // @ts-ignore
   public static VirtualBody: React.FC<TableVirtualBodyProps>
 
   // @ts-ignore
--- a/src/table/src/TableVirtualBody.js
+++ b/src/table/src/TableVirtualBody.js
@@ -130,11 +130,11 @@ const TableVirtualBody = memo(props => {
 
     // Return if we are in a weird edge case in which the ref is no longer valid.
     if (paneRef && paneRef instanceof Node) {
-      const calculatedHeight = paneRef.offsetHeight
+      const tempCalculatedHeight = paneRef.offsetHeight
 
-      if (calculatedHeight > 0) {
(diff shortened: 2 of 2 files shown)


index.d.ts
```diff
@@ -1829,8 +1829,7 @@ export class Table extends React.PureComponent<TableProps> {
   public static Body = TableBody
 
   // @ts-ignore
```
mshwery, line 1831: do we need the `@ts-ignore` still?

src/table/src/TableVirtualBody.js
```diff
@@ -1,136 +1,84 @@
...
-    scrollOffset: PropTypes.number,
-    /**
-     * The scrollToAlignment property passed to react-tiny-virtual-list
-     */
-    scrollToAlignment: PropTypes.oneOf(['start', 'center', 'end', 'auto']),
-    /**
-     * The onScroll callback passed to react-tiny-virtual-list
-     */
-    onScroll: PropTypes.func
+const useForceUpdate = () => useState()[1]
```
mshwery, line 7: I dont understand this

mshwery: Are you trying something like this? https://stackoverflow.com/questions/53215285/how-can-i-force-component-to-re-render-with-hooks-in-react

colinlohner: essentially yes, just trying to force a rerender

index.d.ts
colinlohner: not sure, i'll make sure we do/dont

src/table/src/TableVirtualBody.js
mshwery: I'm not sure what you have will actually work. What you want is something like this:

```js
const useForceUpdate = () => {
  const [_, setValue] = useState()

  return useCallback(() => {
    setValue({}) // set with a new object instance every time it gets called
  }, [setValue])
}
```

colinlohner: It does seem to work, but if you think that what you have there would work out better, I can update it to that

src/table/src/TableVirtualBody.js
```diff
@@ -149,52 +97,44 @@ export default class TableVirtualBody extends PureComponent {
...
 
     // Simply return when we now the height of the pane is fixed.
-    if (this.state.isIntegerHeight) return
+    if (isIntegerHeight) return
 
     // Return if we are in a weird edge case in which the ref is no longer valid.
-    if (this.paneRef) {
-      const calculatedHeight = this.paneRef.offsetHeight
+    if (paneRef && paneRef instanceof Node) {
+      const calculatedHeight = paneRef.offsetHeight
```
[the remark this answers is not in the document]
colinlohner: it is intended, but maybe i should change the variable name so there isnt confusion

src/table/src/TableVirtualBody.js
mshwery: I think you need to call set state with a new object for it to rerender

PR #878, opened 2020-07-13
Update Table Row to FC/Hooks

colinlohner:
## Overview 
Update TableRow to use hooks and be a function component
 
N/A

## Testing
- [x] Existing functionality remains the same

### DIFF
--- a/src/table/src/TableRow.js
+++ b/src/table/src/TableRow.js
@@ -1,163 +1,149 @@
-import React, { PureComponent } from 'react'
+import React, { memo, forwardRef, useState } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 import { Pane } from '../../layers'
-import { withTheme } from '../../theme'
+import { useTheme } from '../../theme'
 import safeInvoke from '../../lib/safe-invoke'
 import { TableRowProvider } from './TableRowContext'
 import manageTableRowFocusInteraction from './manageTableRowFocusInteraction'
 
-class TableRow extends PureComponent {
-  static propTypes = {
-    /**
-     * Composes the Pane component as the base.
-     */
-    ...Pane.propTypes,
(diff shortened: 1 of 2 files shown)


src/table/src/TableRow.js
```diff
@@ -1,164 +1,149 @@
...
+        onSelect()
         e.preventDefault()
       } else if (key === 'ArrowUp' || key === 'ArrowDown') {
         try {
-          manageTableRowFocusInteraction(key, this.mainRef)
+          manageTableRowFocusInteraction(key, mainRef)
         } catch (_) {}
       } else if (key === 'Escape') {
-        if (this.mainRef) this.mainRef.blur()
+        if (mainRef && mainRef instanceof Node) mainRef.blur()
```
mshwery, line 60: Why do we need `instanceof Node` here?

colinlohner: just to make sure that whatever is passed back as the ref is an element, without it, it was failing to `blur`

src/table/src/TableRow.js
```diff
@@ -1,164 +1,149 @@
...
-    this.props.onKeyPress(e)
+    onKeyPress(e)
   }
 
-  onRef = ref => {
-    this.mainRef = ref
-    safeInvoke(this.props.innerRef, ref)
+  const onRef = newRef => {
+    setMainRef(newRef)
+    safeInvoke(ref, newRef)
```
mshwery, line 69: Does this work when the ref is from React.createRef or useRef?

mshwery: tested https://codesandbox.io/s/reverent-frog-dlgzn?file=/src/App.js and it does not. We'll need to manually detect which kind of ref it is.

colinlohner: Gonna do this in a separate PR to update across other components

PR #882, opened 2020-07-13
Update Table to be FC/use hooks

colinlohner:
## Overview 
Update Table to be FC/use hooks 

N/A

## Testing
- [x] Existing functionality remains as is

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1818,7 +1818,7 @@ interface TableVirtualBodyProps extends PaneProps {
 export interface TableProps extends PaneProps {
 }
 
-export class Table extends React.PureComponent<TableProps> {
+export class Table extends React.FC<TableProps> {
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1819,7 +1819,7 @@ interface TableVirtualBodyProps extends PaneProps {
 export interface TableProps extends PaneProps {
 }
 
-export class Table extends React.PureComponent<TableProps> {
+export class Table extends React.FC<TableProps> {
```
akleiner2, line 1822:
```
declare const Table: React.FC<TableProps> & {
     static members here
}
```

colinlohner: I can make this change once all the other type changes are merged

colinlohner:
YAY REMOVED SO MANY TS-IGNORES

PR #885, opened 2020-07-15
cleanup remnants of innerRef and fix bubbling

mshwery:
## Overview 
This PR cleans up a few last remnants of innerRef and typings
 
NA

## Testing

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -162,7 +162,7 @@ const Autocomplete = memo(
           getRootProps,
           ...restDownshiftProps
         }) => (
-          <Pane width="100%" {...getRootProps({ refKey: 'innerRef' })}>
+          <Pane width="100%" {...getRootProps()}>
             <Popover
               bringFocusInside={false}
               isShown={isShown}
--- a/index.d.ts
+++ b/index.d.ts
@@ -566,7 +566,7 @@ export type CardProps = React.ComponentProps<typeof Pane>
 
 export declare const Card: ForwardRefComponent<CardProps>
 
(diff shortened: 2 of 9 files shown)


src/autocomplete/src/Autocomplete.js
```diff
@@ -162,7 +162,7 @@ const Autocomplete = memo(
           getRootProps,
           ...restDownshiftProps
         }) => (
-          <Pane width="100%" {...getRootProps({ refKey: 'innerRef' })}>
```
mshwery, line 165: default is `ref` so we good here.

cstrat:
Not the right place to ask but for the first time I've had to use refs (focus mgmt on a field) and I've had to use `innerRef` which isn't really documented anywhere that I could find... and from this update, it sounds like it is going away in favor of just using `ref`?

My example is using a `<TextInputField>` element.
Using `innerRef` **works** but I am getting proptype errors from `<Text>` and `<Box>` for `innerRef`.
Using `ref` doesn't work at all...

I'm running v5 already ("evergreen-ui": "^5.0.0-2").

mshwery:
@cstrat can you see which version(s) of ui-box you have installed? `yarn why ui-box`

cstrat:
My issue was I was running 5.0.0-2, updated to -5 and its working fine with no warnings and errors using `ref` 👍

cstrat:
By the way, updating to -5 has stuffed up `<IconButton>` padding.

image

I know it is a WIP branch, but is that something you want bug reports on?

cstrat:
https://github.com/segmentio/evergreen/issues/900

However I think I am just running an old branch! Bug doesn't appear in the later releases.

PR #887, opened 2020-07-15
Update Tooltip Component to use hooks and to be a FC

colinlohner:
## Overview 
Updated Tooltip Component to use hooks and to be a function component 

N/A

## Testing
- [x] Verify existing functionality remains as is

### DIFF
--- a/src/tooltip/src/Tooltip.js
+++ b/src/tooltip/src/Tooltip.js
@@ -21,22 +21,19 @@ const Tooltip = memo(props => {
     statelessProps = {}
   } = props
 
-  const [id, setId] = useState(`evergreen-tooltip-${++idCounter}`)
-  const [willShow, setWillShow] = useState(false)
+  const [id] = useState(`evergreen-tooltip-${++idCounter}`)
   const [isShown, setIsShown] = useState(propIsShown || false)
   const [isShownByTarget, setIsShownByTarget] = useState(false)
   const [closeTimeout, setCloseTimeout] = useState(null)
 
   const mouseLeftTarget = () => {
     setIsShownByTarget(false)
-    setWillShow(false)
   }
 
   const handleMouseLeaveTarget = debounce(mouseLeftTarget, hideDelay)
 
   const hide = () => {
     setIsShown(false)
-    setWillShow(false)
   }
 
   const handleHide = debounce(hide, hideDelay)
@@ -58,7 +55,6 @@ const Tooltip = memo(props => {
       return
     }
 
-    setWillShow(true)
     setCloseTimeout(
       setTimeout(() => {
         setIsShown(true)
@@ -69,7 +65,7 @@ const Tooltip = memo(props => {
   const renderTarget = ({ getRef }) => {
     const tooltipTargetProps = {
       onMouseEnter: show,
-      onMouseLeave: hide,
+      onMouseLeave: handleHide,
       'aria-describedby': id
     }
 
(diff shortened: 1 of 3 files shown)


src/tooltip/src/Tooltip.js
```diff
@@ -9,114 +9,64 @@ import TooltipStateless from './TooltipStateless'
...
-    showDelay: PropTypes.number.isRequired,
-
-    /**
-     * When True, manually show the Tooltip.
-     */
-    isShown: PropTypes.bool,
+const Tooltip = memo(props => {
+  const {
+    appearance = 'default',
+    position = Position.BOTTOM,
```
mshwery, line 15: Any of these values that have defaults need to remove the `isRequired` from `propTypes` right?

src/tooltip/src/Tooltip.js
```diff
@@ -9,114 +9,64 @@ import TooltipStateless from './TooltipStateless'
...
-    hideDelay: 120,
-    showDelay: 0
-  }
+  const handleHide = debounce(hide, hideDelay)
+
+  // Component will unmount
+  useEffect(
+    () => () => {
+      clearTimeout(closeTimeout)
+      setCloseTimeout(null)
```
mshwery, line 45: No need to reset `setCloseTimeout`

src/tooltip/src/Tooltip.js
```diff
@@ -161,69 +111,100 @@ export default class Tooltip extends PureComponent {
...
-            )}
-          >
-            {content}
-          </TooltipStateless>
-        )}
-      </Positioner>
-    )
+  // Tooltip was explicitly set to not be shown
+  if (propIsShown === false) {
+    shown = false
```
mshwery, line 125: Couldn't this be combined with the above to make things a bit clearer?

```suggestion
  let shown = typeof propIsShown === 'boolean'
    ? propIsShown
    : (isShown || isShownByTarget) && !isPopoverShown()
```

src/tooltip/src/Tooltip.js
```diff
@@ -161,69 +111,100 @@ export default class Tooltip extends PureComponent {
...
+      }}
+      isShown={shown}
+      position={position}
+      animationDuration={160}
+    >
+      {({ css, style, state, getRef }) => (
+        <TooltipStateless
+          id={id}
+          appearance={appearance}
+          ref={ref => getRef(ref)}
```
mshwery, line 141: Can this just be `ref={getRef}`?

src/tooltip/src/Tooltip.js
colinlohner: Yes. My bad

src/tooltip/src/Tooltip.js
```diff
@@ -118,12 +117,9 @@ const Tooltip = memo(props => {
     setIsShownByTarget(true)
   }
 
-  let shown = (propIsShown || isShown || isShownByTarget) && !isPopoverShown()
-
-  // Tooltip was explicitly set to not be shown
-  if (propIsShown === false) {
-    shown = false
-  }
```
mshwery: Actually now that I'm looking at this again I think this is meant to only capture when `isShown` is explicitly false. When it's set to `true` we do indeed need `&& !isPopoverShown()` as well.

Maybe revert my suggestion 😬

colinlohner: lol okay

PR #888, opened 2020-07-15
Update TagInput to be FC and use hooks

colinlohner:
## Overview 
Update TagInput Component to use hooks and to be a function component 

N/A

## Testing
- [x] Existing functionality remains as is

### DIFF
--- a/src/tag-input/src/TagInput.js
+++ b/src/tag-input/src/TagInput.js
@@ -55,7 +55,6 @@ const TagInput = memo(props => {
       : [inputValue]
 
   const addTags = (value = '') => {
-    console.log(value)
     const newValues = getValues(value)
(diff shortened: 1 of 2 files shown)


src/tag-input/src/TagInput.js
```diff
@@ -20,268 +20,238 @@ const GET_KEY_FOR_TAG_DELIMITER = {
...
+          .split(separator)
+          .map(v => v.trim())
+          .filter(v => v.length > 0)
+      : [inputValue]
 
-  addTags = (value = '') => {
-    const { onAdd, onChange, values } = this.props
-    const newValues = this.getValues(value)
+  const addTags = (value = '') => {
+    console.log(value)
```
mshwery, line 58:
```suggestion
```

PR #889, opened 2020-07-16
build some custom hooks

mshwery:
## Overview 
This swaps out some redundant logic with semantic custom hooks.

NA

## Testing

### DIFF
--- a/src/hooks/index.js
+++ b/src/hooks/index.js
@@ -1,0 +1,2 @@
+export * from './use-force-update'
+export * from './use-id'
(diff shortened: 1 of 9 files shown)


src/hooks/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './use-force-update'
```
akleiner2, line 1: 👍 -- was going to ask if you thought that these would be worth exposing externally, but my hunch is probably not ;)

PR #891, opened 2020-07-17
move storybook stuff to devdeps

mshwery:
## Overview 
This moves a devDep to the devDependencies section
 
## Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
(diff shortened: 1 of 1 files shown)


akleiner2:
Cool, looks good -- curious if you have to re-gen `yarn.lock` (if it marks deps / devdeps differently), but otherwise 👍 from me

mshwery:
I actually removed it and re-added with `-D` and no lockfile changes happened

PR #896, opened 2020-07-20
V5 box types

mshwery:
## Overview 

This improves our typings to support polymorphic box pass-through props. For instance, this now works:

```
<Link is={ReactRouterLink} to="/home" />
```

Both the forwarded `ref` and all ReactRouterLink props will be type-safe.
 
NA 

## Testing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,7 @@
 /* tslint:disable:interface-name max-classes-per-file no-empty-interface */
 
 import * as React from 'react'
-import Box, { extractStyles as boxExtractStyles } from 'ui-box'
-import { BoxProps } from 'ui-box/dist/types/box-types'
+import Box, { extractStyles as boxExtractStyles, BoxProps } from 'ui-box'
 import { StyleAttribute, CSSProperties } from 'glamor'
 import { DownshiftProps } from 'downshift'
 import { TransitionProps, TransitionStatus } from 'react-transition-group/Transition'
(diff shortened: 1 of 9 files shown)


src/select-menu/src/SelectMenu.js
```diff
@@ -36,65 +36,62 @@ const getEmptyView = (close, emptyView) => {
   return {}
 }
 
-const SelectMenu = memo(
-  forwardRef((props, ref) => {
```
mshwery, line 40: Not clear what the ref should be attached to, and Popover doesnt forward refs anyway.

src/side-sheet/src/SideSheet.js
```diff
@@ -125,64 +125,61 @@ const animationStylesClass = {
   }
 }
 
-const SideSheet = memo(
-  forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
akleiner2: 👍

index.d.ts
```diff
@@ -1,13 +1,19 @@
...
 import { DownshiftProps } from 'downshift'
 import { TransitionProps, TransitionStatus } from 'react-transition-group/Transition'
 
-export { configureSafeHref } from 'ui-box'
+export { configureSafeHref, BoxProps } from 'ui-box'
+
+/**
+ * Convenience method for defining your own components that extend Box and pass-through props
+ */
+export type BoxComponent<P = {}, D extends React.ElementType = React.ElementType> = <
```
akleiner2, line 14: Clever! Looks legit -- take component-specific props, take an `is` prop, and do some exclusion on those `BoxProps` and the component-specific props.

mshwery: The big win here is that this type respects consumers passing in other `is` values and it will give them all the props along with that!

PR #898, opened 2020-07-21
Add forwardRef to serach table header cell

colinlohner:
## Overview 
Add forwardRef to searchTableHeaderCell 

N/A

## Testing
- [x] ref forwards as expected

### DIFF
--- a/src/table/src/SearchTableHeaderCell.js
+++ b/src/table/src/SearchTableHeaderCell.js
@@ -21,7 +21,7 @@ const invisibleInputClass = css({
   }
 }).toString()
 
-const SearchTableHeaderCell = memo(forwardRef((props, forwardedRef) => {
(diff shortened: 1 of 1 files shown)


src/table/src/SearchTableHeaderCell.js
```diff
@@ -21,7 +21,7 @@ const invisibleInputClass = css({
   }
 }).toString()
 
-const SearchTableHeaderCell = memo(props => {
+const SearchTableHeaderCell = memo(forwardRef((props, forwardedRef) => {
```
akleiner2, line 24: Tiniest of non-blocking nits, I think a lot of the codebase just passes in `ref` as this argument and then we can just do `ref={ref}` later on. But obviously non-blocking by any means

PR #905, opened 2020-07-23
update v5 migration docs

mshwery:
## Overview 
This updates the migration guide, improves `diff` syntax highlighting, and allows for markdown [extended syntax for custom header ids](https://www.markdownguide.org/extended-syntax/#:~:text=Heading%20IDs,same%20line%20as%20the%20heading)
 
image

## Testing

### DIFF
--- a/docs/src/components/DocsMDXProvider.js
+++ b/docs/src/components/DocsMDXProvider.js
@@ -8,23 +8,40 @@ import MDXPlayground from './MDXPlayground'
  */
 const Components = {}
 
-const generateHeadingId = (props) => {
-  if (props && typeof props.children === 'string') {
-    return `heading_${props.children.toLowerCase().replace(' ', '_')}`
+// eslint-disable-next-line prefer-named-capture-group
+const regex = /\{#([a-zA-Z][\w\-_]*)\}/g
+
+const extractId = value => {
+  if (!value || typeof value !== 'string') {
+    return value
   }
 
-  return false
+  let [heading, id] = value.split(regex)
+  if (!id) {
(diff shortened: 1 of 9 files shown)


docs/src/components/DocsMDXProvider.js
```diff
@@ -8,23 +8,40 @@ import MDXPlayground from './MDXPlayground'
  */
 const Components = {}
 
-const generateHeadingId = (props) => {
-  if (props && typeof props.children === 'string') {
-    return `heading_${props.children.toLowerCase().replace(' ', '_')}`
```
mshwery, line 13: This was only replacing the first instance of a space character.

docs/src/pages/whats-new/migration-docs/v5.mdx
```diff
@@ -3,110 +3,124 @@ import MigrationDocsLayout from '../../../components/MigrationDocsLayout'
...
+   const inputRef = React.useRef()
+-  return <TextInput innerRef={inputRef} />
++  return <TextInput ref={inputRef} />
+ }
+```
+
+### Popover triggers {#popovers}
+With the removal of `innerRef` in Evergreen, there are some potential changes required with implementations of `Popover`.
+The direct `child` of a `Popover` **must be** a component that can forward refs to a DOM node. If you are using class components,
+you'll need to update our code. We've found the easiest path is to use an Evergreen component – which already handle ref forwarding.
```
colinlohner, line 40: * you'll need to update **your** code

docs/src/components/DocsMDXProvider.js
colinlohner: nice catch

PR #916, opened 2020-07-29
render optimizations

mshwery:
## Overview 
We noticed some perf/latency with the `SelectMenu` so I started adding a bunch of optimizations to avoid re-renders
 
NA

## Testing


akleiner2:
Seems like a sane first pass, but there are probably further optimizations we can make to the component tree, remove indirection etc

PR #920, opened 2020-08-03
Fix toaster types

AntsiferovMaxim:
This resolves the type error when passing react component to toast

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2312,19 +2312,19 @@ export const toaster: {
   /**
    * Opens a Toast with an intent of none.
    */
-  notify: (title: string, settings?: ToasterSettings) => void
+  notify: (title: string | React.ReactNode, settings?: ToasterSettings) => void
   /**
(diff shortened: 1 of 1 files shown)


akleiner2:
@AntsiferovMaxim Thanks for this PR! Left a comment inline - let me know if you have any questions!

index.d.ts
```diff
@@ -2312,19 +2312,19 @@ export const toaster: {
   /**
    * Opens a Toast with an intent of none.
    */
-  notify: (title: string, settings?: ToasterSettings) => void
+  notify: (title: string | React.ReactNode, settings?: ToasterSettings) => void
```
akleiner2, line 2315: We can simplify these title types to just be `React.ReactNode` - `React.ReactNode` is a catch-all type that encompasses `string` as a part of it.

AntsiferovHive: Good idea👍

PR #926, opened 2020-08-07
Call `onStateChange` if passed to `<Combobox>`

vadimdemedes:
## Overview

Because `onStateChange` is being overriden inside `<Combobox>`, it's not possible to customize Downshift's behavior. This PR checks if `onStateChange` function is passed inside `autocompleteProps` and calls it if it is.

Use case for this in particular is typeahead - autocomplete, but without forcing user to select one of the `items`. User can type any value and autocomplete is optional. See "Any value" example in `src/combobox/stories/index.stories.js`.

## Testing

- [ ] Updated Typescript types and/or component PropTypes (no need)
- [ ] Added / modified component docs (no need)

### DIFF
--- a/src/combobox/src/Combobox.js
+++ b/src/combobox/src/Combobox.js
@@ -28,11 +28,18 @@ const Combobox = memo(function Combobox(props) {
 
   const [isOpenedByButton, setIsOpenedByButton] = useState(false)
 
-  const handleStateChange = changes => {
+  const handleStateChange = (changes, stateAndHelpers) => {
     if (Object.prototype.hasOwnProperty.call(changes, 'isOpen')) {
       if (!changes.isOpen) {
         setIsOpenedByButton(false)
       }
+    }
+
+    if (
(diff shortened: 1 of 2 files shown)


src/combobox/src/Combobox.js
```diff
@@ -28,12 +28,19 @@ const Combobox = memo(function Combobox(props) {
...
       if (!changes.isOpen) {
         setIsOpenedByButton(false)
       }
     }
+
+    if (
+      autocompleteProps &&
+      typeof autocompleteProps.onStateChange === 'function'
+    ) {
+      autocompleteProps.onStateChange(changes, stateAndHelpers)
```
akleiner2, line 42: Just out of curiosity, where is this second argument being consumed from? Is this from Downshift itself?

vadimdemedes: Yep.

colinlohner:
Will be cutting a new release today with this in it

PR #932, opened 2020-08-10
Fix(switch): set default value (checked)

wangcch:
## Overview
https://codesandbox.io/s/switch-default-value-nwg2g?file=/src/App.js

### now:

ezgif-6-120897d41368

### after:

checked: `undefined` -> `false` to control the value of `input:checkbox`.

ezgif-6-85f942c6e26d

## Testing


colinlohner:
Will be cutting a new release today with this in it

Issue #929, opened 2020-08-09
React strict mode complains about validationMessage on TextInputField 


Procrat:
#### Summary
When passing the `validationMessage` prop to a `TextInputField`, it gets passed down to the underlying `input` element.

#### Steps to reproduce
```
import React from 'react';
import { TextInputField } from 'evergreen-ui';

export default () => <TextInputField label="" validationMessage="some message" />;
```

#### Observed behaviour
The rendered `input` element has a `validationmessage` attribute.

With React's strict mode turned on, we also get the following warning in the console:
```
Warning: React does not recognize the `validationMessage` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `validationmessage` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in input (created by Box)
    in Box (created by ForwardRef(Text))
    in ForwardRef(Text)
    in ForwardRef(Text) (created by ForwardRef(TextInput))
    in ForwardRef(TextInput)
    in ForwardRef(TextInput) (created by ForwardRef(TextInputField))
    in div (created by Box)
    in Box (created by ForwardRef(FormField))
    in ForwardRef(FormField)
    in ForwardRef(FormField) (created by ForwardRef(TextInputField))
    in ForwardRef(TextInputField)
    in ForwardRef(TextInputField) (at <my own code>)
```

#### Expected behaviour
No `validationmessage` on the `input` element and (thus) no warning.

PS: Thank you for the awesome work that has gone into creating and maintaining this beautiful library! 👏

colinlohner:
Thanks for calling this out! Will get to resolving this asap

colinlohner:
Okay will be cutting a new release of Evergreen today with this included in it

colinlohner:
Released!

Available in v5.0.4

Procrat:
Wow, thanks a lot, Colin! I cannot express the gratitude I feel for the speed at which this and my other issue got fixed, merged and released (#930)! 🙏

ValiDraganescu:
I still get this bug in 5.1.2

PR #933, opened 2020-08-10
Resolve issue, where incorrect props where being passed around

colinlohner:
## Overview
Resolves issue: #929 

We were splitting the entire props object and passing the results down. This was resulting in props being passed to places we didn't want them to be.

n/a 

## Testing
- [x] Verified that existing functionality remains uneffected
- [x] #929 is resolved

### DIFF
--- a/src/text-input/src/TextInputField.js
+++ b/src/text-input/src/TextInputField.js
@@ -36,7 +36,7 @@ const TextInputField = memo(
     /**
      * Split the wrapper props from the input props.
      */
-    const { matchedProps, remainingProps } = splitBoxProps(props)
+    const { matchedProps, remainingProps } = splitBoxProps(restProps)
 
     return (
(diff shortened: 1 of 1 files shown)


src/text-input/src/TextInputField.js
```diff
@@ -48,7 +48,6 @@ const TextInputField = memo(
         validationMessage={validationMessage}
         labelFor={id}
         {...matchedProps}
-        {...restProps}
```
akleiner2, line 51: Still worth spreading `remainingProps` here?

akleiner2: Oh just kidding! See that they're sent below. Nevermind!

colinlohner: no that gets spread bellow on the input element

Issue #925, opened 2020-08-07
v5 - Menu Icons & Button Icon padding


cstrat:
Hi,

Hopefully I didn't just overlook a step in the migration guide (padding doesn't seem to be mentioned), but I was using the pre-release without issue prior to upgrading to 5.0.3.
Padding seems to have dropped off.

![image](https://user-images.githubusercontent.com/4209076/89609560-947e2380-d8bb-11ea-9395-8a70be5478b4.png)

Repo:
https://codesandbox.io/s/v5-padding-bug-dohc8?file=/src/App.js

cstrat:
I think I noticed why this is happening.
Looks like the elements are passing in `marginLeft` to the `Icon` - but I am using `FontAwesome` icon's and because of that the marginLeft doesn't mean anything and is the cause for the errors in the console.

colinlohner:
Hey @cstrat, yeah it does look like an incompatibility issue with FontAwesome. I took a second and worked up a quick fix you could implement to solve it. This isn't an Evergreen issue specifically as you are using a thirdparty icon library.

Hope this helps, happy to answer any questions or talk further.

https://codesandbox.io/s/v5-padding-bug-forked-cr2yv?file=/src/App.js

colinlohner:
One other issue I noticed is that FontAwesome has its own size prop that doesn't map to what we use. In the codesandbox I strip that prop out, however you could map our sizes to the FontAwesome ones if you wanted.

mshwery:
@colinlohner @cstrat we may want to consider passing all of these props to an evergreen/ui-box-controlled component so that the spacing concerns are kept intact regardless of the component being passed in.

colinlohner:
@mshwery you raise a good point. Still wouldn't solve color and size, but would at least preserve spacing. I'll get a PR together for this

colinlohner:
Okay this will be shipped as part of a release we are going to cut today

colinlohner:
You may still have to do something similar to the codesandbox in order to handle colors or size. FontAwesome does respect font size so you may be okay without dealing with size

colinlohner:
Released you can now use as you were and the positioning will be handled.

Available in v5.0.4

PR #934, opened 2020-08-10
Add box wrapper to maintain positioning for custom icons

colinlohner:
## Overview
Resolves issue #925 

Added a `ui-box` wrapper to the `IconWrapper` component to help maintain positioning when using custom icons

**Using FontAwesome**
image
image

## Testing
- [x] Third-party icons now get positioned correctly
- [x] Internal icons still get positioned correctly

### DIFF
--- a/src/icons/src/IconWrapper.js
+++ b/src/icons/src/IconWrapper.js
@@ -67,5 +67,5 @@ IconWrapper.propTypes = {
    * readers will use it for aural feedback.
    * By default, this is set to the icon's name for accessibility.
    */
-  title: PropTypes.string,
+  title: PropTypes.string
 }
--- a/src/icons/src/IconWrapper.js
+++ b/src/icons/src/IconWrapper.js
@@ -1,13 +1,14 @@
 import React, { forwardRef, memo } from 'react'
 import PropTypes from 'prop-types'
 import ReactIs from 'react-is'
+import Box from 'ui-box'
 
 /**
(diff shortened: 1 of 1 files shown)


src/icons/src/IconWrapper.js
```diff
@@ -1,13 +1,14 @@
...
 import ReactIs from 'react-is'
+import Box from 'ui-box'
 
 /**
  * This is an internal helper component for rendering custom or Evergreen icons
  * Box props are applied to the outer Box container, and Evergreen icon-specific props are added to the icon element.
  */
 export const IconWrapper = memo(
-  forwardRef(function Icon({ icon, ...props }, ref) {
+  forwardRef(function Icon({ icon, marginLeft, marginRight, marginTop, marginBottom, ...props }, ref) {
```
mshwery, line 11: It might be better to pull out all `icon`-specific props here and pass those down instead of `...props` (e.g. for `React.cloneElement`, etc).

Then we can pass `...props` directly to the outer Box.

src/icons/src/IconWrapper.js
```diff
@@ -20,7 +21,16 @@ export const IconWrapper = memo(
       iconWithProps = React.cloneElement(icon, { ...props, ...icon.props, ref })
     }
 
-    return iconWithProps
+    return (
+      <Box
+        display="flex"
```
mshwery, line 26: I don't think we can use `flex` in every scenario. I think `inline-flex` is probably closer to the current behavior?

PR #935, opened 2020-08-11
Add missing disabled property to SegmentedControlOwnProps interface

Ozonid:
First of all congrats on releasing v5! The Icon changes are especially welcome.

After upgrading my project to v5 I'm no longer able to build the code as TS complains about an invalid `disabled` prop I'm providing to `SegmentedControl`. As far as I could see the feature is still there, so I've added the missing disabled prop to the appropriate interface.


colinlohner:
Will get this merged in to a release today

PR #936, opened 2020-08-12
Update layout-primitives.mdx

designdevy:
## Overview

In the "Use the height property to resize components" section, I assume the margin for the button is supposed to be set for the left side not top and adding a display flex property will make the example look better.

Before:
image

After:
image

## Testing


designdevy:
@mshwery when does the change roll out to the docs? Next release?

PR #942, opened 2020-08-18
V6 status indicator

colinlohner:
## Overview
Created the new Status Indicator component

image
image
image

## Testing


mshwery:
Did you mean to target `master`?

colinlohner:
@mshwery nope! thanks

index.d.ts
```diff
@@ -1626,6 +1626,15 @@ export declare const Stack: React.FC<StackProps>
 
 export declare const StackingContext: React.Context<number>
 
+export interface StatusIndicatorOwnProps extends TextOwnProps {
+  children?: React.ReactNode
```
mshwery, line 1630: iirc you shouldn't ever have to specify `children`

yarn.lock
```diff
@@ -1791,6 +1791,25 @@
   dependencies:
     babel-runtime "^6.26.0"
 
+"@storybook/addon-storysource@^5.3.19":
```
mshwery, line 1794: Wonder if we can add something that reveals the props and let's you modify them too? I recall something like this for storybook.

src/status-indicator/src/StatusIndicator.js
```diff
@@ -0,0 +1,45 @@
...
+const StatusIndicator = memo(
+  forwardRef(function StatusIndicator(props, ref) {
+    const { children, disabled, color = 'info', ...rest } = props
+    const dotColor = useIconColor(color === 'none' ? 'info' : color)
+
+    return (
+      <Text display="inline-flex" alignItems="center" ref={ref} {...rest}>
+        <Box
+          flexShrink={0}
+          marginRight={8}
```
mshwery, line 16: We should source this from the theme spacing (or at least from the scale)?

src/status-indicator/src/StatusIndicator.js
```diff
@@ -0,0 +1,45 @@
...
+          borderRadius="50%"
+          height={7}
+          width={7}
```
mshwery, line 19: Should this be controllable via `dotSize` or similar?

yarn.lock
colinlohner: Yeah, gonna add knobs in somepoint soon. but its a little more involved

PR #945, opened 2020-08-21
fix unnecessary re-renders for popover component

tusharf5:
## Overview

Hi there, I am submitting this PR to fix an issue that I recently noticed with the Popover component. When you pass it a function for the `content` prop. It continuously re-renders. I am also attaching a video of the issue. This doesn't happen when you don't pass a function.

```jsx
// this cause continuous re-render
<Popover
    content={({ close }) => <ClosablePopoverContent close={close} />}
>...</Popover>
```

```jsx
// this works fine
<Popover
    content={<PopoverContent />} 
>...</Popover>
```

**Before Changes**
https://www.loom.com/share/4735e08de770484db240870c251f8dc5

**After Changes**
https://www.loom.com/share/bc157742c36b434ea77d14135a325692

## Fix

I've added various optimization hooks to the Popover component which resolves this issue. I was also thinking of encouraging users in the documentation to not use an inline-function but a memoized function or in case of a class component a class method but it is not really required after the fix.

## Testing

I've done local testing with all the storybook variants of the component and it is working as expected. Let me know if I need to make any more changes from my side. Happy to work on this :)

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -4,7 +4,9 @@ import React, {
   useRef,
   useState,
   useEffect,
-  useImperativeHandle
+  useImperativeHandle,
+  useCallback,
+  useMemo
 } from 'react'
 import cx from 'classnames'
 import { css as glamorCss } from 'glamor'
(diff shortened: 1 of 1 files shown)


tusharf5:
hey @colinlohner Did you get a chance to review the PR?

colinlohner:
@tusharf5 yes, I'll get this merged in today and release

PR #947, opened 2020-08-25
Break apart theme pieces to support v5 and v6 theming

akleiner2:
## Overview
A lot of stuff still in flight for this PR, but this PR basically copies over the `default-theme` implementation to separate folders, generalizes some of the `useXAppearance` hooks to accept arbitrary styles, and starts building upon those. 

This should set us up to support arbitrary theme objects very soon.

## Testing


src/segmented-control/src/SegmentedControlRadio.js
```diff
@@ -53,7 +49,7 @@ const SegmentedControlRadio = memo(
           cursor="pointer"
           htmlFor={id}
           fontWeight={500}
-          size={textSize}
```
mshwery, line 56: Breaking change? Should we start cataloging everything that is breaking in a .md file?

src/theme/src/themes/v5/component-specific/buttons.js
```diff
@@ -0,0 +1,67 @@
+import { defaultControlStyles } from '../shared'
+import scales from '../foundational-styles/scales'
```
mshwery, line 2: Should this come from the theme?

src/theme/src/themes/v5/component-specific/getAvatarProps.js
```diff
@@ -0,0 +1,21 @@
+import fills from '../foundational-styles/fills'
```
mshwery, line 1: From the theme?

PR #953, opened 2020-08-27
fix classnames overwrites in Tab, TagInput, FilePicker, TextDropdown

### LINKED ISSUE #946: Tab's className has been replaced
When I try to
```
<Tab className="hello">
```
I will lose the theme classes defined by the lib, do you think it makes sense to add the capability to allow the user to extend their own classes?

https://github.com/segmentio/evergreen/blob/master/src/tabs/src/Tab.js#L90

tusharf5:
## Overview

fixes #946 

Adds `className` prop to some components using the `classnames` library .

The only place where it is missing now is `MenuItem.js`. I did not add it there as another PR is modifying that file which would result in merge conflicts.

Not applicable.

## Testing

Tested locally. Does not break any existing component styles.

### DIFF
--- a/src/file-picker/src/FilePicker.js
+++ b/src/file-picker/src/FilePicker.js
@@ -1,6 +1,7 @@
 import React, { memo, forwardRef, useState, useRef, useCallback } from 'react'
 import PropTypes from 'prop-types'
 import Box from 'ui-box'
+import cx from 'classnames'
 import { Button } from '../../buttons'
 import { TextInput } from '../../text-input'
 import safeInvoke from '../../lib/safe-invoke'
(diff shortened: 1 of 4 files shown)


src/file-picker/src/FilePicker.js
```diff
@@ -182,7 +186,12 @@ FilePicker.propTypes = {
   /**
    * Placeholder of the text input
    */
-  placeholder: PropTypes.string
+  placeholder: PropTypes.string,
+
+  /**
+   * Class name passed to the FilePicker.
```
colinlohner, line 192: Can you add "Only use this if you know what you are doing." to these comments

colinlohner: @tusharf5 sorry just noticed, that we want to add this

tusharf5: Sure, I'll add it just now.

Issue #796, opened 2020-04-22
Unable to disable MenuItem


dustincjensen:
## Issue

In the [documentation](https://evergreen.segment.com/components/menu) for `Menu` it mentions the following.

`Arrow keys within the menu will cycle through all of the menu items and skip disabled items.`

However, by default you cannot disable a `MenuItem`.
If you set the `is` property to make the `MenuItem` a `Button` then you can, but then you have to mess with padding and widths.

## Suggestion

- Update the documentation to have an example of the suggested approach to disable a menu item.

colinlohner:
This has been fixed in v5.0.6

PR #954, opened 2020-08-27
fix menu disable items functionality and appearence

tusharf5:
## Overview

Hi there, this PR addresses this (#796 ) issue by fixing the menu disabled functionality. This PR makes the following changes.

- Gives a disabled appearance to a menu item when `disabled` prop is true.
- Fixes memory leaks in the Menu component (`keydown` listeners were note being deregistered on component unmount)
- Fixes an issue when Popover won't bring the focus to the MenuItem if the first menu item is disabled.
- Modifies the query selector in Popover to only select children that are not disabled thus no need to skip any elements when navigating through keys.

https://www.loom.com/share/9055c6fdfb3c4011ae8579b7f19fa784

## Note

Since this PR also modifies the Popover component and another PR #945 makes some major changes to that component so this might result in a merge conflict. In that case, I would suggest merging that PR first.

## Testing

### DIFF
--- a/src/menu/src/Menu.js
+++ b/src/menu/src/Menu.js
@@ -8,89 +8,89 @@ import MenuOption from './MenuOption'
 import MenuOptionsGroup from './MenuOptionsGroup'
 
 const Menu = memo(function Menu(props) {
-  const menuRef = useRef()
+
+  const menuRef = useRef(null)
   const firstItem = useRef()
   const lastItem = useRef()
 
   const menuItems = useRef()
 
   useEffect(() => {
-    if (menuRef && menuRef.current) {
-      menuItems.current = [
-        ...menuRef.current.querySelectorAll(
-          '[role="menuitemradio"], [role="menuitem"]'
-        )
-      ]
 
-      if (menuItems.current.length === 0) {
-        throw new Error('The menu has no menu items')
-      }
+    menuItems.current = menuRef.current ? [
+      ...menuRef.current.querySelectorAll(
+        '[role="menuitemradio"], [role="menuitem"]'
+     )].filter(el => el.getAttribute('disabled') === null)
+    : []
 
-      firstItem.current = menuItems.current[0]
-      lastItem.current = menuItems.current[menuItems.current.length - 1]
+    if (menuItems.current.length === 0) {
+      throw new Error('The menu has no menu items')
+    }
 
-      const focusNext = (currentItem, startItem) => {
-        // Determine which item is the startItem (first or last)
-        const goingDown = startItem === firstItem.current
+    firstItem.current = menuItems.current[0]
+    lastItem.current = menuItems.current[menuItems.current.length - 1]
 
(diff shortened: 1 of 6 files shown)


tusharf5:
There is an issue that when the first menu item is disabled, popover doesn't focus on `keydown` event. I'll fix that and then re-submit the PR.

tusharf5:
fixed now.

tusharf5:
Hi @colinlohner @mshwery could you review this PR whenever you get some free time? I'm happy to make any changes if required.

src/menu/src/MenuItem.js
```diff
@@ -122,7 +147,12 @@ MenuItem.propTypes = {
   /**
    * Callback to invoke onkeypress
    */
-  onKeyPress: PropTypes.func
+  onKeyPress: PropTypes.func,
+
+  /**
+   * Flag to indicate whether the menu item is disabled or not
+   */
+  disabled: PropTypes.bool
```
mshwery, line 155: Can you update the typedefs as well in `index.d.ts`?

src/menu/src/MenuItem.js
```diff
@@ -46,6 +47,29 @@ const MenuItem = memo(
...
+    const disabledProps = useMemo(() => {
+      return disabled ? {
+        backgroundColor: theme.colors.background.tint1,
+        cursor: 'not-allowed',
+        disabled: true,
+        onClick: null,
+        onKeyPress: null,
+        tabIndex: -1,
+        'aria-disabled': 'true',
+        'data-isselectable': 'false'
```
mshwery, line 69: 🙌

src/menu/src/Menu.js
```diff
@@ -8,89 +8,89 @@ import MenuOption from './MenuOption'
...
+      if (e.key === 'End') {
+        e.preventDefault()
+        lastItem.current.focus()
+      }
+    }
+
+    menuItems.current.forEach(menuItem => menuItem.addEventListener('keydown', onKeyPressListener))
+
+    return () => {
+      menuItems.current.forEach(menuItem => menuItem.removeEventListener('keydown', onKeyPressListener))
```
mshwery, line 93: Not in love with adding event listeners to every element just to capture keypresses. Seems like we could have a single event listener added and remove when the menu is opened/closed?

(realize this is out of scope for your changes, but just seeing this with fresh eyes)

tusharf5: I agree, we could use [event delegation](https://davidwalsh.name/event-delegate) here to only register a single event listener on the parent. will look into this.

src/menu/src/MenuItem.js
tusharf5: Done ✅

src/menu/src/Menu.js
mshwery: No worries about solving here in this PR. We can do this in a follow-up (if you are interested in taking a stab at it!)

PR #956, opened 2020-08-27
V6 rename themes

colinlohner:
## Overview
Rename the themes

n/a

## Testing


src/theme/stories/index.stories.js
```diff
@@ -4,8 +4,8 @@ import Box from 'ui-box'
 import Component from '@reactions/component'
 import { Heading } from '../../typography'
 import { Checkbox } from '../../checkbox'
-import v5Theme from '../src/themes/v5'
-import v6Theme from '../src/themes/v6'
+import v5Theme from '../src/themes/legacy'
```
mshwery, line 7: would suggest avoiding calling this "legacy" – might be better to call keep it as `v5`. wdyt?

colinlohner: In my mind it was a subtle hint that we wont support it for very long. But can go with v5 if you think that's better

mshwery: why wouldnt we support it? it's a valid theme according to the new theming capabilities, right?

colinlohner: Just thinking we don't want to have to maintain two themes long term. If we added new components etc in the future we would have to update both themes

PR #957, opened 2020-08-27
V6 theme re org

colinlohner:
## Overview
New organization of theming. Default styles are now created using a method so that they are theme aware (currently only applied to buttons). Passing custom styles and custom variants is still allowed/available and now token overwrites work as expected.

Passing custom variant:
```
// Theme object
const customTheme = {
  componentName: {
    variantName: {
      base: {
        backgroundColor: 'red'
      }
    }
  }
}

<ThemeProvider value={customTheme}>
```

Updating token
```
// Theme object
const customTheme = {
  tokens: {
    primary: {
      base: 'green'
    }
  }
}

<ThemeProvider value={customTheme}>
```

n/a 

## Testing


src/theme/src/default-styles/buttons.js
```diff
@@ -0,0 +1,97 @@
+import { getDefaultControlStyles } from './shared'
```
akleiner2, line 1: You're essentially lumping this as the new `default-theme`?

akleiner2:
Cool! I think this looks good 👍 . Can't wait till we're able to get rid of the other cruft here too

akleiner2:
Btw - for disabled buttons, can we add a `cursor: not-allowed`? In addition to just making the background `<color>100`

colinlohner:
@akleiner2 that's actually there, coming from `shared.js` disabled styles

colinlohner:
may not be visible though, due to the `pointer-events: none`

PR #958, opened 2020-08-28
Update inputs to new theming style

colinlohner:
## Overview
Update inputs to new theming style

n/a

## Testing


src/theme/src/default-styles/inputs.js
```diff
@@ -0,0 +1,46 @@
+export default function getDefaultStyles(theme) {
```
mshwery, line 1: What are "default" styles? Is this meant to be part of the v6 theme ?

src/theme/src/hooks/useInputAppearance.js
```diff
@@ -17,7 +18,9 @@ const hoverPlaceholderState = '&:hover::placeholder'
 const focusPlaceholderState = '&:focus::placeholder'
 
 const getInputStyles = (theme, appearance) => {
-  const { inputs } = theme
+  const { inputs: themeStyles } = theme
+  const defaultStyles = getDefaultStyles(theme)
+  const inputStyles = { ...defaultStyles, ...themeStyles }
```
mshwery, line 23: Hm... IMO we should either:
1. explicitly tell consumers that if they don't pass in a complete theme object we will merge it with the v6 theme
2. provide a `createTheme` method for users to pass partial theme objects

src/theme/src/hooks/useInputAppearance.js
```diff
@@ -17,7 +18,9 @@ const hoverPlaceholderState = '&:hover::placeholder'
 const focusPlaceholderState = '&:focus::placeholder'
 
 const getInputStyles = (theme, appearance) => {
-  const { inputs } = theme
+  const { inputs: themeStyles } = theme
+  const defaultStyles = getDefaultStyles(theme)
```
mshwery, line 22: Can we instead pull this into the themes that rely on it (e.g. v5 and v6)? All styles should come from the theme itself, not from imported modules.

colinlohner:
Closing this, since we are no longer going to do this approach

PR #959, opened 2020-08-28
fix undefined props merging and list styles

mshwery:
# Overview
This PR fixes an issue where our list items were getting props overridden with `undefined` values – we shouldn't pass in undefined to child components if we can avoid it. There was also a style issue because we are using `em` for the list style margin, but that's not reliable (different apps could have different relative units).

**Before (ignore the icon, look at the left alignment)**
Screen Shot 2020-08-28 at 10 44 38 AM

**After**
Screen Shot 2020-08-28 at 11 29 08 AM

## Testing

### DIFF
--- a/src/lib/remove-undefined.js
+++ b/src/lib/remove-undefined.js
@@ -1,0 +1,14 @@
+/**
+ * Cleans an object of undefined values
+ */
+export default function removeUndefined(input = {}) {
+  const obj = { ...input }
+
+  Object.keys(obj).forEach(key => {
+    if (obj[key] === undefined) {
+      delete obj[key]
+    }
+  })
+
+  return obj
+}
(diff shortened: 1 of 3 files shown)


src/lib/remove-undefined.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Cleans an object of undefined values
+ */
+export default function removeUndefined(input = {}) {
```
akleiner2, line 4: Do we want _just_ undefined? Or falsy values?

mshwery: in this case we want to strip undefined only. `null` `0` `false` are all valid _defined_ values.

In our case we were passing in `undefined`s that were overriding defined values, which isn't the desired behavior.

akleiner2: got it!

PR #962, opened 2020-09-02
fix Popover focus issues

tusharf5:
## Overview

I was trying out the latest release 5.0.6 which introduced some changes in the Popover and Menu component. I noticed that the focus behaviour or Menu inside a Popover is not working properly. I've attached the video. **Note** that I am only using the keyboard in the video to scroll (using ArrowDown key) and not the mouse scroll.

This PR aims to fix those issues that were introduced by those changes.

### Issue 1 (Popover not bringing focus to Menu)

Popover is not bringing focus to menu (it does sometimes). This is caused because the useEffect was missing a dependency which was causing the bringFocusInside function to quit early. Just adding the dependency fixed the issue.

### Issue 2 (Body scroll on keydown)

If the content of the body has an overflow scroll, the first time when pressing the key down (when popover is open) body also receives the key down event which triggers a scroll (can see in the video) The subsequent key down events (since they are handled by Menu.js component) doesn't trigger a body scroll as it calls `event.preventDefault()`. So I've added that to Popover.js as well.

### Improvement

As suggested by @mshwery I've modified the Menu.js component to use event delegation for registering the keydown event on menu parent component a single time instead of adding a keydown event listener on every menu item.

https://www.loom.com/share/aaa98e46ff0847dc8c1d030c5c15e602

## Testing

Local testing

### DIFF
--- a/src/menu/src/Menu.js
+++ b/src/menu/src/Menu.js
@@ -17,10 +17,12 @@ const Menu = memo(function Menu(props) {
 
   useEffect(() => {
 
-    menuItems.current = menuRef.current ? [
-      ...menuRef.current.querySelectorAll(
-        '[role="menuitemradio"], [role="menuitem"]'
-     )].filter(el => el.getAttribute('disabled') === null)
+    const currentMenuRef = menuRef.current
+
+    menuItems.current = currentMenuRef ? [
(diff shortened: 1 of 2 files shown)


mshwery:
Thanks @tusharf5 !

PR #964, opened 2020-09-02
add size prop to buttons

mshwery:
## Overview
This PR updates the base functional styles for Buttons, sets the `type = 'button'` when not otherwise specified, and adds a `size` prop to Buttons/IconButtons.

**TODO**
- [ ] support `size` prop for TextDropdownButton? 
- [ ] move `size` style values into the theme hook?

NA (no visual change)

## Testing


src/buttons/src/Button.js
```diff
@@ -27,12 +27,34 @@ const ButtonIcon = memo(function ButtonIcon({ icon, size, spacing, edge }) {
...
+  },
+  medium: {
+    height: 32,
+    textSize: 300
+  },
+  large: {
+    height: 40,
+    textSize: 400
+  }
+}
```
mshwery, line 43: I don't love this. It should probably move into the theme hook.

src/theme/src/hooks/useSegmentedControlAppearance.js
```diff
@@ -77,6 +77,7 @@ const getSegmentedControlStyles = theme => {
       }
     },
     [focusState]: {
+      zIndex: 1,
```
mshwery, line 80: I noticed that the focus style was funky when hovering other options in a segmented control. This helps it not look weird.

colinlohner (approved):

src/buttons/src/TextDropdownButton.js
```diff
@@ -6,15 +6,7 @@ import { CaretDownIcon } from '../../icons'
...
-
-const styles = {
-  position: 'relative',
-  fontFamily: 'ui',
-  fontWeight: 500,
-  display: 'inline-flex',
-  alignItems: 'center',
-  flexWrap: 'nowrap'
-}
+import { styles } from './Button'
```
akleiner2, line 9: What about just importing the `<Button />` component, and overriding some of the box props that are passed in? Don't think the glamor styles are too different

src/theme/src/hooks/useSegmentedControlAppearance.js
akleiner2: I think this comes from `constants/StackingOrder` -- would avoid arbitrary `zIndex` throughout here.

src/buttons/src/TextDropdownButton.js
mshwery: tbh this should be a variant... "unstyled"

src/buttons/src/Button.js
akleiner2: +1 - what if we wanted to let users configure these values? I like the size API, but as a user, I'd likely want to control  what `small` `medium` and `large` map to as sizes for height / text size. 

Should these be semantic tokens? I understand the APIs aren't super intuitive because we use `<Text />` underneath (which expects `size`), but I think that's something we can solve (at least for now) via documentation + types.

src/theme/src/hooks/useSegmentedControlAppearance.js
mshwery: Hm, this is different than stacking order though – this is relative to sibling z-index. zIndex 1 is all you need. There is no other zIndex value?

PR #971, opened 2020-09-03
Update inputs to use new theming style

colinlohner:
## Overview
Update inputs to follow new theming standard

image
image

## Testing


src/tag-input/src/TagInput.js
```diff
@@ -151,7 +151,7 @@ const TagInput = memo(
 
     const { tokens } = useTheme()
     const themedContainerClassName = useTagInputAppearance()
-    const themedInputClassName = useInputAppearance('none')
+    const { className: themedInputClassName, boxProps } = useInputAppearance({ appearance: 'none' })
```
mshwery, line 154: If there's only one appearance, should this just be the base style? (No modifiers?)

src/tag-input/src/TagInput.js
```diff
@@ -176,6 +176,7 @@ const TagInput = memo(
           size={300}
```
akleiner2, line 176: What do you think about converting this inner `<Text is "input" />` into something a bit more in line with the way we're going with stuff. Also - can't we bring in `<TextInput />` directly here?

src/text-input/src/TextInput.js
```diff
@@ -12,33 +11,30 @@ const TextInput = memo(
...
         is="input"
         className={cx(themedClassName, className)}
         type="text"
-        size={300}
         width={width}
-        height={height}
         required={required}
         disabled={disabled}
         placeholder={placeholder}
-        paddingLeft={Math.round(height / 2.6)}
```
akleiner2, line 35: Loving this cleanup!

src/textarea/src/Textarea.js
```diff
@@ -26,14 +26,13 @@ const Textarea = memo(
...
+    const { className: themedClassName, boxProps } = useInputAppearance()
 
     return (
-      <Text
+      <Box
         is="textarea"
         ref={ref}
         className={cx(themedClassName, className)}
-        size={300}
         width={width}
```
akleiner2, line 36: What do you think about pushing `height` out here like we did in `<TextInput />`?

colinlohner: I think you want to be able to control the height of the textarea as opposed to an input. Feels like we can leave this one

src/text-input/src/TextInput.js
```diff
@@ -12,33 +11,30 @@ const TextInput = memo(
       required,
       placeholder,
       width = 280,
-      height = 32,
```
mshwery, line 15: Will this work for us in app where we specify height in many places?

src/tag-input/src/TagInput.js
colinlohner: Probably could, haven't looked at that. I can do that aswell

src/theme/src/components/inputs.js
```diff
@@ -0,0 +1,85 @@
+import { minorScale } from '../../../scales'
+
+export default function getInputStyles(theme) {
```
mshwery, line 3: Can you have the filenames in this directory be singular? They should also match the name of the component where they are consumed so it's easy for us to track them down.

src/theme/src/components/inputs.js
```diff
@@ -0,0 +1,85 @@
...
+  const { tokens } = theme
+
+  return {
+    baseStyle: {
+      WebkitAppearance: 'none',
+      MozAppearance: 'none',
+      lineHeight: '12px',
+      border: 'none',
+      borderRadius: '4px',
```
mshwery, line 12: Can this come from tokens?

src/theme/src/themes/default/component-specific/inputs.js
```diff
@@ -1,46 +1,3 @@
...
-    },
-    focus: {
-      outline: 'none'
-    },
-    disabled: {
-      backgroundColor: tokens.colors.gray100
-    }
-  }
-}
+const inputs = {}
```
akleiner2, line 1: Any reason to still keep this around?

src/theme/src/components/inputs.js
```diff
@@ -0,0 +1,85 @@
...
+
+        _placeholder: {
+          color: tokens.colors.gray600
+        },
+
+        _placeholderHover: {
+          color: tokens.colors.gray700
+        }
+      },
+      none: {
```
mshwery, line 45: What's the difference between default vs none?

src/theme/src/themes/default/component-specific/inputs.js
colinlohner: Just following what matt had done with buttons

src/textarea/src/Textarea.js
mshwery: But you can still specify height? Won't it pass through via restProps?

src/theme/src/themes/default/component-specific/inputs.js
mshwery: Fine with deleting them!

src/theme/src/components/inputs.js
akleiner2: I think the `none` appearance should be `transparent` for the background! But otherwise it's meant to be one of those "floating" inputs (we use this in `<TagInput />` for example)

colinlohner: default is the bordered visual, none has no borders

src/tag-input/src/TagInput.js
colinlohner: This is an actual variant of the component. Maybe we could rename it to something else to make it more clear what it is

mshwery: Ah gotcha! Seems like it's not much of a variant if this is the only value we pass in! But I guess this is because we are reusing the same hook in multiple components?

src/textarea/src/Textarea.js
colinlohner: yeah, suppose one difference is just the default `height` of the textarea as opposed to the other inputs

src/tag-input/src/TagInput.js
colinlohner: Since we switched this to use the actual `<TextInput />` component this goes away :)

src/theme/src/components/input.js
```diff
@@ -0,0 +1,85 @@
+import { minorScale } from '../../../scales'
+
+export default function getInputStyles(theme) {
+  const { tokens } = theme
+
+  return {
+    baseStyle: {
```
mshwery, line 7: Should this also include `outline: 'none'` and some similar styles that buttons have?

src/text-input/src/TextInput.js
```diff
@@ -12,33 +11,32 @@ const TextInput = memo(
...
         height={height}
         required={required}
         disabled={disabled}
         placeholder={placeholder}
-        paddingLeft={Math.round(height / 2.6)}
-        paddingRight={Math.round(height / 2.6)}
-        borderRadius={minorScale(1)}
         spellCheck={spellCheck}
         aria-invalid={isInvalid}
         {...(disabled ? { color: 'muted' } : {})}
```
mshwery, line 37: Oh shoot – this will fail if you aren't using `Text`

src/text-input/src/TextInput.js
mshwery: I believe this will break a ton of the app that specify a height or rely on the default assignment here.

colinlohner: look at line `23`

src/text-input/src/TextInput.js
colinlohner: oh good catch

src/text-input/src/TextInput.js
colinlohner: Followed the pattern that you did for buttons

src/text-input/src/TextInput.js
colinlohner: Should probably move this into the style config

src/text-input/src/TextInput.js
colinlohner: So i think this should still work fine. There is now a size prop that has the correct default height. And if someone passes their own height that should get passed through with restprops

src/theme/src/components/input.js
```diff
@@ -0,0 +1,90 @@
...
```
mshwery, line 7: Do we need a base `color` for the font?

mshwery (approved):

PR #973, opened 2020-09-04
V6 pane appearance

mshwery:
## Overview
This PR moves some styles (as many as we can) into the pane's style config. Pane does a few funky things though that I punted on:
- it transforms `border` into `borderLeft|Right|Top|Bottom` (don't recall why)
- it applies `elevation` `hoverElevation` and `activeElevation` props as style modifiers and we aren't quite ready for supporting arbitrary modifiers in the style config object.

NA no change

## Testing


src/theme/src/hooks/usePaneAppearance.js
```diff
@@ -0,0 +1,18 @@
...
+  _hover: '&:hover',
+  _active: '&:active'
+}
+
+function usePaneAppearance(modifiers, internalStyles) {
+  const theme = useTheme()
+  const paneStyles = useMemo(() => getPaneStyles(theme), [theme])
+
+  return useStyleConfig(paneStyles, modifiers, pseudoSelectors, internalStyles)
+}
```
mshwery, line 16: I think this can end up getting pulled into `useStyleConfig`, but would require a key lookup on the `components/*` object to find the `getFooStyles` function. Then all of these `useFooAppearance` hooks go away.

colinlohner (approved):

src/theme/src/components/pane.js
```diff
@@ -0,0 +1,9 @@
+export default function getPaneStyles() {
+  return {
+    baseStyle: {
+      transitionDuration: '150ms',
+      transitionProperty: 'box-shadow, transform',
+      transitionTimingFunction: `cubic-bezier(0.0, 0.0, 0.2, 1)`
+    }
+  }
```
mshwery, line 8: `elevation` `hoverElevation` and `activeElevation` props can introduce `boxShadow` + `transform` css properties.

We could pull this into a modifier, but it's fairly redundant and could easily conflict with base styles or another modifier's styles:

```js
hoverElevations: {
  0: {
    _hover: {
      boxShadow: theme.tokens.elevations.0
    }
  },
  1: {
    translate: 'translateY(-2px)',
    _hover: {
      boxShadow: theme.tokens.elevations.1
    }
  },
  2: {
    translate: 'translateY(-2px)',
    _hover: {
      boxShadow: theme.tokens.elevations.2
    }
  },
  3: {
    translate: 'translateY(-2px)',
    _hover: {
      boxShadow: theme.tokens.elevations.3
    }
  },
}
```

colinlohner (approved):

src/theme/src/hooks/usePaneAppearance.js
akleiner2: Still think this abstraction makes sense because each internal `pseudoSelectors` are bound to the `useXAppearance()` hook, and not exposed anywhere else. 

Unless I'm missing something?

src/layers/src/Pane.js
```diff
@@ -32,90 +79,26 @@ const Pane = memo(
...
-        ':hover': {
-          ...(css[':hover'] || {}),
-          transform: 'translateY(-2px)',
-          boxShadow: theme.elevations[hoverElevation]
-        }
-      }
+    const styles = {
+      ...css,
+      background: theme.colors.background[background] || background,
+      boxShadow: theme.elevations[elevation],
```
akleiner2, line 85: I was going to leave a comment if it makes sense to namespace elevations under `tokens`, but I think once we go through and do all the theme cleanup of olde, we'll be left with `theme.tokens` housing everything, aka we should probably just call everything in there `theme`. Don't mind me!

src/layers/src/Pane.js
```diff
@@ -32,90 +79,26 @@ const Pane = memo(
...
-        transitionProperty: 'box-shadow, transform',
-        transitionTimingFunction: `cubic-bezier(0.0, 0.0, 0.2, 1)`,
-        ':hover': {
-          ...(css[':hover'] || {}),
-          transform: 'translateY(-2px)',
-          boxShadow: theme.elevations[hoverElevation]
-        }
-      }
+    const styles = {
+      ...css,
```
akleiner2, line 83: Oof - while we're here, can we yank this out? 😅

akleiner2:
@mshwery Oh! And one last thing - let's get rid of `getElevation` on the theme, now that we're pulling elevations directly

src/theme/src/hooks/usePaneAppearance.js
mshwery: That's correct, but it may make a ton more sense to move all of this out of the theme directory and have it colocated next to the component that uses it. If `useComponentAppearance` hook goes away, then the only remaining code (that could sit next to the component itself) is `pseudeoSelectors` + the style config fn.

Also I would love a world where the pseudoSelectors are actually standardized and not ad-hoc per component. Not too sure if that's possible yet.

src/theme/src/components/pane.js
mshwery: If we allow access to props in the style config fn we could do:
```js
baseStyles: {
  boxShadow: theme.elevations[props.elevation] || undefined,

  _hover: {
    ...(props.hoverElevation ? {
      boxShadow: theme.elevations[props.hoverElevation],
      transform: `translateY(-2px)`
    } : {}) // this is a bit ugly, and could be pulled out into a function for clarity.
}
```

src/layers/src/Pane.js
mshwery: Unfortunately no, because of the other internal components that rely on it.

akleiner2: Roger that - just noticed `<CornerDialog />`, `<Dialog />`, and `<SideSheet />` are still using it

src/theme/src/components/pane.js
```diff
@@ -0,0 +1,65 @@
...
+
+  if (border === true) {
+    return `1px solid ${theme.colors.border.default}`
+  }
+
+  return value
+}
+
+export default function getPaneStyles(theme, props) {
```
mshwery, line 48: I ended up going down this path because it makes the most sense for now and keeps _all the styles_ in one spot. cc @colinlohner @akleiner2

akleiner2: Spicy. Sounds good though!

src/theme/src/hooks/usePaneAppearance.js
```diff
@@ -0,0 +1,17 @@
+import useStyleConfig from '../../../hooks/use-style-config'
+import getPaneStyles from '../components/pane'
+import useTheme from '../useTheme'
+
+const pseudoSelectors = {
+  _hover: '&:hover',
+  _active: '&:active'
+}
+
+function usePaneAppearance(props, internalStyles) {
```
akleiner2, line 10: `internalStyles` is nonexistent at the callsite here, can we just add an empty obj for now as a default arg?

mshwery: That happens internally already, so we're good. I'd rather keep the useXAppearance signature consistent throughout

PR #976, opened 2020-09-05
 Convert <TextDropdownButton /> to new styling mechanism + do some cleanup / bugfixes

akleiner2:
## Overview
This moves over a bunch of stuff for `<TextDropdownButton />` into the new theming style, and fixes up a couple of bugs underneath around loading indicator, etc. 

## Testing

### DIFF
--- a/src/theme/src/components/text-dropdown-button.js
+++ b/src/theme/src/components/text-dropdown-button.js
@@ -1,0 +1,56 @@
+import { minorScale } from '../../../scales'
+import { controlStyles } from '../shared'
+
+export default function getTextDropdownButtonStyles(theme) {
+  const { tokens } = theme
+  const disabledStyles = controlStyles._disabled
+
+  return {
+    baseStyle: {
+      WebkitFontSmoothing: 'antialiased',
+      WebkitAppearance: 'none',
+      MozAppearance: 'none',
+      verticalAlign: 'middle',
+      textDecoration: 'none',
+      border: 'none',
+      fontFamily: tokens.fontFamilies.ui,
+      paddingX: minorScale(1),
+      marginX: minorScale(-1),
+      paddingY: 2,
+      marginY: -2,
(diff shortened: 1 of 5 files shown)


src/theme/src/components/text-dropdown-button.js
```diff
@@ -0,0 +1,56 @@
...
+      background: 'none',
+      borderRadius: tokens.borderRadius
+    },
+
+    appearances: {
+      default: {
+        _hover: {},
+        _disabled: { ...disabledStyles },
+        _focus: {
+          boxShadow: `0 0 0 2px ${tokens.colors.blue100}`,
```
mshwery, line 32: Should this whole value be a theme token   (focus ring)?

src/theme/src/components/text-dropdown-button.js
```diff
@@ -0,0 +1,56 @@
...
+          transition: 'box-shadow 80ms ease-in-out'
+        }
+      }
+    },
+
+    sizes: {
+      small: {
+        ...tokens.text[300],
```
mshwery, line 40: Fwiw this allows for users to pretty much pass in whatever they want unlike the other parts of this style config object.

mshwery (approved):

src/theme/src/components/text-dropdown-button.js
akleiner2: Yeah - wanted to do that in a more broad change later :thumbsup:

src/theme/src/components/text-dropdown-button.js
akleiner2: Yep, same case with `<Button />` too :/ Although, I think that we can likely solve this with some kind of strict interfaces. Think it's worth revisiting how we get the `<Text />` behavior a bit more elegantly, because I'm with you - this currently isn't super great

PR #979, opened 2020-09-08
Move table cell appearance over to `useStyleConfig` and avoid exposing on the main API

akleiner2:
##  Overview
Title, and updates some of the selectable table styles, too -- we can revisit these once we see what they look like on a broader scale. 

## Testing

### DIFF
--- a/src/theme/src/components/table-cell.js
+++ b/src/theme/src/components/table-cell.js
@@ -1,0 +1,23 @@
+export default function getTableCellStyles(theme) {
+  const { tokens } = theme
+  return {
+    baseStyle: {
+      paddingX: 12,
+      boxSizing: 'border-box',
+      flex: 1,
+      display: 'flex',
+      alignItems: 'center',
+      flexShrink: 0,
+      overflow: 'hidden'
+    },
+    appearances: {
+      default: {
+        _focus: {
+          outline: 'none',
+          background: tokens.colors.blue50,
+          boxShadow: `inset 0 0 0 1px ${tokens.colors.blue500}`
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


src/theme/src/components/table-cell.js
```diff
@@ -0,0 +1,23 @@
...
+      display: 'flex',
+      alignItems: 'center',
+      flexShrink: 0,
+      overflow: 'hidden'
+    },
+    appearances: {
+      default: {
+        _focus: {
+          outline: 'none',
+          background: tokens.colors.blue50,
```
akleiner2, line 17: Probably worth spitting this out into a `tableCellSelectedBackgroundColor` and `tableCellSelectedBorder` or something similar.

colinlohner (approved):

src/theme/src/hooks/useTableCellAppearance.js
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react'
+import useStyleConfig from '../../../hooks/use-style-config'
+import getTableCellStyles from '../components/table-cell'
+import useTheme from '../useTheme'
+
+const pseudoSelectors = {
+  _focus:
+    '&[data-isselectable="true"]:focus, &[aria-expanded="true"][aria-haspopup="true"]'
```
mshwery, line 8: Curious why we need `&[data-isselectable="true']:focus` instead of `&:focus`?

src/theme/src/components/table-cell.js
mshwery: I guess it depends on if you want changes to the foundational tokens to impact these component-specific values or not?

src/theme/src/hooks/useTableCellAppearance.js
akleiner2: We only want to apply focus styles to a selectable table cell - unsure of what reason, but this is preserved behavior.

mshwery: I wonder if the only cells that can receive focus on the ones that are selectable though...

PR #985, opened 2020-09-09
Update Row Styles

colinlohner:
## Overview
Update to use style config for row styling

Screen Shot 2020-09-09 at 3 38 32 PM
Screen Shot 2020-09-09 at 3 38 19 PM
Screen Shot 2020-09-09 at 3 38 14 PM

## Testing


src/menu/src/MenuOption.js
```diff
@@ -53,6 +52,7 @@ const MenuOption = memo(function MenuOption(props) {
       height={40}
```
akleiner2, line 52: Can we push `height`, `display='flex'`, and `alignItems='center'` into the theme config?

src/table/src/TableRow.js
```diff
@@ -83,6 +81,7 @@ const TableRow = memo(
           onKeyDown={handleKeyDown}
           height={height}
           borderBottom="muted"
```
akleiner2, line 83: Similar comment about pushing `borderBottom` into the props config too. I _think_ `splitBoxProps` should know how to spit that back out, but please correct me if I'm wrong.

src/theme/src/hooks/useRowApperance.js
```diff
@@ -0,0 +1,36 @@
...
+ * Get the className of an `row` element - either `<Table.Row />`, or `<MenuItem />` .
+ * @param {string} appearance
+ * @return {string} the appearance class name.
+ */
+
+function useRowApperance(modifiers, internalStyles) {
+  const theme = useTheme()
+  const rowStyles = useMemo(() => getRowStyles(theme), [theme])
+
+  const defaultModifiers = {
```
akleiner2, line 23: Can these be lifted out and passed in directly from default props?

src/themes/classic/index.js
```diff
@@ -31,8 +31,6 @@ import { headings, text, fontFamilies, paragraph } from './typography'
  * Component Specific.
  */
 import {
-  getRowClassName,
```
akleiner2, line 34: Beautiful.

src/table/src/TableRow.js
```diff
@@ -128,11 +127,6 @@ TableRow.propTypes = {
    */
   isHighlighted: PropTypes.bool,
 
-  /**
-   * The intent of the alert.
-   */
-  intent: PropTypes.oneOf(['none', 'success', 'warning', 'danger']),
```
mshwery, line 134: I think we are considering keeping `intent` around just to minimize API changes, especially since we can support it pretty easily with the theming solution we have now.

PR #986, opened 2020-09-09
Update checkbox theme to use style config

colinlohner:
## Overview
Update checkbox/radio to use style config

image
image

## Testing

### DIFF
--- a/src/checkbox/src/Checkbox.js
+++ b/src/checkbox/src/Checkbox.js
@@ -60,7 +60,7 @@ const Checkbox = memo(
       }
     }, [ref, indeterminate])
 
-    const themedClassName = useCheckboxAppearance(appearance)
+    const { className: themedClassName, boxProps } = useCheckboxAppearance({ appearance })
 
     return (
       <Box
@@ -83,6 +83,7 @@ const Checkbox = memo(
           disabled={disabled}
           aria-invalid={isInvalid}
           ref={callbackRef}
+          {...boxProps}
         />
         <Box
           boxSizing="border-box"
--- a/src/theme/src/hooks/useCheckboxAppearance.js
+++ b/src/theme/src/hooks/useCheckboxAppearance.js
@@ -1,127 +1,19 @@
 import { useMemo } from 'react'
-import { css } from 'glamor'
+import useStyleConfig from '../../../hooks/use-style-config'
+import getCheckboxStyles from '../components/checkbox'
 import useTheme from '../useTheme'
 
-const defaultState = '& + div'
-const disabledState = '&[disabled] + div'
-const hoverState = '&:not([disabled]):hover + div'
-const focusState = '&:not([disabled]):focus + div'
-const activeState = '&:not([disabled]):active + div'
-const checkedState = '&:checked + div, &[type=checkbox]:indeterminate + div'
-const checkedHoverState =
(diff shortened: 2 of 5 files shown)


src/theme/src/hooks/useCheckboxAppearance.js
```diff
@@ -1,140 +1,35 @@
...
+  _checkedDisabled: '&[disabled]:checked + div, &[type=checkbox][disabled]:indeterminate + div'
 }
 
 /**
  * Get the className of a `Checkbox`.
  * @param {string} appearance
  * @return {string} the appearance class name.
  */
-function useCheckboxAppearance() {
+function useCheckboxAppearance(modifiers) {
```
akleiner2, line 24: Sweet - does `size` make sense as a modifier for input components?

src/theme/src/hooks/useCheckboxAppearance.js
```diff
@@ -1,140 +1,35 @@
...
   const theme = useTheme()
-  const className = useMemo(() => css(getCheckboxStyles(theme)).toString(), [
-    theme
-  ])
-  return className
+  const checkboxStyles = useMemo(() => getCheckboxStyles(theme), [theme])
```
akleiner2, line 26: The thing that I'm _slowly_ starting to come around on a full theme implementation is that we need some way for _most_ of the visual styles to be derived from the theme in some way (along with some of the basic box props). 

The previous `useCheckboxAppearance` was an early-ish experiment to see what overriding styles could look like, but with this "style config" approach, I think that we can push a lot of the "visual" styles for the checkbox (i.e. checked background color, border, box shadow) into the theme, while leaving a lot of the internal "reset" styles to be ...well, internal. What do you think?

mshwery:
Looks like there's a dark border on these that isn't in the designs: 
Screen Shot 2020-09-12 at 4 12 52 PM

src/checkbox/src/Checkbox.js
```diff
@@ -60,7 +60,7 @@ const Checkbox = memo(
       }
     }, [ref, indeterminate])
 
-    const themedClassName = useCheckboxAppearance(appearance)
+    const { className: themedClassName, boxProps } = useCheckboxAppearance({ appearance })
```
mshwery, line 63: This will become
```suggestion
    const { className: themedClassName, ...styleProps } = useStyleConfig(
      'Checkbox',
      { appearance },
      pseudoSelectors,
      internalStyles
    )
```

src/theme/src/hooks/useCheckboxAppearance.js
```diff
@@ -1,140 +1,35 @@
...
-      ...checkedStyle,
-      color: 'white',
-      boxShadow: `inset 0 0 0 -1px ${primary.active}`,
-      backgroundColor: primary.active,
-      ...checkedActiveStyles
-    }
-  }
+const pseudoSelectors = {
+  _display: '& + div',
+  _tickSvg: '& + div > svg',
```
mshwery, line 8: Hm, this presents a challenge for how we handle subcomponent styling. I don't know that aliasing a selector is the right move, but happy to move forward with this to unblock things for now. It feels non-standard.

src/theme/src/hooks/useCheckboxAppearance.js
mshwery: IMO yes! We can bring that as a follow up PR if that's too many changes though :)

PR #987, opened 2020-09-10
Forward TextareaField ref

sepehr500:
## Overview
Forward ref for the TextareaField to match TextInputField. I ran into this issue with I was trying to use this library with [react-hook-form](https://github.com/react-hook-form/react-hook-form) which uses refs. The library works with TextInputField because it forwards refs but not TextareaField because it does not forward a ref.
## Testing


akleiner2 (approved):
Seems legit to me -- +1 for consistency

colinlohner (approved):

PR #991, opened 2020-09-10
move onto a new theming paradigm

mshwery:
## Overview
This PR moves a bunch of things around with a more flexible theming approach.

NA

## Testing


src/theme/src/theme-tools.js
```diff
@@ -0,0 +1,47 @@
+export function get(obj, path, fallback) {
+  const keys = path && path.split ? path.split('.') : [path]
+
+  let value = obj
+  for (const key of keys) {
+    value = value ? value[key] : undefined
+  }
+
+  return value === undefined ? fallback : value
```
akleiner2, line 9: Recursing here _one_ more time would give us token aliases, which I think would be a pretty nice feature.

mshwery: True! Though you could also define the alias with a hard value instead of a token string (you should have access to it when defining the theme anyway)

src/themes/default/tokens/space.js
```diff
@@ -0,0 +1,3 @@
+const space = {}
```
akleiner2, line 1: Would be genuinely curious to see how you think this should be used.

src/themes/default/tokens/sizes.js
```diff
@@ -0,0 +1,3 @@
+const sizes = {}
```
akleiner2, line 1: Same with sizes across the theme - should everything just fall back to these?

akleiner2 (approved):
Looks awesome

src/themes/default/tokens/space.js
mshwery: Thinking this would/could be used for padding/margin values. I'm not 100% sure how we can safely migrate to this coming from the theme instead of using `majorScale` because today we already pass unitless numbers to `margin*` props. It would make it impossible whether to use the theme scales or treat the number as a pixel value.

src/themes/default/tokens/sizes.js
mshwery: Sizes would be used for widths/heights/etc. But I could see this going away.

PR #992, opened 2020-09-10
Migrate from ava -> jest 

akleiner2:
## Overview 
More modernizing of the client-side tooling. 

## Testing


colinlohner (approved):

.eslintrc.json
```diff
@@ -1,7 +1,8 @@
 {
   "parser": "babel-eslint",
   "env": {
-    "browser": true
+    "browser": true,
+    "jest": true
```
mshwery, line 5: Should we pull this into overrides so it only affects `*.test.*` files?

package.json
```diff
@@ -31,10 +31,11 @@
   ],
   "sideEffects": false,
   "scripts": {
-    "test": "yarn run eslint && ava",
+    "test": "yarn run eslint && yarn run jest",
```
mshwery, line 34:
```suggestion
    "test": "yarn eslint && yarn jest",
```

package.json
```diff
@@ -31,10 +31,11 @@
...
     "eslint": "eslint 'src/**/*.js'",
     "prepublishOnly": "rm -rf esm commonjs umd && yarn build",
     "dev": "start-storybook -p 6006",
+    "jest": "jest",
```
mshwery, line 38: This can be removed

mshwery (approved):
Curious if `npx yarn-deduplicate && yarn install` yields further yarn.lock changes?

.eslintrc.json
akleiner2: Probably - I think a lot of the jest-specific syntax only happens in `*.test.js` files, but we can be double safe 👍

PR #993, opened 2020-09-10
get Code from theme

mshwery:
## Overview
This PR moves `<Code />` over to the new theming format (most styles from the theme)

Screen Shot 2020-09-10 at 5 15 36 PM

## Testing


colinlohner (approved):
Lgtm

src/typography/src/Code.js
```diff
@@ -1,21 +1,29 @@
...
+const internalStyles = {}
+
 const Code = memo(
   forwardRef(function Code(props, ref) {
     const { className, appearance = 'default', ...restProps } = props
 
-    const { boxProps } = useCodeAppearance({ appearance })
+    const styleProps = useStyleConfig(
+      'Code',
+      { appearance },
```
mshwery, line 15: This really should pass `props` in entirety but we also need the default value somehow... I think we could define the theme defaults alongside the component theming object itself?

PR #995, opened 2020-09-11
v6 buttons in new style paradigm

akleiner2:
## Overview
Ports over v5 buttons and selects into the new paradigm 

## Testing


colinlohner (approved):
Luv et

src/themes/classic/components/select.js
```diff
@@ -0,0 +1,43 @@
+import { defaultControlStyles } from '../deprecated/shared'
```
mshwery, line 1: If we want to keep `shared` around, feel free to move it under `/components` instead of in `deprecated`!

src/buttons/src/IconButton.js
```diff
@@ -2,14 +2,32 @@ import React, { memo, forwardRef } from 'react'
...
+      size = 'medium',
+      intent,
+      appearance,
+      ...restProps
+    } = props
+
+    // Composes the exact same styles as button
+    const { className: themedClassName, ...boxProps } = useStyleConfig(
+      'Button',
+      { appearance, intent, size },
```
mshwery, line 26: Thinking for consumers we _may_ want to pass the entire props object... but tbd. Kind of false advertising if we dont

mshwery (approved):
🔥

src/buttons/src/IconButton.js
akleiner2: Yeah, probably - would make writing themes be the most complete. I guess we can revisit and see if we want to or not!

PR #997, opened 2020-09-11
v5/v6 Alert themes

mshwery:
## Overview
This PR migrates Alert to theme-based styling.

Kapture 2020-09-11 at 09 41 48

## Testing


src/alert/src/Alert.js
```diff
@@ -60,7 +68,8 @@ const Alert = memo(
               marginBottom={0}
               fontWeight={500}
               lineHeight={1}
-              color={intents[intentToken].text}
+              // Get this from the theme / props on the Alert
+              color="inherit"
```
mshwery, line 72: Couldn't think of a simpler way to handle this... I dont like reaching into the theme directly here, so I chose to inherit whatever color is defined on the Alert wrapper itself.

src/alert/src/Alert.js
```diff
@@ -21,24 +33,20 @@ const Alert = memo(
...
 
-    const {
-      tokens: { intents }
-    } = useTheme()
-
     const intentToken = intent === 'none' ? 'info' : intent
-    const className = useAlertApperance(intentToken)
+    const { className: themedClassName, ...styleProps } = useStyleConfig(
+      'Alert',
+      { ...props, appearance, intent: intentToken },
```
mshwery, line 39: Would like to pass `props` as-is but unfortunately we have different intents (`none` and `info`)... Thoughts on allowing for both? It's a simple update in the theme to add tokens for `none` and `info` that are the same.

src/alert/src/Alert.js
```diff
@@ -1,18 +1,30 @@
...
 import { getIconForIntent } from './getIconForIntent'
 
+const pseudoSelectors = {}
+
+const internalStyles = {
+  position: 'relative',
+  overflow: 'hidden',
+  display: 'flex',
+  // 15 instead of 16 in order to maintain height with 1px border
+  padding: '15px'
```
mshwery, line 19: TODO evaluate moving this into the theme styles too?

PR #998, opened 2020-09-11
v6 tooltips in the new paradigm

akleiner2:
## Overview
This ports over Tooltips to the new styling paradigm and removing some additional stuff. 

image

image

## Testing


colinlohner (approved):

src/tooltip/src/TooltipStateless.js
```diff
@@ -2,13 +2,14 @@ import React, { memo, forwardRef } from 'react'
...
 import { Paragraph } from '../../typography'
-import useTooltipStyle from '../../theme/src/hooks/useTooltipStyle'
+import useStyleConfig from '../../hooks/use-style-config'
 
 const TooltipStateless = memo(
   forwardRef(function TooltipStateless(props, ref) {
     const { children, appearance, ...restProps } = props
-    console.log(appearance)
-    const { color, ...themedProps } = useTooltipStyle(appearance)
+    const { ...boxProps } = useStyleConfig('Tooltip', { appearance }, {}, {})
```
mshwery, line 10: This doesn't need to be spread
```suggestion
    const styleProps = useStyleConfig('Tooltip', { appearance }, {}, {})
```

Also can you make those empty objects consistent:
```js
const pseudoSelectors = {}
const internalStyles = {}
```

(We could also make them optional and have this behavior inside `useStyleConfig` to preserve stable object references and reduce hook reruns).

mshwery (approved):
Minor nits, but looks great!

src/tooltip/src/TooltipStateless.js
akleiner2: +1 to making these optional, but out of scope, for now :)

PR #1000, opened 2020-09-11
v6 TextInput, TagInput, and Textarea all in new theme paradigm.

akleiner2:
## Overview
Backwards-compat input, taginput, and textarea in the new theming paradigm 👍 

## Testing


src/text-input/src/TextInput.js
```diff
@@ -2,7 +2,26 @@ import React, { forwardRef, memo } from 'react'
...
-import useInputAppearance from '../../theme/src/hooks/useInputAppearance'
+import useStyleConfig from '../../hooks/use-style-config'
+
+const pseudoSelectors = {
+  _focus: '&:focus',
+  _disabled: '&:disabled',
+  _invalid: '&[aria-invalid="true"]',
+  _placeholder: '&::placeholder',
+  _placeholderHover: '&:hover::placeholder',
+  _placeholderFocus: '&:focus::placeholder'
```
mshwery, line 13: I'm still not a huge fan of placeholder hover/focus states being different :)

src/themes/classic/components/index.js
```diff
@@ -4,12 +4,16 @@ import Code from './code'
...
 
 export default {
   Alert,
   Button,
   Code,
   Pane,
   Select,
-  Tooltip
+  Tooltip,
+  Input,
```
mshwery, line 17: Mind alphabetizing? (should really be automatic formatting imo)

src/textarea/src/Textarea.js
```diff
@@ -26,7 +41,12 @@ const Textarea = memo(
       ...restProps
     } = props
 
-    const { className: themedClassName, boxProps } = useInputAppearance()
+    const { className: themedClassName, ...boxProps } = useStyleConfig(
+      'Input',
```
mshwery, line 45: We probably need to call out that these don't match the component names in the theming docs.

mshwery: Or in the theme you could split them up and import the same file for both exports `TextInput` and `Textarea`?

src/themes/classic/components/input.js
```diff
@@ -0,0 +1,84 @@
...
+  fontFamily: 'fontFamilies.ui',
+  lineHeight: '12px',
+  border: '1px solid transparent'
+}
+
+const appearances = {
+  default: {
+    backgroundColor: 'white',
+    boxShadow: theme =>
+      `inset 0 0 0 1px ${theme.scales.neutral.N5A}, inset 0 1px 2px ${theme.scales.neutral.N4A}`,
```
mshwery, line 12: Is this in the `shadows`?

mshwery (approved):
🔥

src/themes/classic/components/index.js
akleiner2: i'll follow up with something that'll enforce this more broadly across the eg codebase 👍

src/textarea/src/Textarea.js
akleiner2: Yeah, I definitely want to break these apart.

PR #1001, opened 2020-09-11
move InlineAlert to new theming

mshwery:
## Overview
This PR moves `InlineAlert` to the new theming

Screen Shot 2020-09-11 at 2 08 07 PM
Screen Shot 2020-09-11 at 2 08 11 PM

## Testing

### DIFF
--- a/src/alert/src/InlineAlert.js
+++ b/src/alert/src/InlineAlert.js
@@ -1,36 +1,51 @@
 import React, { memo, forwardRef } from 'react'
+import cx from 'classnames'
 import PropTypes from 'prop-types'
 import { spacing, dimensions, position, layout } from 'ui-box'
 import { Pane } from '../../layers'
 import { Text } from '../../typography'
-import { useTheme } from '../../theme'
+import useStyleConfig from '../../hooks/use-style-config'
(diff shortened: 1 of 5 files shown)


src/alert/src/InlineAlert.js
```diff
@@ -1,36 +1,51 @@
...
 
-    const {
-      tokens: { intents }
-    } = useTheme()
-
     const intentToken = intent === 'none' ? 'info' : intent
-    const textColor = intents[intentToken].text
+    const { className: themedClassName, ...styleProps } = useStyleConfig(
+      'InlineAlert',
+      { intent: intentToken },
```
akleiner2, line 31: Feels weird to support `intent="none"` and `intent="info"` and have them map to the same thing, no?

akleiner2 (approved):

src/alert/src/InlineAlert.js
mshwery: 100% agree. Looks like v5 and earlier supported `intent="none"` but internally treated "none" the same as "info".

We could drop it, but it will require rewriting a bunch of components to use `intent="info"` (or omit the intent)

PR #1004, opened 2020-09-11
fix IconButton in v5/v6

mshwery:
## Overview
This fixes the IconButton appearance/intent for v6 and v5 theming

Screen Shot 2020-09-11 at 6 28 10 PM
Screen Shot 2020-09-11 at 6 28 13 PM

## Testing

### DIFF
--- a/src/select-menu/src/SelectMenuContent.js
+++ b/src/select-menu/src/SelectMenuContent.js
@@ -24,7 +24,7 @@ const DefaultTitleView = ({ close, title, headerHeight }) => (
     <IconButton
       icon={CrossIcon}
       appearance="minimal"
-      height={24}
+      size="small"
       onClick={close}
       border="none"
     />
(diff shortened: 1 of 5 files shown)


src/select-menu/src/SelectMenuContent.js
```diff
@@ -24,7 +24,7 @@ const DefaultTitleView = ({ close, headerHeight, title }) => (
     <IconButton
       icon={CrossIcon}
       appearance="minimal"
-      height={24}
+      size="small"
```
akleiner2, line 27: Should we decouple this from the actual theme?

mshwery: Hm... curious what you mean?

mshwery: We could use `height` internally, since we dont know what's in the theme.

akleiner2: Yeah, think we should use `height` directly instead. This doesn't need to respect the theme imo.

PR #1009, opened 2020-09-13
v6 checkbox, radios, and switch all consuming from style config

akleiner2:
**Overview**
Checkboxes, radios, and switches all over into the new style config. Still some work cleaning up some of 

**Documentation**


src/checkbox/src/Checkbox.js
```diff
@@ -35,6 +34,42 @@ MinusIcon.propTypes = {
...
+  _hover: '&:not([disabled]):hover + div',
+  _focus: '&:not([disabled]):focus + div',
+  _active: '&:not([disabled]):active + div',
+  _checked: '&:checked + div, &[type=checkbox]:indeterminate + div',
+  _checkedHover:
+    '&:not([disabled]):checked:hover + div, &[type=checkbox]:not([disabled]):indeterminate:hover + div',
+  _checkedActive:
+    '&:not([disabled]):checked:active + div, &[type=checkbox]:not([disabled]):indeterminate:active + div',
+  _checkedDisabled:
+    '&[disabled]:checked + div, &[type=checkbox][disabled]:indeterminate + div'
```
mshwery, line 49: Not for this PR but I would love to really understand why we need these combinations (presumably to preserve specificity). They feel non-standard since they are highly tailored to the checkbox component.

src/checkbox/src/Checkbox.js
```diff
@@ -35,6 +34,42 @@ MinusIcon.propTypes = {
 
 const noop = () => {}
 
+const pseudoSelectors = {
+  _base: '& + div',
```
mshwery, line 38: What does `_base` mean?

src/checkbox/src/Checkbox.js
```diff
@@ -35,6 +34,42 @@ MinusIcon.propTypes = {
...
+  position: 'absolute',
+  whiteSpace: 'nowrap',
+  WebkitFontSmoothing: 'antialiased',
+  textDecoration: 'none',
+  WebkitAppearance: 'none',
+  MozAppearance: 'none',
+  width: '1px',
+  opacity: '0',
+
+  [pseudoSelectors._base]: {
```
mshwery, line 67: You don't need this. You can just do `_base`. The pseudoSelectors placeholders aren't replaced until the final step after all the styles (including `internalStyles`) have been merged together.

src/switch/src/Switch.js
```diff
@@ -65,6 +65,35 @@ CheckIcon.propTypes = {
 
 const noop = () => {}
 
+const pseudoSelectors = {
+  _base: '& + div',
+  _disabled: '&[disabled] + div',
+  _hover: '&:not([disabled]):hover + div',
+  _focus: '&:not([disabled]):focus + div',
```
mshwery, line 72: On the parent component we could use `_focusWithin: '&:focus-within'` instead of all this sibling stuff.

We would probably need to bubble the `disabled` attr to the wrapping component, perhaps with `aria` if it makes sense.

src/themes/classic/components/checkbox.js
```diff
@@ -0,0 +1,66 @@
+import { getPrimaryButtonStylesForIntent } from '../deprecated/helpers'
```
mshwery, line 1: If it's worth keeping, we can move it out of `/deprecated/*`

src/themes/classic/components/checkbox.js
```diff
@@ -0,0 +1,66 @@
...
+    },
+    _hover: {
+      background: theme =>
+        `linear-gradient(to top, ${theme.scales.neutral.N2A}, ${theme.scales.neutral.N1A})`,
+      boxShadow: theme =>
+        `inset 0 0 0 1px ${theme.scales.neutral.N4A}, inset 0 -1px 1px 0 ${theme.scales.neutral.N2A}`
+    },
+    _focus: {
+      boxShadow: theme =>
+        `0 0 0 2px ${theme.scales.blue.B4A}, inset 0 0 0 1px ${theme.scales.neutral.N5A}, inset 0 -1px 1px 0 ${theme.scales.neutral.N3A}`
```
mshwery, line 27: There is a `theme.shadows.focusRing` you can use to pull that in.

src/themes/default/components/checkbox.js
```diff
@@ -0,0 +1,52 @@
...
+    _disabled: {
+      cursor: 'not-allowed',
+      background: 'colors.gray100',
+      boxShadow: theme => `inset 0 0 0 1px ${theme.colors.gray100}`
+    },
+    _hover: {
+      boxShadow: theme => `inset 0 0 0 1px ${theme.colors.gray600}`
+    },
+    _focus: {
+      boxShadow: theme => `0 0 0 2px ${theme.colors.blue100}`
```
mshwery, line 19:
```suggestion
      boxShadow: 'shadows.focusRing'
```

mshwery (approved):
LGTM! Left some non-blocking comments about some of the legacy patterns in place here, but beyond the scope of this initial PR.

src/themes/classic/components/checkbox.js
akleiner2: Yeah - this is definitely worth keeping. I'll follow up since we use this in a few places.

src/themes/classic/components/checkbox.js
akleiner2: Cool! Wasn't sure if you landed your changes already :)

src/checkbox/src/Checkbox.js
akleiner2: `_base` is the target element for the styles to get applied to: notice that we put the classname on the actual input to do a lot of the work around hiding the default DOM checkbox. Otherwise, all of these styles would go to the DOM `input`, which isn't what we want.

src/checkbox/src/Checkbox.js
akleiner2: I'm with you here - I think it's worth re-visiting a lot of the psuedoselectors throughout EG in general.

colinlohner: +1 on revisiting the selectors

src/checkbox/src/Checkbox.js
colinlohner: Could we call it _displayElement or something a little more descriptive?

colinlohner (approved):
Approved, agree with @mshwery 's comments tho

PR #1012, opened 2020-09-14
Update Dialog "title" prop type in declaration file

williamgrosset:
**Overview**
This PR introduces a 1-line change in the declaration file to match the [expected](https://github.com/segmentio/evergreen/blob/master/src/dialog/src/Dialog.js#L254) `title` prop type of the `<Dialog />` component.

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -766,7 +766,7 @@ export interface DialogProps {
   /**
    * Title of the Dialog. Titles should use Title Case.
    */
-  title?: string
+  title?: string | React.ReactNode
   /**
    * When true, the header with the title and close icon button is shown.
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -766,7 +766,7 @@ export interface DialogProps {
   /**
    * Title of the Dialog. Titles should use Title Case.
    */
-  title?: string
+  title?: string | React.ReactNode
```
akleiner2, line 769: @williamgrosset `React.ReactNode` is a superset type that includes `string`, so let's just keep it as `React.ReactNode` for simplicity :) 

```suggestion
  title?: React.ReactNode
```

PR #1020, opened 2020-09-14
Support `<Text />` and `<Paragraph />` via the new theme

akleiner2:
**Overview**
Title - backfitting a bunch of the new typography tokens into `<Text />`, `<Paragraph />`, etc.

**Documentation**


src/typography/src/Text.js
```diff
@@ -1,19 +1,15 @@
...
+import useStyleConfig from '../../hooks/use-style-config'
+
+const emptyObject = {}
 
 const Text = memo(
   forwardRef(function Text(props, ref) {
-    const {
-      className,
-      color = 'default',
-      fontFamily = 'ui',
```
mshwery, line 11: We probably need to support `fontFamily` still?

src/typography/src/Text.js
```diff
@@ -1,19 +1,15 @@
...
-import useTextStyle from '../../theme/src/hooks/useTextStyle'
+import useStyleConfig from '../../hooks/use-style-config'
+
+const emptyObject = {}
 
 const Text = memo(
   forwardRef(function Text(props, ref) {
-    const {
-      className,
-      color = 'default',
```
mshwery, line 10: We probably want to keep supporting `color` values from the theme?

src/typography/src/Paragraph.js
```diff
@@ -1,30 +1,22 @@
...
   forwardRef(function Paragraph(props, ref) {
-    const { color = 'default', marginTop, size = 400, ...restProps } = props
+    const { marginTop, size = 400, ...restProps } = props
 
-    const { marginTop: defaultMarginTop, ...textStyle } = useParagraphStyle(
-      size,
-      color
+    const textStyle = useStyleConfig(
+      'Paragraph',
+      { size },
```
mshwery, line 14: Same here – think we need to support theme lookup for `color` and `fontFamily` like we used to do in v5. (We may need to check that inputs still support this as well!)

mshwery (approved):
seems legit! a couple breaking changes in here w/r/t `color` and `fontFamily` but otherwise looks great!

src/typography/src/Text.js
akleiner2: These happen in the theme itself!

mshwery: We support these as props though – you can do `<Text fontFamily="mono" />` (see `Code` for instance).

PR #1023, opened 2020-09-14
Update Text, Paragraph to be aware of `fontFamily` and `color`

akleiner2:
**Overview**
The themes are aware of `size` as a prop, but not the other props on the theme. We could probably pass those in, but this should roughly accomplish the same thing, as we compose `<Text />` quite heavily throughout the codebase internally. 

Still need to clean up all of the `marginTop="default"` throughout the codebase. 

N/A

**Documentation**


mshwery (approved):
Looks good! Can you bring the same thing for `fontFamily` to inputs?

colinlohner (approved):

PR #1025, opened 2020-09-14
Move tabs over to new style

colinlohner:
**Overview**
Update tabs to use new theming style

V6 Theme
image

V5 theme
image

**Documentation**


src/themes/classic/components/tab.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+
+  _focus: {
+    boxShadow: 'shadows.focusRing'
+  }
+}
+
+const appearances = {
+  primary: tabStyles,
+  secondary: tabStyles,
```
akleiner2, line 31: Does the default theme have a `secondary` appearance for tabs if they're the same?

src/themes/classic/components/tab.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export default {
+  baseStyle,
+  appearances
```
akleiner2, line 36: For the sake of consistency elsewhere, should we have a `const sizes = {}`?

src/themes/classic/components/tab.js
colinlohner: The default theme has two yes

src/tabs/src/Tab.js
```diff
@@ -1,27 +1,60 @@
...
+import useStyleConfig from '../../hooks/use-style-config'
 import safeInvoke from '../../lib/safe-invoke'
 import warning from '../../lib/warning'
-import useTabApperance from '../../theme/src/hooks/useTabApperance'
 import { Text } from '../../typography'
 
 const noop = () => {}
 
+const getInternalStyles = (direction) => ({
+  fontWeight: 500,
```
mshwery, line 11: Thinking `fontWeight` should probably be in `baseStyle` in the theme?

src/tabs/src/Tab.js
```diff
@@ -73,12 +106,14 @@ const Tab = memo(
 
     return (
       <Text
-        className={tabClassName}
+        className={themedClassName}
```
mshwery, line 109: Can we do `cx(className, themedClassName)` to make sure we support external classnames being passed in?

src/themes/classic/components/tab.js
mshwery: Seems like if there is no difference in appearances for v6 theme, I would move this all into `baseStyle` instead

src/themes/default/components/tab.js
```diff
@@ -0,0 +1,82 @@
...
+    paddingBottom: '16px',
+    paddingLeft: '2px',
+    paddingRight: '2px',
+    position: 'relative',
+
+    ':not(:last-child)': {
+      marginRight: '20px'
+    },
+    
+    ':before': {
```
mshwery, line 18: Can we just bridge support for this by adding pseudo placeholders for `_before` and `_after`?

src/themes/classic/components/tab.js
colinlohner: hmm not sure how i feel about having unused objects floating around

mshwery (approved):
Lookin good! Left a few minor comments.

akleiner2 (approved):

PR #1036, opened 2020-09-15
Row and MenuItem

colinlohner:
**Overview**
Move table row and menu item/option to usestyleconfig

image
image

**Documentation**


src/themes/default/components/table-row.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const getBackgroundForIntentAndState = (intent, state) => colorMap[intent][state]
+
+const baseStyle = {
+  outline: 'none',
+  textDecoration: 'none',
+  height: 64,
+
+  '&[data-isselectable="true"]': {
```
mshwery, line 42: Can we push this into the pseudoSelectors placeholders instead?

src/table/src/TableRow.js
```diff
@@ -67,22 +72,32 @@ const TableRow = memo(
...
+    const { className: themedClassName, ...boxProps } = useStyleConfig(
+      'TableRow',
+      { appearance, intent },
+      pseudoSelectors,
+      {
+        display: 'flex'
+      }
+    )
+
+    const { height: themeHeight, ...restBoxProps } = boxProps
```
mshwery, line 84: Might as well pull this out at the same time as we do `themedClassName` above to avoid the cost of 2 spreads.

src/table/src/TableRow.js
```diff
@@ -67,22 +72,32 @@ const TableRow = memo(
...
     }
 
-    const themedClassName = theme.getRowClassName(appearance, intent)
+    const { className: themedClassName, ...boxProps } = useStyleConfig(
+      'TableRow',
+      { appearance, intent },
+      pseudoSelectors,
+      {
+        display: 'flex'
+      }
```
mshwery, line 81: Please pull this out into a static object so the benefits of memoization come into play. Right now this passes a new object each render.

src/menu/src/MenuOption.js
```diff
@@ -1,12 +1,19 @@
...
-import { useTheme } from '../../theme'
+import { pseudoSelectors } from '../../table/src/TableRow'
 import { Text } from '../../typography'
 
 const noop = () => {}
 
+const internalStyles = {
+  display: 'flex',
+  alignItems: 'center',
+  height: 40,
```
mshwery, line 14: Should this height be in the theme?

src/menu/src/MenuItem.js
```diff
@@ -44,7 +48,14 @@ const MenuItem = memo(
...
 
-    const themedClassName = theme.getMenuItemClassName(appearance, 'none')
+    const { className: themedClassName, ...boxProps } = useStyleConfig(
+      'MenuItem',
+      { appearance },
+      pseudoSelectors,
+      internalStyles
+    )
+
+    const { height: themeHeight, ...restBoxProps } = boxProps
```
mshwery, line 58:
```suggestion
    const { className: themedClassName, height: themedHeight, ...boxProps } = useStyleConfig(
      'MenuItem',
      { appearance },
      pseudoSelectors,
      internalStyles
    )
```

mshwery (approved):
Looks great! left a couple minor suggestions

src/menu/src/MenuOption.js
colinlohner: I think i could move it out of Menu option. but for table row it should

PR #1037, opened 2020-09-16
Add a codemod to replace `marginTop="default"` with the actual value

akleiner2:
**Overview**
This should help ease some of the burden of consumers upgrading, especially if they relied on this behavior before. 

**Documentation**


mshwery (approved):
Nice!

PR #1038, opened 2020-09-16
V6 use clickable

mshwery:
**Overview**
This introduces a new hook (a couple actually!) called `use-clickable`. It implies the ability to gain focus (through `tabIndex`). It uses `onKeyDown` to capture Enter or Space keys. This is useful for things like Button or Tab even when they aren't a `button` element (e.g. have `role="button"`). It also ensures consistent behavior across browsers.

- added `useLatest` which just holds onto a mutable value so it can be can be used in other hooks
- removed deprecated style config hook (no longer used)
- unified exports for useStyleConfig to make consistent with other `/hooks/*`

NA

**Documentation**


colinlohner (approved):
This looks awesome

src/tabs/src/Tab.js
```diff
@@ -59,19 +60,18 @@ const Tab = memo(
...
+    const onClickRef = useLatest(props.onClick)
+    const handleClick = useCallback(
+      event => {
+        safeInvoke(onClickRef.current, event)
+        if (!disabled) {
+          onSelect()
+        }
+      },
+      [disabled, onSelect]
+    )
```
mshwery, line 72: I had considered a `use-combined-callback` hook, but might be a little overkill for now (unless we see this pattern happening all over the place).

mshwery:
Alright, made some updates here – this is ready for review again @akleiner2 @colinlohner

src/hooks/use-clickable.js
```diff
@@ -0,0 +1,44 @@
...
+        event.preventDefault()
+        event.currentTarget.click()
+      }
+    },
+    [disabled]
+  )
+
+  return {
+    // TODO import useFocusable as well (needs to be focusable)
+    tabIndex: disabled ? -1 : tabIndex,
```
mshwery, line 41: If we had a ref to the underlying object (in the future) we could detect whether it's an element that supports `disabled` which means we could pass `undefined` instead of `-1`

src/menu/src/MenuItem.js
```diff
@@ -1,15 +1,20 @@
...
 import { Pane } from '../../layers'
-import safeInvoke from '../../lib/safe-invoke'
-import { pseudoSelectors } from '../../table/src/TableRow'
+import { pseudoSelectors as rowPseudos } from '../../table/src/TableRow'
 import { Text } from '../../typography'
 
 const noop = () => {}
 
+const pseudoSelectors = {
+  ...rowPseudos,
```
akleiner2, line 14: Sigh - would love to decouple the two here, but I guess out of scope for the PR

src/menu/src/MenuItem.js
```diff
@@ -20,34 +25,27 @@ const MenuItem = memo(
...
-          event.preventDefault()
-        }
-
-        safeInvoke(onKeyPress, event)
-      },
-      [onSelect, onKeyPress]
-    )
+    // Pass all props, so the hook can handled `disabled`, `onKeyDown`, `tabIndex`
+    // and any other explicit props that are passed through to the underlying component
+    const { onKeyDown, tabIndex } = useClickable(props)
```
akleiner2, line 48: This doesn't consume `onSelect` though?

mshwery: Nope. Why would it need to?

src/menu/src/MenuItem.js
```diff
@@ -116,6 +99,8 @@ MenuItem.propTypes = {
    */
   is: Box.propTypes.is,
 
+  className: PropTypes.string,
```
akleiner2, line 102: Can you leave the same comment for this prop discouraging use of `className`? We use this throughout the codebase, and is something that I'd love to keep pushing for.

src/menu/src/MenuItem.js
```diff
@@ -146,11 +131,6 @@ MenuItem.propTypes = {
    */
   intent: PropTypes.oneOf(['none', 'success', 'warning', 'danger']),
 
-  /**
-   * Callback to invoke onkeypress
-   */
-  onKeyPress: PropTypes.func,
```
akleiner2, line 152: What about turning this into` onKeyDown`?

src/menu/src/MenuItem.js
mshwery: Yeah not a fan that it was imported to begin with. We can duplicate it.

src/menu/src/MenuOption.js
```diff
@@ -25,15 +25,7 @@ const MenuOption = memo(function MenuOption(props) {
 
   const handleClick = useCallback(e => onSelect(e), [onSelect])
 
-  const handleKeyPress = useCallback(
-    e => {
-      if (e.key === 'Enter' || e.key === ' ') {
-        onSelect(e)
```
akleiner2, line 31: Same comment / question re: `onSelect` missing in the `onKeydown`

src/menu/src/MenuItem.js
mshwery: Can add the description!

src/menu/src/MenuItem.js
mshwery: It doesnt need to be included because box props already include it.

src/menu/src/MenuOption.js
mshwery: `onSelect` isn't needed. That's the whole point of `use-clickable` – you dont have to manually duplicate "click" functionality. It enables other elements (aside from ones that support it OOTB like `button`) to trigger click events for keyboard clicks (Enter/Space). This behaves the same as how native buttons work.

src/menu/src/MenuItem.js
akleiner2: Didn't know that!

mshwery: Rather – the box _types_ include it. Personally don't think we need to include these sorts of pass-through props in PropTypes.

PR #1042, opened 2020-09-16
Adds a `defaultDeprecatedTheme` to ease the majority migration case

akleiner2:
**Overview** 
This PR adds a `defaultDeprecatedTheme` export to Evergreen to ease the migration path for us internally (as well as hopefully other folks externally). Right now, it just composes the fills, palette, scales, and colors - as this is the majority of the use case that we believe. 

It also removes the `themer/` directory, and cleans up some of the deprecated theme stuff in `themes/default` (we can't remove `tokens` _just_ yet. 

**Documentation**


src/themes/deprecated/index.js
```diff
@@ -0,0 +1,11 @@
+import colors from './foundational-styles/colors'
+import fills from './foundational-styles/fills'
+import palette from './foundational-styles/palette'
+import scales from './foundational-styles/scales'
+
+export default {
```
mshwery, line 6: What about all the typography stuff (that we deleted)?

src/themes/deprecated/foundational-styles/index.js
```diff
@@ -0,0 +1,5 @@
+export { default as colors } from './colors'
+export { default as fills } from './fills'
+export { default as palette } from './palette'
+export { default as tokens } from './tokens'
```
mshwery, line 4: I don't think `tokens` was a thing previously.

src/themes/default/index.js
```diff
@@ -1,16 +1,9 @@
...
-} from './deprecated/foundational-styles'
+import { tokens as ogTokens } from './deprecated/foundational-styles'
 
 import tokens from './tokens'
 
 export default {
-  /* DEPRECATED */
-  palette,
-  scales,
   tokens: ogTokens,
```
mshwery, line 7: Wondering if we can move off `tokens` throughout this theme since they are all defined at the root level of the theme object?

akleiner2: There's one last component that uses `theme.tokens.{}`, so will wait for that to land before this.

src/themes/deprecated/foundational-styles/index.js
akleiner2: This isn't even a valid file on disk 😅 , but build wasn't complaining because we don't re-import it anywhere

colinlohner (approved):

PR #1043, opened 2020-09-17
Start adding strongly typed props for building theme objects

akleiner2:
**Overview**
The types look confusing, but the idea is as follows: 

- Most of the juice happens in `BaseThemeObject<T>` - this basically composes `sizes`, `apperances`, and `pseudoSelectors` (where defined) per component. 
- For each of the `baseStyles` - they look to see if the key is a pseudoselector, in which case we don't support double-nesting pseudo-selectors, and also provide a theme function lookup with bound props for the props object. 

This should serve as a start for adding more component typedefs! 

**Documentation**


mshwery (approved):
Seems pretty solid!

colinlohner (approved):

colinlohner:
Looks great!

PR #1045, opened 2020-09-17
Add types for `deprecatedDefaultTheme` that was introduced in #1042

akleiner2:
**Overview**
Title - nothing super crazy, just backporting types!

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -305,6 +305,17 @@ export interface Theme {
 }
 
 export const defaultTheme: Theme
+export const classicTheme: Theme
+
+interface DeprecatedDefaultTheme {
+  colors: Colors,
+  scales: ColorScales,
+  typography: Typography,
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -305,6 +305,17 @@ export interface Theme {
 }
 
 export const defaultTheme: Theme
+export const classicTheme: Theme
+
+interface DeprecatedDefaultTheme {
+  colors: Colors,
+  scales: ColorScales,
+  typography: Typography,
```
mshwery, line 313: FYI I had removed stuff from `Typography` (`headings` I think).

mshwery (approved):

index.d.ts
akleiner2: looks like headings and text! will add back.

PR #1046, opened 2020-09-17
Make select menu better

colinlohner:
**Overview**
Making the select menu better

Screen Recording 2020-09-17 at 11 25 53 AM

Via keyboard navigation:
Screen Recording 2020-09-23 at 2 07 54 AM

**Documentation**


src/themes/default/components/option.js
```diff
@@ -0,0 +1,50 @@
+const baseStyle = {
+  outline: 'none',
+  textDecoration: 'none',
+  display: 'flex',
+  position: 'relative',
+  backgroundColor: 'white',
+  borderBottom: (theme) => `1px solid ${theme.colors.border.muted}`,
+
+  '&:before': {
```
mshwery, line 9: Can we introduce `_before` (via pseudoSelectors) instead?

src/themes/classic/components/option.js
```diff
@@ -0,0 +1,49 @@
+const baseStyle = {
+  outline: 'none',
+  textDecoration: 'none',
+  display: 'flex',
+  position: 'relative',
+  backgroundColor: 'white',
+  borderBottom: (theme) => `1px solid ${theme.colors.border.muted}`,
+
+  '&:before': {
```
mshwery, line 9: Same here: `_before`

src/select-menu/src/Option.js
```diff
@@ -37,38 +52,47 @@ const Option = memo(
...
+        <Text
           height={height}
-          textProps={textProps}
+          display="flex"
+          alignItems="center"
+          {...textProps}
           paddingX={16}
           borderRight={null}
           flex={1}
-          alignSelf="stretch"
```
mshwery, line 63: Curious how this visually changes the item

src/select-menu/src/Option.js
```diff
@@ -37,38 +52,47 @@ const Option = memo(
...
-        isSelected={isSelected}
+      <Pane
         style={style}
-        display="flex"
-        alignItems="center"
-        borderBottom="muted"
-        {...(isSelected
-          ? { boxShadow: `inset 2px 0 0 ${tokens.selectedOptionColor}` }
-          : {})}
+        className={themedClassName}
```
mshwery, line 74: Given that we previously supported `className` (via `...rest`) we should use `cx(themedClassName, className)` here.

Also note that if a consumer passed in a className right now it would override our themed className!

src/select-menu/src/Option.js
```diff
@@ -37,38 +52,47 @@ const Option = memo(
...
+
     return (
-      <TableRow
-        isSelectable={isSelectable && !disabled}
-        isHighlighted={isHighlighted}
-        onSelect={onSelect}
-        onDeselect={onDeselect}
-        isSelected={isSelected}
+      <Pane
         style={style}
```
mshwery, line 73: Do we still need to support this? If we aren't touching `style` we can let it come through via `...rest`?

src/select-menu/src/Option.js
```diff
@@ -1,13 +1,28 @@
...
 import { useTheme } from '../../theme'
+import { Text } from '../../typography'
+
+const pseudoSelectors = {
+  _focus: ':focus',
+  _hover: ':hover',
+  _active: '&[aria-current="true"]:active, &[data-isselectable="true"]:active',
+  _current: '&[aria-current="true"]',
+  _isSelectable: '&[data-isselectable="true"]'
+}
```
mshwery, line 15: What about `_disabled`?

src/autocomplete/src/AutocompleteItem.js
```diff
@@ -1,6 +1,98 @@
...
+const selectedProps = { color: 'selected' }
+const emptyProps = {}
+
+const pseudoSelectors = {
+  _focus: ':focus',
+  _hover: ':hover',
+  _active: '&[aria-current="true"]:active, &[data-isselectable="true"]:active',
+  _current: '&[aria-current="true"]',
+  _isSelectable: '&[data-isselectable="true"]'
+}
```
mshwery: Would `_disabled` be useful here?

Also, `_isSelectable` actually seems like the less standard pseudo – maybe we can remove it if we respect enabled vs disabled options?

src/autocomplete/src/AutocompleteItem.js
```diff
@@ -1,6 +1,98 @@
...
+    return (
+      <Pane
+        style={style}
+        ref={ref}
+        className={themedClassName}
+        {...listBehaviorProps}
+        {...boxProps}
+        {...rest}
+      >
+        <TextTableCell
```
mshwery: I find it odd that we are using (faux) table elements here.

src/autocomplete/src/AutocompleteItem.js
```diff
@@ -1,6 +1,98 @@
...
+      textProps = selectedProps
+    }
+
+    const { getRef, ...listBehaviorProps } = useListBehavior({
+      isSelectable,
+      isSelected,
+      disabled,
+      onSelect,
+      ref
+    })
```
mshwery: User-provided handlers will overwrite these `listBehaviorProps` based on how they are spread on the component below.

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,72 @@
+import { useState, useCallback } from 'react'
+
+/**
+ * Hook that returns bind props + accessibility props for
+ * "selectable" elements
+ */
+
+const resetStyles = {
+  outline: 'none'
+}
```
mshwery, line 10: If a hook is responsible for setting any styles it should do so via inline styles (and merge them with user-provided styles), IMO.

This works only because it happens to (now) be a Box-supported prop, but that was only recently added. This could be a gotcha easily where a css prop is used that Box does not support.

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    },
+    [onSelect, disabled]
+  )
+
+  return {
+    'aria-selected': isHighlighted,
+    'aria-current': isSelected,
+    'data-isselectable': isSelectable && !disabled,
+    tabIndex: isSelectable && !disabled ? 0 : undefined,
```
mshwery, line 66: If we are returning a tabIndex, it should support the consumer-specific tabIndex

mshwery: Also we can't return `undefined` – this will only work on elements that support `disabled`. If the element does not support `disabled` but is still tabbable, we need `tabIndex={-1}`
```suggestion
    tabIndex: isSelectable && !disabled ? tabIndex : -1,
```

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,72 @@
...
+
+        if (nextItemToFocus && nextItemToFocus.hasAttribute('tabindex')) {
+          nextItemToFocus.focus()
+        }
+
+        if (key === 'Enter' || key === ' ' || key === 'Spacebar') {
+          e.preventDefault()
+          e.currentTarget.click()
+          onSelect(e)
+        }
```
mshwery, line 47: Prefer composing `use-clickable` in some way instead of rewriting this same behavior for the same purpose.

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,72 @@
...
+
+export function useListBehavior({
+  disabled,
+  isHighlighted,
+  isSelectable,
+  isSelected,
+  onSelect,
+  ref: inputRef
+}) {
+  const [ref, setRef] = useState(inputRef)
```
mshwery, line 20: Using state for refs will trigger unnecessary rerenders. Would recommend using `useRef`

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,72 @@
...
+        const rowIndex = tableBodyChildren.indexOf(ref)
+
+        if (key === 'ArrowUp' && rowIndex - 1 >= 0) {
+          nextItemToFocus = tableBodyChildren[rowIndex - 1]
+        } else if (
+          key === 'ArrowDown' &&
+          rowIndex + 1 < tableBodyChildren.length
+        ) {
+          nextItemToFocus = tableBodyChildren[rowIndex + 1]
+        }
```
mshwery, line 37: If we are implementing arrow key navigation we should be using the roving tabindex pattern here – which this is not.

TLDR: only the next focused item should have a `tabIndex` value set, and all other items should have `tabIndex={-1}`.

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Hook that returns bind props + accessibility props for
+ * "selectable" elements
+ */
+
+const resetStyles = {
+  outline: 'none'
+}
+
+export function useListBehavior({
```
mshwery, line 12: IMO this should be a hook that is responsible for composing/combining multiple hooks that are responsible for different behaviors:
- making elements keyboard clickable
- implementing roving tabindex pattern
- implementing escape to close pattern
- implementing external click to close pattern (maybe this isn't included here)

src/hooks/use-list-behavior.js
akleiner2: +1 to this - was mostly experimenting here.

src/select-menu/src/Option.js
colinlohner: seemed to not actually affect things, which is why i removed it

src/themes/classic/components/option.js
```diff
@@ -0,0 +1,53 @@
...
+
+  _focus: {
+    backgroundColor: 'colors.neutral.lightest'
+  },
+
+  _active: {
+    backgroundColor: 'colors.blue.lightest',
+  },
+
+  _current: {
```
mshwery, line 38: Should this be `_selected` since it is the selected one?

src/roving-tabindex/tabindex-provider.js
```diff
@@ -0,0 +1,177 @@
+import React, { createContext, useReducer, useMemo, useEffect } from 'react'
+import PropTypes from 'prop-types'
+
+export const ActionTypes = {
+  REGISTER: 'REGISTER',
+  UNREGISTER: 'UNREGISTER',
+  TAB_TO_PREVIOUS: 'TAB_TO_PREVIOUS',
+  TAB_TO_NEXT: 'TAB_TO_NEXT',
```
mshwery, line 8: This is a bit misleading since roving tabindex doesnt use _tabs_ but rather uses arrow keys?

src/roving-tabindex/use-roving-tabindex.js
```diff
@@ -0,0 +1,105 @@
...
+  Previous: 'Previous'
+}
+
+export default function useRovingTabIndex({
+  disabled,
+  isSelectable,
+  isSelected,
+  ref = null
+}) {
+  const tabIndexId = useRef(uniqueId('eg-roving-tabindex_'))
```
mshwery, line 22: Why not https://github.com/segmentio/evergreen/blob/master/src/hooks/use-id.js

src/roving-tabindex/use-roving-tabindex.js
```diff
@@ -0,0 +1,105 @@
...
+    },
+    [context.state]
+  )
+
+  const handleClick = useCallback(() => {
+    context.dispatch({
+      type: ActionTypes.CLICKED,
+      payload: { id: tabIndexId.current }
+    })
+  }, [ tabIndexId ])
```
mshwery, line 91: `tabIndexId` is a ref object so it will never change.

src/roving-tabindex/use-roving-tabindex.js
```diff
@@ -0,0 +1,105 @@
...
+  const tabIndex = isFocused ? 0 : -1
+
+  return { 
+    tabIndex,
+    'aria-selected': isFocused,
+    'aria-current': isSelected,
+    'aria-checked': isSelected,
+    'data-isselectable': isSelectable,
+    onKeyDown: handleKeyDown,
+    onClick: handleClick
```
mshwery, line 103: Curious why not make this massively simpler (and faster) by binding `onKeyDown` + `onClick` on the parent component (Tablist, Menu, etc)?

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,55 @@
...
+    isSelectable,
+    isSelected,
+    ref
+  })
+
+  const { onKeyDown: useClickableOnKeyDown, tabIndex: useClickableTabIndex } = useClickable({
+    disabled,
+    ref,
+    tabIndex: rovingTabIndex
+  })
```
mshwery, line 25: Use clickable accepts `onKeyDown` and will invoke it. That would simplify things a little here because you could remove `keyDownHandler`

mshwery:
Generally would prefer these types of PRs to be separated into more than 1 since there are multiple things happening here:
- useStyleConfig on new components
- a new bind props hook 
- changing component internal structure

src/roving-tabindex/tabindex-provider.js
colinlohner: that makes sense. I'll update to `MOVE_TO`

src/roving-tabindex/use-roving-tabindex.js
colinlohner: Forgot we had that, that's a great improvment

src/roving-tabindex/use-roving-tabindex.js
colinlohner: Good point

src/roving-tabindex/use-roving-tabindex.js
colinlohner: good call out

colinlohner:
Gonna make these tweaks, as well as break this PR up a little bit more

colinlohner:
Braking this up

phoet:
has there been a follow-up to this? was just looking and trying to find out if autocomplete or combo-box can be used with up/down/enter keys. it's an important feature for our usecase.

PR #1047, opened 2020-09-17
Add back <SidebarTab /> with a deprecation warning in favor for <Tab />`

akleiner2:
**Overview**
This PR adds back the `<SidebarTab />` export, but adds a deprecation warning to it, and marks the corresponding types as deprecated. This should now just call to `<Tab direction="vertical" />`, we can make some more updates to this component / its terminology, but this unblocks stuff for now. 

**Documentation**


colinlohner (approved):

index.d.ts
```diff
@@ -1615,6 +1615,10 @@ export declare const SegmentedControl: BoxComponent<
   'div'
 >
 
+/** @deprecated This component will be removed in the next major version of Evergreen */
```
mshwery, line 1618: You'll need this above each export

mshwery (approved):

index.d.ts
akleiner2: nice spot - fixing

PR #1058, opened 2020-09-24
fix(button): merge props value

wangcch:
**Overview**

https://codesandbox.io/s/evergreen-ui-button-yhwd8?file=/src/App.js

Merge props value (boxProps and restProps).
Initialize the undefined value in `restProps` according to `boxProps`.

more:
I suspect that all components that use `resetProps` after `boxProps` have problems🤔

Screen Shot 2020-09-24 at 2 34 38 PM

**Documentation**

### DIFF
--- a/src/buttons/src/Button.js
+++ b/src/buttons/src/Button.js
@@ -1,5 +1,6 @@
 import React, { memo, forwardRef } from 'react'
 import cx from 'classnames'
+import merge from 'lodash.merge'
 import PropTypes from 'prop-types'
 import Box, { spacing, dimensions, position, layout } from 'ui-box'
 import { useStyleConfig } from '../../hooks'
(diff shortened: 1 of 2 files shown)


akleiner2:
Hey there - thanks for the PR! There are a few things to unpack here: 

1. In the FilePicker example, we should be setting a default `height` or `size` on the `FilePicker` component and passing that down to the underlying `<TextInput />` and `<Button />` 

2. `<Button height={undefined} />` correctly unsets `height` - and this is expected behavior; we don't want to merge in the theme's provided `height` with an explicit unsetting of `height`. The bigger issue though is that we also set a `lineHeight` on the `<Button />`, and that's what's giving the appearance of the button having a bigger height than it should be. This is a pretty legit issue I'll look into some time later today. 

Going to close this out for now - but let us know if you have any other questions!

wangcch:
Hi @akleiner2 

Thank you for your reply.

I think `<Button height={undefined} />` and `<Button height={null} />` are different.
And we can get the default value from `boxProps`.

```js
console.log(lodash.merge({}, { a: 1 }, { a: undefined })); // {a: 1}
console.log(lodash.merge({}, { a: 1 }, { a: null }));      // {a: null}
```

PR #1060, opened 2020-09-24
Fix button prop type errors for classic theme 

akleiner2:
**Overview**
We were getting a bunch of prop type errors in the console for buttons - this should be fixed in `ui-box`, but this is a workaround for now. 

Before: 
image

After: 
image

**Documentation**

### DIFF
--- a/src/themes/classic/components/button.js
+++ b/src/themes/classic/components/button.js
@@ -7,7 +7,7 @@ import { defaultControlStyles } from '../deprecated/shared'
 const baseStyle = {
   fontFamily: 'fontFamilies.ui',
   borderRadius: 'radii.1',
-  border: 0,
+  border: '0',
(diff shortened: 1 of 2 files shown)


src/themes/classic/components/button.js
```diff
@@ -7,7 +7,7 @@ import { defaultControlStyles } from '../deprecated/shared'
 const baseStyle = {
   fontFamily: 'fontFamilies.ui',
   borderRadius: 'radii.1',
-  border: 0,
+  border: '0',
```
mshwery, line 10: Hm... Actually yeah this is still invalid. Could be 'none'

akleiner2: Same effect as `border: 'none'` though? One sets border style, the other sets border width (effectively) - point is to show _no_ border, and `Box`s prop types don't know how to handle `border={number}`

Issue #858, opened 2020-06-13
Depedency error: The required JS dependency "dom-helpers/util/inDOM" is not available, it was required by "node_modules/evergreen-ui/commonjs/portal/src/Portal.js".


beders:
**Describe the bug**
Can't compile evergreen-ui.

I'm getting the following error:
The required JS dependency "dom-helpers/util/inDOM" is not available, it was required by "node_modules/evergreen-ui/commonjs/portal/src/Portal.js".

**To Reproduce**
`$ npm install --save evergreen-ui `

**Expected behavior**
All dependencies work

**Evergreen Version**
4.27.4

DOM-helpers version: 5.1.4
(so vastly newer than the one evergreen needs)
(and indeed:
`$ ls: node_modules/dom-helpers/util: No such file or directory`
)
also: in Portal.js:
```
var _inDOM = _interopRequireDefault(require("dom-helpers/util/inDOM"));
```

**Environment (please complete the following information):**
 - OS: MacOS catalina
 - Browser Firefox
 - Version latest

**Additional context**
Error created by the Shadow-cljs compiler (this is a ClojureScript project).

mshwery:
This is probably happening because you have a conflict in your resolved version of `dom-helpers`. You can see that Evergreen relies on `dom-helpers@^3.2.1` per the package.json and yarn.lock files.

mshwery:
This seems specific to the way the project is being compiled/bundled and we have not been able to reproduce it ourselves. Closing for now!

tggreene:
Hi @mshwery, I believe this is a result of the fact that evergreen-ui's [dependency on react-transition-group "4.4.1"](https://github.com/segmentio/evergreen/blob/15d8b463ae4da6eef700e6d97729c4c614055209/package.json#L70) [has a dependency on dom-helpers "5.0.1"](https://github.com/reactjs/react-transition-group/blob/399f26998a5cf2449798a02c755e5808e9b39ddd/package.json#L61) whilst evergreen-ui also [has a dependency on dom-helpers "3.2.1"](https://github.com/segmentio/evergreen/blob/15d8b463ae4da6eef700e6d97729c4c614055209/package.json#L61).

[...]

PR #1066, opened 2020-09-25
Remove dom-helpers, add local lib/canUseDom

tggreene:
This eliminates a dependency on an older version of `dom-helpers` (`3.2.1`) which conflicts with that required by `react-transition-group` (`dom-helpers` `5.0.1`). This addresses an issue using evergreen with shadow-cljs which disallows multiple versions of the same package #858 (I realise it's closed currently, there's a little more detail in my comment [here](https://github.com/segmentio/evergreen/issues/858#issuecomment-698800374)).

I ran tests with `ava` successfully but did trip over some issues with `xo` but I believe they're not introduced by my changes.

There are no ui changes with this and the portal appears to work perfectly well with the updated changes.

I wasn't sure if I should update `examples/ssr-next` or not, I had a bit more difficulty running that with something like an incompatible node version. I can dig further into this if it's helpful and update the yarn.lock there too!

Thanks!

### DIFF
--- a/src/lib/canUseDom.js
+++ b/src/lib/canUseDom.js
@@ -1,0 +1,1 @@
+export default Boolean(typeof window !== 'undefined' && window.document)
--- a/src/portal/src/Portal.js
+++ b/src/portal/src/Portal.js
@@ -1,7 +1,7 @@
 import { Component } from 'react'
-import canUseDom from 'dom-helpers/util/inDOM'
 import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
+import canUseDom from '../../lib/canUseDom'
 
 let portalContainer
 
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "@types/react-transition-group": "^4.4.0",
     "arrify": "^1.0.1",
     "classnames": "^2.2.6",
-    "dom-helpers": "^3.2.1",
     "downshift": "^5.2.0",
     "fuzzaldrin-plus": "^0.6.0",
     "glamor": "^2.20.40",
--- a/src/ssr/src/autoHydrate.js
+++ b/src/ssr/src/autoHydrate.js
@@ -1,5 +1,6 @@
 import { hydrate as boxHydrate } from 'ui-box'
 import { rehydrate } from 'glamor'
+import canUseDom from '../../lib/canUseDom'
 
 /**
  * You shouldn't have to manually run this.
@@ -16,7 +17,7 @@ export function hydrate(hydration) {
 }
 
 export default function autoHydrate() {
-  if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+  if (canUseDom) {
     const hydration = document.querySelector('#evergreen-hydrate')
 
     if (hydration) {
--- a/src/toaster/src/Toaster.js
+++ b/src/toaster/src/Toaster.js
@@ -1,9 +1,7 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
+import canUseDom from '../../lib/canUseDom'
 import ToastManager from './ToastManager'
-
-const isBrowser =
-  typeof window !== 'undefined' && typeof window.document !== 'undefined'
 
 /**
  * The Toaster manages the interactions between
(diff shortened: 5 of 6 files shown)


src/portal/src/Portal.js
```diff
@@ -1,5 +1,5 @@
 import { Component } from 'react'
-import canUseDom from 'dom-helpers/util/inDOM'
+import canUseDom from 'dom-helpers'
```
mshwery, line 2: Does this need to be `import { canUseDOM } from 'dom-helpers'`?

Honestly it seems silly to import this lib just for checking if the window object is not undefined, especially since we have that check in a couple other places too, internally.

tggreene: Hi Matt,

Thanks for looking at this, I'm not a javascript guy by trade so forgive me if I'm a little slow, `dom-helpers` [uses export default](https://github.com/react-bootstrap/dom-helpers/blob/995f5aff37f7e7baedc75f530d8ceb11c5ff9893/src/canUseDOM.ts#L1) for each module hence the default import syntax.

I'm easy on the direction you'd prefer here; it looks like `canUseDom` is potentially being more cautious by checking for `createElement` and additionally constrains the check to one place.

```javascript
export default !!(
  typeof window !== 'undefined' &&
  window.document &&
  window.document.createElement
)
```

Regardless, I'll happily follow your lead. I'll make other usages consistent if that's suitable as well.

mshwery: The problem with the latest version is that the default export of the library is not `canUseDOM` but rather a suite of functions:
https://github.com/react-bootstrap/dom-helpers/blob/995f5aff37f7e7baedc75f530d8ceb11c5ff9893/src/index.ts#L69-L101

tggreene: Aha, yes I've definitely got the wrong end of at least one stick.

However, this syntax `import { canUseDom } from 'dom-helpers'` yields this error on build:

     Error: 'canUseDom' is not exported by node_modules/dom-helpers/esm/index.js, imported by src/portal/src/Portal.js

Although this is fine `import canUseDom from 'dom-helpers/canUseDom'`.

mshwery: Aha, yes, it seems it is because it's not actually being exported by the library directly! 😅 we'll have to reference the path like you've done in your last comment: `import canUseDom from 'dom-helpers/canUseDom'` 

I'm also not opposed to doing:
```sh
yarn remove dom-helpers
```

And then instead of pulling in this whole lib just for a simple function we can have this module internally:
```js
export default function canUseDom() {
  return typeof window !== 'undefined' && window.document
}
```
Then anywhere we do `typeof window !== 'undefined'` or `typeof document !== 'undefined'` we can replace it with this function call. Wdyt?

tggreene: That sounds good to me, I'll do it asap, thanks Matt!

docs/src/components/TopBar.js
```diff
@@ -6,7 +6,7 @@ import GitHubButton from 'react-github-button'
 import GitHubIcon from './GitHubIcon'
 import SpectrumIcon from './SpectrumIcon'
 import LogoWordmark from './LogoWordmark'
-// eslint-disable-next-line import/no-unassigned-import
+// eslint-disable-next-line import/no-unassigned-import, import/extensions
```
tggreene, line 9: welp, look at me thinking I'm being clever

tggreene:
I think this is ready for a recheck.

I think I bumped into a cached dependency difference with xo with respect to the above failure! I think when the dependency cache is killed this guy will flare up 👉 https://github.com/xojs/xo/commit/d3abdb68bbe223f642952aaedf8be9b12fa7fcf3.

This change turned out to be quite noddy in the end, but appreciate your consideration all the same!

src/lib/canUseDom.js
```diff
@@ -0,0 +1,3 @@
+export default function canUseDom() {
+  return typeof window !== 'undefined' && window.document
+}
```
mshwery, line 3: Ah, I see that all the call-sites use this as a value/variable not a function:
```suggestion
export default !!(typeof window !== 'undefined' && window.document)
```

mshwery (commented):
One final change and this looks good to me!! Thank you 🙌

src/lib/canUseDom.js
tggreene: Sorry yes, I should have picked up on that 🤦, I took the note from [the linter](https://eslint.org/docs/rules/no-implicit-coercion) to use Boolean here too ¯\_(ツ)_/¯

mshwery (approved):
Amazing! Thanks

PR #1070, opened 2020-09-29
Update option to use style config

colinlohner:
**Overview**
Updated `Option` to use style config

Screen Recording 2020-09-29 at 12 03 40 PM

**Documentation**

### DIFF
--- a/src/select-menu/src/Option.js
+++ b/src/select-menu/src/Option.js
@@ -1,5 +1,4 @@
 import React, { memo, forwardRef } from 'react'
-import cx from 'classnames'
 import PropTypes from 'prop-types'
 import { useStyleConfig } from '../../hooks'
 import { Image } from '../../image'
@@ -27,7 +26,6 @@ const emptyObject = {}
 const Option = memo(
   forwardRef(function Option(props, ref) {
     const {
-      className,
       disabled,
       height,
       icon,
@@ -50,7 +48,7 @@ const Option = memo(
 
     return (
       <TableRow
-        className={cx(themedClassName, className)}
+        className={themedClassName}
         isSelectable={isSelectable && !disabled}
         isHighlighted={isHighlighted}
         onSelect={onSelect}
@@ -79,7 +77,6 @@ const Option = memo(
 )
 
 Option.propTypes = {
-  className: PropTypes.string,
   label: PropTypes.string,
   icon: PropTypes.string,
   style: PropTypes.any,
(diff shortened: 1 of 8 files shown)


src/themes/default/components/option.js
```diff
@@ -0,0 +1,60 @@
...
+  },
+
+  _active: {
+    backgroundColor: 'intents.info.background',
+  },
+
+  _selected: {
+    backgroundColor: 'intents.info.background',
+
+    ' span': {
```
mshwery, line 42: Should this be `& span`? IIRC it will work in glamor, but think we should be consistent.

mshwery: Also not a huge fan that we have knowledge of the component architecture in the theme.

src/select-menu/src/Option.js
```diff
@@ -22,45 +41,31 @@ const Option = memo(
...
-    }
+    const { className: themedClassName, ...boxProps } = useStyleConfig(
+      'Option',
+      emptyObject,
+      pseudoSelectors,
+      internalStyles
+    )
 
     return (
       <TableRow
```
akleiner2, line 52: Is it worth breaking apart from `<TableRow />` or is that going to come in a follow-up PR?

src/select-menu/src/Option.js
```diff
@@ -1,14 +1,33 @@
...
+  alignItems: 'center',
+  display: 'flex'
+}
+
+const emptyObject = {}
 
 const Option = memo(
   forwardRef(function Option(props, ref) {
     const {
+      className,
```
akleiner2, line 30: Do we need to add the additional classname here? I don't think we export `<Option />` in EG, so consumers can't style this initially.

akleiner2 (approved):
Non-blocking comments to consider!

akleiner2:
Also - unsure why Chromatic is failing 🤔

mshwery (approved):

src/select-menu/src/Option.js
colinlohner: In a follow up

src/themes/default/components/option.js
colinlohner: Agreed, we will change this in the follow up pr

src/select-menu/src/Option.js
colinlohner: that makes sense, i will remove this

PR #1072, opened 2020-09-30
use-list-behavior, use-roving-tabindex, tabindex-provider

colinlohner:
**Overview**
Introduce some new hooks and a provider to allow for list behavior to be used across components

Screen Recording 2020-09-30 at 1 31 47 PM

**Documentation**


src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,50 @@
+import { useCallback } from 'react'
+import safeInvoke from '../lib/safe-invoke'
+import { useRovingTabindex } from '../roving-tabindex'
+import { useClickable } from './use-clickable'
+
+export function useListBehavior({
+  disabled,
```
akleiner2, line 7: What happens here if `disabled` is undefined? Should any of these props have default values?

src/hooks/use-list-behavior.js
```diff
@@ -0,0 +1,50 @@
...
+  const clickHandler = useCallback(
+    event => {
+      safeInvoke(rovingTabIndexOnClick, event)
+      safeInvoke(onSelect, event)
+    },
+    [ onSelect, rovingTabIndexOnClick ]
+  )
+
+  const checkboxProps = {
+    role: isMultiSelect ? 'checkbox' : 'radio',
```
akleiner2, line 37: For my own edification - the aria spec suggests if a list or combobox implements multiselect, then each item should have a role of checkbox? Otherwise radio?

src/select-menu/src/Option.js
```diff
@@ -46,44 +46,54 @@ const Option = memo(
...
+        {...listBehaviorProps}
+        ref={callbackRef}
       >
-        <TextTableCell
+        <Box
+          is="span"
+          height={height}
+          display="flex"
+          alignItems="center"
+          paddingX={16}
```
akleiner2, line 74: Is this fixed padding something we can move into the theme?

akleiner2 (approved):
Seems good from a glance! We should really test behaviors like these, so I encourage thinking about a testing strategy for stuff going forward 👍 . 

Would love to prevent any regressions in the behaviors.

PR #1075, opened 2020-10-01
Smattering of bug fixes across Menus, SelectMenu, and FormField

akleiner2:
**Overview**
Fixes some blocking issues around errors being thrown. Specifically: 

- Fix `useClickable` to have a default assignment to an empty object. Otherwise, destructuring here is broken, and an error throws.
- In `SelectMenu`: fixed the ability to select an already selected name in non-multiselect mode. 
- In `SelectMenu` unmount the react virtual list if filtered options are down to zero. This is behavior that is true for `Autocomplete`, as well. 
- Fix reference to hard-coded text color, for `FormFieldErrorMessage`, and threaded through `themedColor` for typography stuff. 
- Also strung together some hackery for the default theme to get back button focus states for primary buttons. Not super happy about it, but 🤷‍♂️ 

https://www.loom.com/share/15339dd473c849d389fbc685b5aefaae

**Documentation**

### DIFF
--- a/src/select-menu/src/OptionsList.js
+++ b/src/select-menu/src/OptionsList.js
@@ -237,32 +237,34 @@ const OptionsList = memo(function OptionsList(props) {
         </TableHead>
       )}
       <Pane flex={1}>
-        <VirtualList
-          height={listHeight}
-          width="100%"
-          itemSize={optionSize}
-          itemCount={options.length}
-          overscanCount={20}
-          scrollToAlignment="auto"
-          scrollToIndex={scrollToIndex || undefined}
-          renderItem={({ index, style }) => {
-            const item = options[index]
-            const isItemSelected = isSelected(item)
-            return renderItem({
-              key: item.value,
-              label: item.label,
-              icon: item.icon,
-              style,
-              height: optionSize,
-              onSelect: () => handleSelect(item),
-              onDeselect: () => handleDeselect(item),
-              isSelectable: !isItemSelected || isMultiSelect,
-              isSelected: isItemSelected,
-              disabled: item.disabled,
-              tabIndex: 0
-            })
-          }}
-        />
+        {options.length > 0 && (
+          <VirtualList
+            height={listHeight}
+            width="100%"
+            itemSize={optionSize}
+            itemCount={options.length}
+            overscanCount={20}
+            scrollToAlignment="auto"
+            scrollToIndex={scrollToIndex || undefined}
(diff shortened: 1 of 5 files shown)


src/select-menu/src/OptionsList.js
```diff
@@ -139,7 +139,7 @@ const OptionsList = memo(function OptionsList(props) {
 
   const handleSelect = useCallback(
     item => {
-      if (isSelected(item)) {
+      if (isSelected(item) && isMultiSelect) {
```
akleiner2, line 142: Only de-select an item if we're in multiselect, otherwise this is a no-op, and we should just "select" regardless -- we have an effect hook that checks if select ever goes out of date, and updates the `OptionsList`s internal state.

src/select-menu/src/OptionsList.js
```diff
@@ -237,32 +237,34 @@ const OptionsList = memo(function OptionsList(props) {
...
-              onSelect: () => handleSelect(item),
-              onDeselect: () => handleDeselect(item),
-              isSelectable: !isItemSelected || isMultiSelect,
-              isSelected: isItemSelected,
-              disabled: item.disabled,
-              tabIndex: 0
-            })
-          }}
-        />
+        {options.length > 0 && (
```
akleiner2, line 240: We have this exact line in `Autocomplete`, as well - this fixes weird issues where passing a 0 index to virtual list to try to render 0 items errors out.

src/themes/default/components/button.js
```diff
@@ -56,7 +56,7 @@ const getPrimaryButtonAppearance = (appearance, intent) => {
     },
     _focus: {
       backgroundColor: `colors.${color}500`,
-      boxShadow: theme => `0 0 0 2px ${theme.colors[`${color}100`]}`,
+      boxShadow: `0 0 0 2px ${theme && theme.colors[`${color}100`]}`,
```
akleiner2, line 59: Just...ignore this for now...

src/typography/src/Paragraph.js
```diff
@@ -20,7 +20,8 @@ const Paragraph = memo(
     const { colors, fontFamilies } = theme
 
     const themedFontFamily = fontFamilies[fontFamily] || fontFamily
-    const themedColor = colors[color] || color
+    const themedColor =
```
akleiner2, line 23: Makes paragraph similarly text-color aware.

colinlohner (approved):

PR #1091, opened 2020-10-08
Migrate `<Label />` away from using `<Text />`

colinlohner:
**Overview**
Update `<Label />` to no longer compose a `<Text />` component under the hood. Also shift it to use `<Heading />` styles as those are the correct visual styles for the component.

**Default Theme (v6)**
image

**Classic Theme (v5)**
image

**Documentation**

### DIFF
--- a/src/typography/src/Label.js
+++ b/src/typography/src/Label.js
@@ -16,7 +16,7 @@ const Label = memo(
     } = props
 
     const { className: themedClassName, ...boxProps } = useStyleConfig(
-      'Heading',
+      'Label',
       { size },
       pseudoSelectors,
       internalStyles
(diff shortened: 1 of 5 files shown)


src/typography/src/Label.js
```diff
@@ -1,14 +1,49 @@
...
   forwardRef(function Label(props, ref) {
-    return <Text is="label" fontWeight={500} {...props} ref={ref} />
+    const {
+      className,
+      size = 400,
+      ...restProps
+    } = props
+
+    const { className: themedClassName, ...boxProps } = useStyleConfig(
+      'Heading',
```
akleiner2, line 19: Doesn't this still couple `<heading />` and `<Label />`?

colinlohner: I guess it does stylistically. I can decouple this more

akleiner2 (approved):

PR #1093, opened 2020-10-10
docs: fix theming playground rendering

danielhochman:
**Overview**
Fix https://evergreen.segment.com/get-started/theming, which needs a `noInline` property to render the MDX as written.

Currently displays:
```js
SyntaxError: Unexpected token (1:8)
1 : return (const newTheme = {
```

**Before**
Screenshot before

**After**
Screenshot after


mshwery (approved):
Thanks!

PR #1094, opened 2020-10-11
Add new component <Stack />

cdedreuille:
**Overview**
As discuss this is a first attempt to create a new layout component called Stack to evenly space children elements. It also simplifies the alignement using left, center or right. Perhaps later it might be interesting to look into adding a divider between each children that could be themed in v6. 

Screenshot 2020-10-11 at 00 58 32

**Documentation**

### DIFF
--- a/src/stackComponent/src/Stack.js
+++ b/src/stackComponent/src/Stack.js
@@ -1,0 +1,49 @@
+import React, { memo, forwardRef } from 'react'
+import PropTypes from 'prop-types'
+import { spacing, dimensions, position, layout } from 'ui-box'
+import { Pane } from '../../layers'
+
+const Stack = memo(
+  forwardRef(function Stack(props, ref) {
+    const { className, children, space, align, ...restProps } = props
+
+    let justifyItems = align
+    if (align === 'left') justifyItems = 'start'
+    if (align === 'right') justifyItems = 'end'
+
+    return (
+      <Pane
+        display="grid"
+        rowGap={space}
+        justifyItems={justifyItems}
+        ref={ref}
+        className={className}
+        {...restProps}
+      >
+        {children}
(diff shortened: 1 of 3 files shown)


src/stackComponent/src/Stack.js
```diff
@@ -0,0 +1,49 @@
...
+const Stack = memo(
+  forwardRef(function Stack(props, ref) {
+    const { className, children, space, align, ...restProps } = props
+
+    let justifyItems = align
+    if (align === 'left') justifyItems = 'start'
+    if (align === 'right') justifyItems = 'end'
+
+    return (
+      <Pane
```
colinlohner, line 15: Could we actually swap this `<Pane />` out for a `<Box />` from `ui-box`? We are trying to reduce inter-component dependencies

colinlohner (changes requested):

colinlohner:
👍 Thank you for making this PR!

src/stackComponent/src/Stack.js
```diff
@@ -0,0 +1,49 @@
+import React, { memo, forwardRef } from 'react'
```
mshwery, line 1: Can we move this into the `/layers` directory alongside Pane/Card?

src/stackComponent/src/Stack.js
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from 'prop-types'
+import { spacing, dimensions, position, layout } from 'ui-box'
+import { Pane } from '../../layers'
+
+const Stack = memo(
```
mshwery, line 6: We _could_ introduce a breaking change and export this as `Stack`. I think we'd have to rename our existing zindex `Stack` component as something else – maybe `Layer`?

PR #1104, opened 2020-10-30
Add `Table` style config key, and add v6 table styles

akleiner2:
**Overview**
Tiny updates to make tables a bit prettier by default in v6. 

image

**Documentation**

### DIFF
--- a/src/themes/classic/components/table.js
+++ b/src/themes/classic/components/table.js
@@ -1,0 +1,11 @@
+const baseStyle = {}
+
+const appearances = {}
+
+const sizes = {}
+
+export default {
+  baseStyle,
+  appearances,
+  sizes
+}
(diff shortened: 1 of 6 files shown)


src/themes/classic/components/table.js
```diff
@@ -0,0 +1,11 @@
+const baseStyle = {}
```
mshwery, line 1: Should this include `border: 'default'`?

src/themes/default/components/table.js
```diff
@@ -0,0 +1,14 @@
+const baseStyle = {
+  borderRadius: 'radii.1',
+  border: 'muted'
```
mshwery, line 3: Should this be `'default'`?

mshwery (approved):

PR #1139, opened 2021-01-23
Custom item width combobox

edukure:
**Disclaimer**
I couldn't find a way to do this without modifying the Combobox component, but if there's already a way to do this please let me know! Also, this is my first PR so I'm sorry if I did something wrong.

**Overview**
Added `popoverMinWidth` to Combobox component so the 240 default size can be changed to fit smaller items.

Before: Popover had fixed width of 240
Screenshot from 2021-01-22 22-32-10

After: Popover with custom min width to fit combobox size
Screenshot from 2021-01-22 22-32-18

Comboboxes with smaller texts (like just numbers) don't have much leftover space
Screenshot from 2021-01-22 22-34-44

**Documentation**
- [X] Added `popoverMinWidth` prop to Combobox component
- [X] Added Custom Minimum Width Popover to Storybook stories

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,14 +1,37 @@
 /* tslint:disable:interface-name max-classes-per-file no-empty-interface */
 
 import * as React from 'react'
-import { extractStyles as boxExtractStyles, BoxProps, BoxComponent, PolymorphicBoxProps } from 'ui-box'
+import {
+  extractStyles as boxExtractStyles,
+  BoxProps,
+  BoxComponent,
+  PolymorphicBoxProps
+} from 'ui-box'
 import { StyleAttribute, CSSProperties } from 'glamor'
 import { DownshiftProps } from 'downshift'
-import { TransitionProps, TransitionStatus } from 'react-transition-group/Transition'
+import {
+  TransitionProps,
+  TransitionStatus
+} from 'react-transition-group/Transition'
 
-export { configureSafeHref, BoxProps, BoxOwnProps, BoxComponent, PolymorphicBoxProps, EnhancerProps } from 'ui-box'
+export {
+  configureSafeHref,
+  BoxProps,
+  BoxOwnProps,
+  BoxComponent,
+  PolymorphicBoxProps,
+  EnhancerProps
+} from 'ui-box'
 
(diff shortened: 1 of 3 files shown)


akleiner2:
Hey @edukure thanks for the PR! Instead of introducing a new prop to `Combobox`, you can actually pass through `popoverMinWidth` to the underlying `<Autocomplete />` component via `autocompleteProps`. Check out [this CodeSandbox](https://codesandbox.io/s/cool-easley-osv55?file=/src/App.js). 

If you're using Typescript, I just realized that the types for `autocompleteProps` aren't quite right, so if you want to update the following line to instead be `Partial<AutocompleteProps>`, that'd be 👍  :) 

https://github.com/segmentio/evergreen/blob/master/index.d.ts#L642

edukure:
> Instead of introducing a new prop to Combobox, you can actually pass through popoverMinWidth to the underlying <Autocomplete /> component via autocompleteProps. Check out this CodeSandbox.

Thank you so much for this! I tried doing this but I guess I only used {} instead of {{}}. I feel so dumb right now D:
I updated `autocompleteProps` type to `Partial<AutocompleteProps>` as you asked, although I don't fully understand how that helps.

akleiner2:
@edukure right now -- it's expecting all of the required _and_ optional properties to be passed down to the underlying Autocomplete (Combobox composes Autocomplete, it's a bit confusing). This doesn't make a ton of sense, since we already pass through a smattering of autocomplete props: 

```
<Autocomplete
      items={items}
      selectedItem={selectedItem}
      initialSelectedItem={initialSelectedItem}
      itemToString={itemToString}
      onChange={onChange}
      isFilterDisabled={isOpenedByButton}
      {...autocompleteProps}
      onStateChange={handleStateChange}
    >
```

Without being overly explicit about the interface of "things you can pass down to the underlying autocomplete" - maybe this is something we actually want to do - `Partial<T>` gives us some typesafety, while still requiring all underlying properties to be _optional_ (which is effectively what `autocompleteProps` is doing). Not only do we want the presence of `autocompleteProps` to be optional, but its members to be optional, too.

akleiner2:
Also -- looks like Prettier / autoformatting took over the type definition 😅 . Do you mind reverting / turning it off so it just contains your change? Thanks so much for the iteration on this!

PR #1142, opened 2021-02-08
fix warning: The `value` prop is required for the `<Context.Provider>`

### LINKED ISSUE #1140: Table.Row - Context.Provider missing 'value' prop
**Description**
There seems to be bug in `Table.Row` that `value` prop is missing in `<Context.Provider>` as described in console warning

```js
Warning: The `value` prop is required for the `<Context.Provider>`. Did you misspell it or forget to pass it?
    at TableRow
```
image

**Additional context**
Code sandbox warning example using `evergreen-ui` table `sample` bellow
[Warning example](https://codesandbox.io/s/blazing-wind-4bkv6?file=/src/App.js)

billyct:
**Overview**

with react 17.x, use the `TableRow` component will Warning: The `value` prop is required for the `<Context.Provider>`

Fixes #1140


mifi:
@stale This is not an issue, it’s a pr and it should be merged, not closed

src/table/src/TableRow.js
```diff
@@ -99,7 +99,7 @@ const TableRow = memo(
     const height = rest.height || themeHeight
 
     return (
-      <TableRowProvider height={height}>
+      <TableRowProvider value={height}>
```
mshwery, line 102: Would be good to verify that this doesn't introduce any issues, given that it's been like this for a long time and things "work"

mshwery:
Screen Shot 2021-07-14 at 2 49 37 PM

Looks like it does mess things up a little. I wonder if we can/should get rid of the context entirely since it's been broken forever.

Should look like this:
Screen Shot 2021-07-14 at 2 49 32 PM

mshwery:
The fix is to remove the context altogether since it is inert but causing extra work! https://github.com/segmentio/evergreen/pull/1256

PR #1147, opened 2021-02-17
Allow button to accept color prop

colinlohner:
**Overview**
Enabling the button theme to make use of the color prop as expected

```jsx
<Button
    appearance="minimal"
    color={colors.blue500}
>
    Click Me
</Button>
```
image

### DIFF
--- a/src/themes/default/components/button.js
+++ b/src/themes/default/components/button.js
@@ -30,7 +30,7 @@ const colorKeyForAppearanceOrIntent = (appearance, intent) => {
   }
 }
 
-const colorKeyForIntent = (intent) => {
+const colorKeyForIntent = intent => {
   if (intent === 'danger') {
     return `red500`
   } else if (intent === 'success') {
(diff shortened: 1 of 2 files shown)


drew-thompson (approved):
I'm sure there's a good use-case for this, so why not? 🙂

src/themes/default/components/button.js
```diff
@@ -102,7 +103,8 @@ const appearances = {
   },
   minimal: {
     backgroundColor: 'transparent',
-    color: (theme, props) => theme.colors[colorKeyForIntent(props.intent)],
+    color: (theme, props) =>
+      props.color || theme.colors[colorKeyForIntent(props.intent)],
```
mshwery, line 107: Should the theme be responsible for dictating how prop specificity works? Or should the component let more explicit styles take precedence over theme styles?

mshwery: The alternative would be that the Button component itself would let `props.color` take priority over `themeProps.color`

colinlohner: Honestly, I could see it going either way. You may not want to be able to overwrite the color for an appearance or intent. That being said, I think it would be reasonable to expect that any style/box props put on a component should be respected. So it might just be that we need to update useStyleConfig to respect the expected hierarchy

colinlohner: thoughts @mshwery ?

mshwery: Will defer to @akleiner2 when he returns. Think we can merge this for now!

PR #1151, opened 2021-02-22
Custom item renderer for select menu

colinlohner:
**Overview**

Allow users of the `SelectMenu` component to pass a custom `itemRenderer` and `itemHeight` enabling custom layouts and items in the select menu options

image

```jsx
<SelectMenu
            title="Select name"
            options={optionsWithIcons}
            selected={state.selected}
            onSelect={item => setState({ selected: item.value })}
            itemHeight={56}
            itemRenderer={item => (
              <Pane display="flex" flexDirection="column" width="100%">
                <Pane
                  display="flex"
                  width="100%"
                  justifyContent="space-between"
                  alignItems="center"
                >
                  <Pane display="flex" alignItems="center">
                    <Image src={item.icon} width={24} marginRight={8} />
                    <Text size={400}>{item.label}</Text>
                  </Pane>
                  <Text color="muted">
                    {Math.floor(Math.random() * 100) + 1}%
                  </Text>
                </Pane>
                <Text color="muted" marginLeft={32}>
                  sub text
                </Text>
              </Pane>
            )}
          >
            <Button>Options with custom content</Button>
          </SelectMenu>
```


src/select-menu/src/SelectMenu.js
```diff
@@ -177,7 +181,11 @@ SelectMenu.propTypes = {
   /*
    * When true, menu closes on option selection.
    */
-  closeOnSelect: PropTypes.bool
+  closeOnSelect: PropTypes.bool,
+
+  itemRenderer: PropTypes.func,
```
akleiner2, line 186: Do you mind adding some comments here so that they get picked up by docgen (and show up on the docs site)?

drew-thompson (approved):
Nice! 🌲

src/select-menu/src/SelectMenu.js
colinlohner: yep done

src/select-menu/src/OptionsList.js
```diff
@@ -296,7 +295,7 @@ OptionsList.propTypes = {
   onFilterChange: PropTypes.func,
   hasFilter: PropTypes.bool,
   optionSize: PropTypes.number,
-  renderItem: PropTypes.func,
+  itemRenderer: PropTypes.func,
```
akleiner2, line 298: Worth noting that this is a breaking change (although it doesn't matter given that v6 is a lot of breaking changes). To minimize API churn, we could potentially keep this named `renderItem` -- what do you think? 

Otherwise, would be good to add this entry into the `v6-migration-guide`

akleiner2:
Do you mind also updating the typescript types to account for these API changes?

valerieernst (approved):

src/select-menu/src/OptionsList.js
colinlohner: I don't think this is really a breaking change, as this prop isn't actually exposed anywhere. Could rename to renderItem if you disagree though

akleiner2: We technically do export `OptionsList` - it's just not documented on the docs site 😬 . Let's keep it as `renderItem` to minimize the code churn and paper-cut upgrade path, if that's okay?

akleiner2:
Nice! Definitely liking this implementation the best 🥳 . If you don't mind updating the types for the `renderItem` prop, that'd be great. Then should be good to 🚢

akleiner2 (approved):

PR #1174, opened 2021-04-29
V6 docs homepage

kamebkj:
Add homepage to v6-docs


docs-v2/components/TopNav.tsx
```diff
@@ -30,13 +30,13 @@ const TopNav: React.FC<Props> = () => {
       paddingX={majorScale(5)}
     >
       <Pane display="flex" alignItems="center" width={236}>
-        <Pane
+        <a href="/"><Pane
```
akleiner2, line 33: Use a `Link` here from `next/Link`. It'll get transformed the same way at build time, but good to have the routing abstraction in case we want to static deploy vs. SPA deploy

docs-v2/components/homepage/CoreValueCard.tsx
```diff
@@ -0,0 +1,29 @@
+import { Heading, Icon, majorScale, Pane, Paragraph, useTheme } from "evergreen-ui"
+
+interface Props {
+  icon: typeof Icon,
+  title: string,
+  description: string
+}
+
+const ValueCard = ({icon, title, description}: Props) => {
+  const { colors } = useTheme()
```
akleiner2, line 10: In order to get rid of the type error, you might want to do `const { colors } = useTheme() as any`

docs-v2/components/homepage/CoreValueCard.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <Icon
+        icon={icon}
+        size={majorScale(3)}
+        borderRadius="50%"
+        padding={majorScale(2)}
+        color={colors.blue500}
+        backgroundColor={colors.blue50}
+        marginBottom={majorScale(4)}
+      />
+      <Heading size={600} marginBottom={8}>{title}</Heading>
```
akleiner2, line 23: Try to avoid using hard-coded here! Opt for `majorScale()` (8px) or `minorScale()` (4px). Makes it easy to find + change all at once when we cross the bridge for spacing tokens.

docs-v2/components/homepage/DesignLanguage.tsx
```diff
@@ -0,0 +1,46 @@
+import { Pane, majorScale, Heading } from "evergreen-ui"
+import React from "react"
+import DesignLanguageCard from "./DesignLanguageCard"
+import IA from "../../utils/IA"
+
+const DesignLanguage = () => {
+  const foundations = IA.foundations.items.slice(0,3)
```
akleiner2, line 7: Nice!!

docs-v2/components/homepage/DesignLanguage.tsx
```diff
@@ -0,0 +1,46 @@
+import { Pane, majorScale, Heading } from "evergreen-ui"
+import React from "react"
```
akleiner2, line 2: Small nit - prefer `React` to be the top-most import.

docs-v2/components/homepage/DesignLanguageCard.tsx
```diff
@@ -0,0 +1,54 @@
+import { Heading, Link, majorScale, Pane, Paragraph } from "evergreen-ui"
+import React from "react"
+import Thumbnail from "../Thumbnail"
+import { Item } from "../../utils/IA"
+
+interface Props {
+  img: string,
+  title: string,
+  description: string,
+  linkText?: string,
```
akleiner2, line 10: Instead of passing in `linkText` and `link`, what do you think about passing in `link: JSX.Element`, so the parent component can decide what the `<Link />` should look like?

docs-v2/components/homepage/GetStartedCard.tsx
```diff
@@ -0,0 +1,35 @@
+import { Heading, Link, majorScale, Pane, Paragraph, useTheme } from "evergreen-ui"
+
+interface Props {
+  title: string,
```
akleiner2, line 4: Seems like each of these cards pull the same Props interface? Might be worth exporting a common set of fields in a `types.ts` file, and then importing them in each of these cards. 

So you could do something like `interface Props extends BaseProps {}`

docs-v2/components/homepage/GetStartedCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const GetStartedCard = ({title, description, linkText, link, img}: Props) => {
+  const { colors } = useTheme()
+
+  return (
+    <Pane
+      padding={majorScale(5)}
+      paddingRight={majorScale(3)}
+      backgroundColor={colors.gray50}
+      borderRadius={8}
```
akleiner2, line 19: Comments re: `majorScale` here :)

docs-v2/pages/index.tsx
```diff
@@ -1,15 +1,22 @@
...
-      <Link href="/about">
-        <a>About</a>
-      </Link>
-    </p>
-  </Layout>
-)
+const IndexPage = () => {
+  return (
+    <Layout title="Evergreen">
+      <HomeHero />
```
akleiner2, line 12: Wooooo-hooooo

docs-v2/pages/foundations/index.tsx
```diff
@@ -30,7 +30,7 @@ const FoundationsPage: React.FC<Props> = () => {
           flexDirection="column"
           justifyContent="flex-start"
           padding={majorScale(5)}
-          maxWidth={1200}
+          maxWidth={1024}
```
akleiner2, line 33: Not related to this PR at all, but let's move these shared UIs out to templated components. Happy to do this in a follow-up.

docs-v2/components/homepage/Resources.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const Resources = () => {
+  const { colors } = useTheme()
+
+  return (
+    <Pane background={colors.gray50}>
+      <Pane
+        paddingY={majorScale(5)}
+        marginX="auto"
+        maxWidth={majorScale(143)}
```
akleiner2, line 13: Let's figure out a consistent maxWidth and put that in a common `layout-constants` file, so pages don't feel jumpy in between.

docs-v2/public/homepage/logo-figma.svg
```diff
@@ -0,0 +1,14 @@
+<svg width="24" height="25" viewBox="0 0 24 25" fill="none" xmlns="http://www.w3.org/2000/svg">
```
akleiner2, line 1: Let's maybe move these into an `icons` folder under `homepage/`? What do you think?

docs-v2/components/homepage/DesignLanguageCard.tsx
kamebkj: Didn't know we can do this. Will change!

I'm using `link` as the routing path for the `Thumbnail` component though. So either we can clean up `Thumbnail`, or I'll need to pass in another prop that is called `type`.

docs-v2/public/homepage/logo-figma.svg
kamebkj: Sounds good! Looks like we have an `icons` folder for all components. Do you recommend we put it there, or a new `icons` folder under `homepage`?

akleiner2: `icons` folder for all components is a good idea - we can re-work these then to be React components themselves. Nicer inter-op with our existing Icon structures too. 


```
const FigmaLogo = ({ size = 24 }: Props) => ( <Pane is="svg" {...} /> )
```

for example

akleiner2 (approved):

PR #1176, opened 2021-05-01
Set value when children are updated

hellooimkat:
**Overview**
Storybook code for the following gifs is found here: https://github.com/segmentio/evergreen/compare/cat-editable-cell...cat-example?expand=1

Previous behavior:
When passing in the key to the parent `Row` component, the behavior is as expected:
prev_key

When not passed in the key to the parent `Row` component, when the value of the cell changed due to conditional rendering, values rendered were not expected:
Prev_no_key

Current:
With key:
Working_key

Without key:
Working_no_key

**Documentation**
None of these changes apply:

### DIFF
--- a/src/table/src/EditableCell.js
+++ b/src/table/src/EditableCell.js
@@ -1,4 +1,4 @@
-import React, { memo, useState } from 'react'
+import React, { memo, useEffect, useState } from 'react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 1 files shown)


akleiner2 (approved):
Looks great! We should be relying on a single source of rendering truth (in this case state), and then having an effect syncing that state with any changes in props.

PR #1203, opened 2021-05-18
Fixup path generation for in-progress pattern components

akleiner2:
**Overview**
We were erroneously relying on the file system to generate this particular path. While for the overview page, we should still count that as a part of path generation, we should skip out the `/props` specifically if the component is in progress.

N/A build should work as expected

**Documentation**


mshwery (approved):

akleiner2:
GH Actions is still down, so going to land anyways (tested this locally)

PR #1227, opened 2021-06-09
Bring back icon search in the docs

akleiner2:
**Overview**
This PR brings back icon searching into the docs site so that users can quickly find what they're looking for. 

image

**Documentation**

### DIFF
--- a/docs/components/IconSearch.tsx
+++ b/docs/components/IconSearch.tsx
@@ -51,7 +51,7 @@ const IconSearch: React.FC<Props> = () => {
 
   return (
     <Pane width="100%" minHeight={250}>
-      <SearchBar query={query} onQueryChange={setQuery} placeholder="Search through icons below:" />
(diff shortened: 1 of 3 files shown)


docs/components/IconSearch.tsx
```diff
@@ -0,0 +1,77 @@
...
+        display="flex"
+        justifyContent="center"
+        alignItems="center"
+        flexDirection="column"
+        cursor="pointer"
+        borderRadius={5}
+        paddingY={majorScale(3)}
+        className={css({
+          ':hover': {
+            opacity: 0.8,
```
kamebkj, line 25: let's not change the opacity and keep it 1 when hover. hover should not dim the target visually

kamebkj (approved):
Wonder if we can add a "InfoSignIcon copied to clipboard" when clicking on the icon. Otherwise left one comment and the rest lgtm.

PR #1238, opened 2021-06-21
New pattern - Empty States

kamebkj:
**Overview**
This PR includes a new pattern component - Empty States, with a few examples and test cases. The documentation for Empty State will be in a separate PR later once this component is in the Evergreen package.

**Screenshots**
- At a high level, an Empty State consists of a logo, some text, and the actions to be taken.
- There are two types of Empty States: `horizontal` (default) and `vertical`, which can be specified with `orientation` prop. 
- Additionally, when using an Empty State in a table, we should use the `light` background, vs the `dark` background when it is not in a table (does not have a border container) and is standalone, by specifying the `background` prop.

Here's a basic example of what a standalone, horizontal Empty State looks like:
image

An example of what it looks like in a table:
image

An example of what it looks like in a smaller viewport such as a popover:
image


src/empty-states/__tests__/EmptyState.test.js
```diff
@@ -0,0 +1,110 @@
...
+
+  it('should render an image and a title', () => {
+    const { getByText } = render(
+      <EmptyState
+        title="My Empty States"
+        image={<LockIcon color={defaultTheme.tokens.colors.gray500} />}
+        imageBgColor={defaultTheme.tokens.colors.gray200}
+      />
+    )
+    expect(getByText('My Empty States')).toBeInTheDocument()
```
akleiner2, line 30: What do you think about updating assertions to use `toBeVisible()`? It also includes `toBeInTheDocument()` as implied.

src/empty-states/__tests__/EmptyState.test.js
```diff
@@ -0,0 +1,110 @@
...
+        <EmptyState
+          title="My Empty States"
+          image={<LockIcon color={defaultTheme.tokens.colors.gray500} />}
+          imageBgColor={defaultTheme.tokens.colors.gray200}
+        />
+      )
+    ).not.toThrowError()
+  })
+
+  it('should render an image and a title', () => {
```
akleiner2, line 22: Let's update this test to check for the presence of the icon / image

src/empty-states/__tests__/EmptyState.test.js
```diff
@@ -0,0 +1,110 @@
...
+        anchorCta={
+          <EmptyState.LinkButton href="https://segment.com/docs/">Link to Documentation</EmptyState.LinkButton>
+        }
+      />
+    )
+    expect(getByRole('link')).toBeInTheDocument()
+  })
+
+  // Small Minimal Example with just image and title
+  // Not sure how to properly test svg
```
akleiner2, line 84: We can get rid of this now :)

src/empty-states/src/EmptyState.js
```diff
@@ -0,0 +1,184 @@
...
+import { useTheme } from '../../theme'
+import { Link, Heading, Paragraph } from '../../typography'
+
+/* eslint-disable react/prop-types */
+
+const HorizontalOrientation = ({
+  anchorCta,
+  background,
+  description,
+  image,
```
akleiner2, line 15: Should we name these props `icon` and `iconBgColor`?

src/empty-states/src/EmptyState.js
```diff
@@ -0,0 +1,184 @@
...
+
+const LinkButton = props => {
+  return <Link {...props} target="_blank" size={300} lineHeight="34px" />
+}
+
+const EmptyState = ({
+  anchorCta,
+  background = 'light',
+  description,
+  image,
```
akleiner2, line 134: One thing that I'm thinking about...do we know / can we guarantee that the icon color and iconbackground color can be compatible? This puts responsibility on consumers to "do the right thing", but I'm just wondering if there's a way we can restrict this API up front

src/empty-states/src/EmptyState.js
```diff
@@ -0,0 +1,184 @@
...
+const PrimaryButton = props => {
+  return <Button appearance="primary" {...props} />
+}
+
+const SecondaryButton = props => {
+  return <Button appearance="minimal" {...props} />
+}
+
+const LinkButton = props => {
+  return <Link {...props} target="_blank" size={300} lineHeight="34px" />
```
akleiner2, line 127: The way this currently behaves - we're assuming links passed in would strictly be external links. Should we remove the explicit `target="_blank"`, and instead leave that up to consumers? I can see folks passing through single-page route changes here... i.e.


```
<EmptyState 
    // ... 
    anchorCta={<EmptyState.LinkButton is={ReactRouterLink} to="/path/in/my-app"}
/>
```

src/empty-states/src/EmptyState.js
kamebkj: Yup good call!

src/empty-states/src/EmptyState.js
kamebkj: Yeah agree that this is putting the responsibility to the consumers. A few ideas:
- Restricted the icon color & background color with intent (success, warn, danger, info) and only allow passing in icon shape (class) and its intent. Not sure if this would be too restricted.
- Restricted to just icon and its color, and we always apply the color minus 200 (red-500 -> red-300) as its background so we get one less prop. This can result in less optimized contrast since our color system was not great.
- Keep as is.

What do you think?

akleiner2: Hm..yeah I say we can keep as-is for now and maybe work to make the API more restrictive in the future.

src/empty-states/src/EmptyState.js
```diff
@@ -0,0 +1,184 @@
...
+import PropTypes from 'prop-types'
+import { Button } from '../../buttons'
+import { Pane } from '../../layers'
+import { majorScale, minorScale } from '../../scales'
+import { useTheme } from '../../theme'
+import { Link, Heading, Paragraph } from '../../typography'
+
+/* eslint-disable react/prop-types */
+
+const HorizontalOrientation = ({
```
akleiner2, line 11: Let's memoize these components to prevent unnecessary re-renders. 

You can do: 

```
import React, { memo } from 'react' 
```

and then `const HoriztontalOrientation = memo(function HorizontalOrientation({ ... }) { return ... })`

src/empty-states/src/EmptyState.js
```diff
@@ -0,0 +1,184 @@
...
+      </Pane>
+    </Pane>
+  )
+}
+
+/* eslint-enable react/prop-types */
+
+/* eslint-disable react/prop-types */
+
+const VerticalOrientation = ({ background, description, icon, iconBgColor, primaryCta, title }) => {
```
akleiner2, line 77: `memo` here.

src/empty-states/src/EmptyState.js
```diff
@@ -0,0 +1,184 @@
...
+
+const SecondaryButton = props => {
+  return <Button appearance="minimal" {...props} />
+}
+
+const LinkButton = props => {
+  return <Link {...props} size={300} lineHeight="34px" />
+}
+
+const EmptyState = ({
```
akleiner2, line 130: memo here :)

akleiner2:
This is looking pretty good! Left a few more small things to touch up. The final thing is to add this component export to the bundle, and add its types to `index.d.ts`. Then, we can add documentation after this lands + we release.

akleiner2 (approved):
Woo!! Nice work - I think this is an excellent first ship :)

Issue #1241, opened 2021-07-02
Select Menu deselect example. Select, deselect, then re-select takes 2 clicks.


dustincjensen:
https://evergreen.segment.com/components/select-menu#multiselect_with_deselect_example

Minor oddity with the Select Menu deselect.
- Open the menu
- select Apricot
- deselect Apricot
- select Apricot again (nothing happens)
- select Apricot again (now selected).

mshwery:
Resolved by #1256

PR #1247, opened 2021-07-07
#1241 - Possible solution for 2-click re-select on multi-select for Select Menu Component

derekvelzy:
**Overview**
The `isSelected(item) && isMultiSelect` condition on line 142 of OptionsList was causing the user to need to click twice to re-select an option on multiselect (per issue #1241). I could not find any example of where this condition was used, but removing it solved this issue.

N/A

**Documentation**


src/select-menu/src/OptionsList.js
```diff
@@ -139,11 +139,12 @@ const OptionsList = memo(function OptionsList(props) {
...
-        onDeselect(item)
-      } else {
-        onSelect(item)
-      }
+      // THE ONDESELECT CONDITION REQUIRES THE USER TO CLICK TWICE TO RE-SELECT ON MULTIPLE
+      // if (isSelected(item) && isMultiSelect) {
+      //   onDeselect(item)
+      // } else {
+      onSelect(item)
+      // }
```
mshwery, line 147: I think we want to keep the conditional in place. It allows multiselects to deselect items by clicking on them again.

What may be the problem is that this callback isn't specifying all of its dependencies: `close` `onSelect` `isSelected`

mshwery: Although, it does look as though onDeselect is still being called under the hood by the underlying TableRow item.

mshwery:
@derekvelzy thanks for submitting this!! I'm going to close it out in favor of the more complete refactor in the other PR, but thank you for the contribution and pointing out a solution :)

PR #1261, opened 2021-07-15
Upgrade deps

mshwery:
**Overview**
This PR updates all deps within their semver ranges, and uses react-transition-group's `nodeRef` to abandon `findDOMNode` (not supported in strict mode). I removed `react-scrollbar-size` because it's not referenced anywhere.

NA

**Documentation**


dlasky (approved):

akleiner2 (approved):
Need to bump size limit (wondering what deps caused the jump), but otherwise LGTM

mshwery:
> Need to bump size limit (wondering what deps caused the jump), but otherwise LGTM

New Blueprintjs icons from that update.

mshwery:
> Need to bump size limit (wondering what deps caused the jump), but otherwise LGTM

btw @akleiner2 it was the blueprintjs icons – CI builds and then runs size-limit, they just werent included in the git commit.

PR #1273, opened 2021-07-19
move @types/* dependencies to devDependencies

krijoh92:
**Overview**
Moved `@types/react` and `@types/react-transition-group` to `devDependencies` instead of `depdendencies`, this way those types shouldn't interfere with typescript builds of projects using evergreen.

**Documentation**

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,8 +56,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "@types/react": "^16.9.5",
-    "@types/react-transition-group": "^4.4.0",
     "arrify": "^1.0.1",
     "classnames": "^2.2.6",
     "downshift": "^5.2.0",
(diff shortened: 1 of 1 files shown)


mshwery (commented):
Hey! Thanks for the PR. Unfortunately these need to be direct deps because they are required in the index.d.ts types. Moving them will actually break TypeScript project builds.

krijoh92:
Ok, we are getting a lot of errors like this:
```
node_modules/evergreen-ui/node_modules/@types/react/index.d.ts:3237:13 - error TS2717: Subsequent property declarations must have the same type.  Property 'stop' must be of type 'SVGProps<SVGStopElement>', but here has type 'SVGProps<SVGStopElement>'.

3237             stop: React.SVGProps<SVGStopElement>;
                 ~~~~

  node_modules/@types/react/index.d.ts:3238:13
    3238             stop: React.SVGProps<SVGStopElement>;
                     ~~~~
    'stop' was also declared here.

```

when building with Evergreen, just assumed it was because of the @types/react being in depdendencies. 

changing `skipLibCheck` from `false` to `true` solved it for us though, so I guess that's the workaround for us.

mshwery:
@krijoh92 what version(s) of evergreen, react, @types/react?

krijoh92:
```
evegreen: 6.1.0
react: 17.0.2
@types/react: 17.0.14
```

That's probably the issue, different versions in react & types for react so there's a different version installed in evergreens node_modules folder in our project

mshwery:
Ah yep! Evergreen isn't ready for React 17 yet either due to some of its deps. Will be soon though!

krijoh92:
Ah ok, missed that part :) But haven't experienced any other issues from using Evergreen with React 17 though :+1:

PR #1274, opened 2021-07-19
Performance tuning

mshwery:
**Overview**
This PR addresses a bunch of small performance gains, especially for components that are used in large numbers (like inside Tables).

There are no breaking or obvious functional changes, but it does fix a few bugs in the stories – like editable cells appearing in the left hand corner of the screen 😬 

NA

**Documentation**


src/autocomplete/src/Autocomplete.js
```diff
@@ -153,13 +155,13 @@ const Autocomplete = memo(
...
           ...restDownshiftProps
         }) => (
-          <div style={{ width: '100%' }}>
+          <div style={containerStyle}>
             <Popover
               bringFocusInside={false}
               isShown={isShown}
               minWidth={popoverMinWidth}
               position={position || (targetWidth < popoverMinWidth ? Position.BOTTOM_LEFT : Position.BOTTOM)}
-              content={() => (
```
mshwery, line 162: Unnecessary new function created every time. Unless we are using the `close` method we can simply pass the JSX directly

src/hooks/use-latest.js
```diff
@@ -1,15 +1,11 @@
...
 /**
  * A React Ref that stores the latest value it is given (updated in useEffect's callback).
+ * @return {{ readonly current: any }}
  */
 export function useLatest(value) {
   const ref = React.useRef(value)
-
-  useIsomorphicLayoutEffect(() => {
-    ref.current = value
-  })
```
mshwery, line 12: This was unnecessary.

src/table/src/EditableCell.js
```diff
@@ -18,82 +19,100 @@ const EditableCell = memo(function EditableCell(props) {
...
 
+  const lessOpacity = useMemo(() => disabled || (!value && placeholder), [disabled, value, placeholder])
+
+  const mergedTextProps = useMemo(
+    () => ({
+      size,
+      opacity: lessOpacity ? 0.5 : 1,
+      ...textProps
+    }),
+    [lessOpacity, size, textProps]
```
mshwery, line 104: Debating on whether we should do `...textProps` but probably not given that this might change on a mounted component? Would it?

src/table/src/EditableCellField.js
```diff
@@ -112,9 +123,21 @@ const EditableCellField = memo(function EditableCellField(props) {
...
+    () => ({
+      left,
+      top,
+      height,
+      minHeight: Math.max(height, minHeight),
+      width,
+      minWidth: Math.max(width, minWidth),
+      zIndex
+    }),
+    [left, top, height, width, minHeight, minWidth, zIndex]
```
mshwery, line 141: I think we may want to consider ripping out some of this measurement/dimension logic in favor of a simple hook. There's a lot of complex logic that has nothing to do with this component in particular.

src/table/src/EditableCellField.js
```diff
@@ -1,44 +1,81 @@
...
+      setTop(calculatedTop)
+
+      // recursively run the updater
+      latestAnimationFrame.current = requestAnimationFrame(() => updater())
+    }
+
+    // kick off the updater
+    updater()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
mshwery, line 74: This could be replaced with a hook – given a ref, we recursively find the dimensions using resize observer. Using resize observer would mean no IE support, but I highly doubt we have great IE support already (and we don't support it in the app).

src/avatar/src/Avatar.js
```diff
@@ -62,6 +64,7 @@ const Avatar = memo(
     )
 
     const [imageHasFailedLoading, setImageHasFailedLoading] = useState(false)
+    const onError = useCallback(() => setImageHasFailedLoading(true), [])
```
dlasky, line 67: this may not be as efficient as one would hope: https://kentcdodds.com/blog/usememo-and-usecallback

src/table/src/EditableCell.js
dlasky: i think even if the wrapped the change in a useMemo the object ref would be changing here given its a single incoming prop (vs say `const Example = ({otherProps, ...textProps}) => {}`

src/table/src/EditableCellField.js
dlasky: plus the perf on it is likely much better. im in favor of resize observer

src/table/src/EditableCellField.js
dlasky: 200% measurement should definitely be a utility hook

dlasky (approved):

src/table/src/EditableCell.js
mshwery: It would, basically if the parent rerenders and they are passing `textProps={{}}` it'll change. but i can't guarantee that it has the same number of keys, therefore it is unsafe to spread as deps because it could change across renders.

src/avatar/src/Avatar.js
mshwery: Yeah for sure, might be too aggressive here. I was considering that there is internal state that will cause rerenders, and so we may as well cache the function, but it might be excessive in this case.

PR #1275, opened 2021-07-20
Add themeable option to colour props for Heading component

### LINKED ISSUE #1242: color property in Heading component not working properly
#### Color names are not working properly in Heading Component, but working fine when using color values or Text Component

```js
// NOT WORKING
<Heading size={600} color="Blue500">
  Lorem Ipsum
</Heading>
```

```html
// WORKING
<Heading size={600} color="#3366FF">
  Lorem Ipsum
</Heading>

<Text color="muted">
  Lorem Ipsum
</Text>
```

kwnath:
**Overview**

- Allows the usage of theme tokens to be used in the colour prop of the Heading component.
- Add a test to test the usage

Fixes #1242 

This is likely going to be needed for other components so maybe good to think about how we do this in the future?

**Testing**

Create a component Heading with a theme token for the colour prop 
```
<Heading color="blue500">Hello World</Heading>
```

**Documentation**


src/typography/src/Heading.js
```diff
@@ -3,27 +3,35 @@ import cx from 'classnames'
...
+    const themedColor = colors[color] || (colors.text && colors.text[color]) || color
+
     return (
       <Box
         is="h2"
         ref={ref}
         className={cx(themedClassName, className)}
         marginTop={0}
         marginBottom={0}
+        color={themedColor}
```
mshwery, line 34: This seems reasonable to me, but note that it will get clobbered by `color` defined in the theme hook for this particular component. That means you can't do instance-specific overrides for color if the component's theme specifies its own.

kwnath: That's true, it's probably also not a good idea to have the heading rely on the theme hook. The original intent was to make it more convenient but if this introduces unexpected behaviour then it's perhaps best not to include this PR.  Or do you perhaps have an alternative approach?

mshwery: I would opt for fixing this in the component-specific theme instead, I think. You can pass `color` to the `useStyleConfig` hook.

Issue #1223, opened 2021-05-31
The menu has no menu items


mifi:
I get this error if I use the `disabled` prop for `Menu.Item` and if all menu items in a `Menu.Group` are disabled. Can we remove this line of code?

https://github.com/segmentio/evergreen/blob/29f9c41ff15527dac0892330b8032ccd585b1412/src/menu/src/Menu.js#L29

Maybe instead return?
I can submit a pr if this is the correct solution

akleiner2:
I'd be fine either removing that line, or just throwing that error in dev -- happy to accept a PR either way.

dudhatdhavalm:
Is this issue still open or not?
if it is still open then I would like to do my first contribution.

mifi:
Still an issue. Feel free to contribute :)

brandongregoryscott:
Published with [v6.7.1](https://github.com/segmentio/evergreen/releases/tag/v6.7.1)

PR #1279, opened 2021-07-27
Issue fix for the menu has no menu items #1223

JeevantheDev:
**Overview**
- Set a useState hook for looking if there are empty menu items or not.
- If there are not any items then it will render  "**No menu items...**" inside the Popover.

**Screenshot**
menu-dropdown

**Documentation**


src/menu/src/Menu.js
```diff
@@ -101,9 +104,19 @@ const Menu = memo(function Menu(props) {
   }, [menuRef])
```
mshwery, line 104: Given that menuRef wont actually trigger this hook to run again when it changed (mutable refs wont trigger renders or hooks to run again), I would opt for us calculating `isEmpty` separately, on every render. We don't need useState for it.

mshwery (commented):
looks good but curious if we could do this without useState or adding the logic to that huge useEffect hook?

src/menu/src/Menu.js
```diff
@@ -26,7 +28,8 @@ const Menu = memo(function Menu(props) {
       : []
 
     if (menuItems.current.length === 0) {
-      throw new Error('The menu has no menu items')
+      setIsEmpty(true)
+      // throw new Error('The menu has no menu items')
```
mshwery, line 32: Can we remove this commented out code?

JeevantheDev: Ya ya sure.

src/menu/src/Menu.js
JeevantheDev: Ok, I got it. Once I do it I will do the PR.

src/menu/src/Menu.js
JeevantheDev: I have committed the code without useState hook. Kindly review it.

brandongregoryscott (approved):

PR #1286, opened 2021-08-07
Empty state fix

kamebkj:
Remove secondary button from Empty State


index.d.ts
```diff
@@ -979,15 +979,12 @@ export interface EmptyStateOwnProps {
   background?: 'light' | 'dark'
   /** the primary cta of the empty state */
   primaryCta?: React.ReactNode
-  /** the secondary cta of the empty state */
-  annchorCta?: React.ReactNode
```
akleiner2, line 983: Oh my! Let's change this `anchorCta`

akleiner2 (approved):
Small comment, otherwise ship!

index.d.ts
kamebkj: oh wow I didn't notice this after reading it for the 5th time. good catch. thanks!

Issue #1234, opened 2021-06-18
Warning in a custom button


msakal20:
Hello, I have a problem with custom button, when I create a new appearance, I have this warning, from it, I can see that the component button takes only 3 values ["default", "minimum", "main"], but appearance work.
![image](https://user-images.githubusercontent.com/84870471/122583968-79934880-d062-11eb-83b1-b0dfdf027f3f.png)
The same warning I see in documentation.
![image](https://user-images.githubusercontent.com/84870471/122584112-a182ac00-d062-11eb-8c21-c4ccc3515315.png)
 
Сan I fix this warning ?

akleiner2:
Annswered over in https://github.com/segmentio/evergreen/discussions/1236

PR #1287, opened 2021-08-08
fix: Update appearance Button propType to stop warnings

CLSnazel:
**Overview**
This branch removes the warning shown in #1236 and #1234. Giving a custom appearance value that is not the preset `default`, `primary` or `minimal` will no longer cause a warning in the browser. This is helpful for those who intend to apply themes of their own and would like to name button appearances without browser warnings.

The Button prop types have been set to `string` instead of `oneOf` the preset values for the default theme.

In storybook, a button appearance of `destructive` is applied to buttons in the `Button types` story. It results in this warning in the console:
destructive-warning
With this branch, warnings like this will not appear.

**Documentation**


src/buttons/src/Button.js
```diff
@@ -139,7 +139,7 @@ Button.propTypes = {
   /**
    * The appearance of the button.
    */
-  appearance: PropTypes.oneOf(['default', 'minimal', 'primary']),
+  appearance: PropTypes.string,
```
brandongregoryscott, line 142: @zkuzmic @kamebkj @akleiner2 This change makes sense to me until we iron out the typed theming support more, similar to the change for `intent` -> `string` in #1341. @CLSnazel Would you mind updating the `ButtonAppearance` type alias in `index.d.ts` to just `string` as well?

```ts
export type ButtonAppearance = string
```

brandongregoryscott (changes requested):

src/buttons/src/Button.js
CLSnazel: Thank you for that pointer! `index.d.ts` has been updated accordingly. Please let me know if there's any additional changes required.

brandongregoryscott (approved):
Thanks for the contribution! 🥳

PR #1290, opened 2021-08-10
Empty state doc

kamebkj:
First draft of the Empty State pattern doc. This includes two new doc-specific components for "Dos and Don'ts" - `RuleCard` and `RuleLayout`. Would love feedback on those as well.


docs/components/MDX/componentMapping.tsx
```diff
@@ -17,12 +19,14 @@ const componentMapping = {
   p: (props: any) => <Paragraph marginBottom={majorScale(3)} {...props} />,
   strong: (props: any) => <Strong {...props} />,
   ol: (props: any) => <Ol {...props} />,
-  ul: (props: any) => <Ul {...props} />,
+  ul: (props: any) => <Ul {...props} marginTop="-16px" />,
```
akleiner2, line 22: Do we want this to be universally true? Can we see a screenshot of what this change looks like?

docs/components/MDX/renderers/RuleCard.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const RuleCard: React.FC<Props> = ({ type, description, image }) => {
+  const color = type === 'Do' ? '#52BD94' : '#D14343'
+  const icon = type === 'Do' ? <TickCircleIcon color={color} /> : <CrossIcon color={color} />
+  return (
+    <Card width="500px" border="default" marginRight={majorScale(5)} padding={majorScale(3)}>
+      {image && (
+        <Pane marginBottom={majorScale(2)}>
+          <img src={image} alt={image} width="100%" />
```
akleiner2, line 17: Should we make `alt` a prop that's passed in here? Like `imgAlt`?

docs/documentation/patterns/empty-states.mdx
```diff
@@ -0,0 +1,217 @@
...
+#### Tone
+
+Tone should be educational and plain-spoken.
+
+<RuleLayout>
+  <RuleCard type="Do" description="Use an educational and plain-spoken tone." image="/empty-state-tone-do.png" />
+  <RuleCard
+    type="Don't"
+    description="Use excessive emotion or exceed 1 line for headline and 2 lines for body."
+    image="/empty-state-tone-dont.png"
```
akleiner2, line 215: One thing that sticks out to me - would it be worth actually having some JSX in here instead of a static image?

src/empty-states/src/EmptyState.js
```diff
@@ -176,14 +176,23 @@ EmptyState.SecondaryButton = SecondaryButton
 EmptyState.LinkButton = LinkButton
 
 EmptyState.propTypes = {
+  /** The title of the empty state */
```
akleiner2, line 179: 🔥

docs/components/MDX/componentMapping.tsx
kamebkj: Yes for now. We'll likely have to revisit the margin for headings and paragraphs because this is just a patch.

With this change:
image

Without:
image

docs/documentation/patterns/empty-states.mdx
kamebkj: Like SVG in JSX? We can certainly do that, just need more prep from the design side. Is the main benefit scalability?

kamebkj: Gonna stick with static image for now until we see further use cases.

akleiner2 (approved):

Issue #1291, opened 2021-08-10
Upgrade @types/react dependency


zuhair-naqvi:
Any reason why "@types/react" is still set to "^16.9.5"? This will force developers to either add  `"skipLibCheck": true` to their ts config OR add a "resolutions" field in order to resolve "duplicate identifier" errors with typescript if the rest of their packages (in a monorepo) use React 17 types

akleiner2:
Was an oversight as part of the React 17 upgrade -- PRs welcome! I'm also wondering if we even need to install `@types/react` alongside Evergreen, and instead push that responsibility to the consumers. 🤔

zuhair-naqvi:
@akleiner2 cheers, PR issued

brandongregoryscott:
In https://github.com/segmentio/evergreen/releases/tag/v7.1.0, we've removed this as a prod dependency. Feel free to open a new issue if you're still seeing discrepancies.

PR #1297, opened 2021-08-11
Upgrade @types/react to ^17

zuhair-naqvi:
Fixes https://github.com/segmentio/evergreen/issues/1291

**Overview**
There's a dependency on "@types/react" which is set to "^16.9.5"

This will force developers to either add "skipLibCheck": true to their ts config OR add a "resolutions" field in order to resolve "duplicate identifier" errors with typescript if the rest of their packages (in a monorepo) use React 17 types

This PR simply drops @types/react from the deps and makes it a responsibility for the consumers.


package.json
```diff
@@ -57,7 +57,6 @@
   "dependencies": {
     "@babel/runtime": "^7.1.2",
     "@segment/react-tiny-virtual-list": "^2.2.1",
-    "@types/react": "^16.9.5",
```
mshwery, line 60: Won't this cause TS projects to fail if the type exports are different across versions (since Evergreen relies on them in its own  typedefs)?

akleiner2: Since Evergreen now supports react@17 though, maybe we can just update this instead? To avoid react versioning mismatches

zuhair-naqvi: Added @types/react@^17.0.0 to devDependencies

brandongregoryscott:
Closing this per https://github.com/segmentio/evergreen/issues/1291#issuecomment-1413791249

PR #1309, opened 2021-08-24
Info Intent bug on toaster.notify() rectified

hyperloo:
**Overview**
This warning was occurring because info was not provided in the intent's propTypes array. Like in the SS. 

I have added 'info' as the Proptype option in all the files wherever required. Here are thee file names:
1. InlineAlert.js
2. IconButton.js
3. CornerDialog.js
4. Dialog.js
5. MenuItem.js
6. TableRow.js
7. Toast.js

From
image

To
image

**Documentation**

### DIFF
--- a/src/buttons/src/IconButton.js
+++ b/src/buttons/src/IconButton.js
@@ -78,7 +78,7 @@ IconButton.propTypes = {
   /**
    * The intent of the button.
    */
-  intent: PropTypes.oneOf(['none', 'success', 'warning', 'danger']),
+  intent: PropTypes.oneOf(['none', 'success', 'warning', 'danger', 'info']),
 
   /**
    * The appearance of the button.
--- a/src/alert/src/Alert.js
+++ b/src/alert/src/Alert.js
@@ -98,7 +98,7 @@ Alert.propTypes = {
   /**
    * The intent of the alert.
    */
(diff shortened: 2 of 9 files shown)


src/buttons/src/IconButton.js
```diff
@@ -78,7 +78,7 @@ IconButton.propTypes = {
   /**
    * The intent of the button.
    */
-  intent: PropTypes.oneOf(['none', 'success', 'warning', 'danger']),
+  intent: PropTypes.oneOf(['none', 'success', 'warning', 'danger', 'info']),
```
mshwery, line 81: I don't think many of these actually support styles for an `info` intent, as it's considered the same as `none`

mshwery (commented):
@akleiner2 @kamebkj @zkuzmic

src/buttons/src/IconButton.js
kamebkj: +1. The style for `info` is the same as `none`. Instead of adding `info` back, should we clean up where we have `info`?

brandongregoryscott:
This shouldn't be an issue now that we've widened the type in all `PropTypes` calls for `intent` as well as the TS definitions to `string`. See https://github.com/segmentio/evergreen/pull/1341.

PR #1324, opened 2021-09-21
Associate library-generated `description`, `hint` and `validationMessage` nodes with `aria-describedby`

### LINKED ISSUE #1323: Components that accept `description`, `hint`, and `validationMessage` do not create `aria-describedby` associations.
Because `description`, `hint`, and `validationMessage` add extra information to the various `*Field` components, and the relationship between the input and the text is visually apparent, those associations should also be reflected in the accessibility tree. 

Currently, authors have to pass JSX as the value of `description`, `hint`, and `validationMessage` if they want to associate these hints with their inputs. Improving the markup generated when the author supplies hint, description, or validation messages as strings will make it easier for authors to create accessible forms.

## (Contrived) examples

if I write the following code:

``` jsx
<TextInputField
  label="Token"
  hint="Tokens are comprised of 3 words separated by hyphens"
/>
```

I see the hint text I expect in the rendered DOM, but there is no `aria-describedby` attribute connecting the text input to the hint. This means that a user of assistive technology could miss the hint that tells them about the shape of the token.

Again, it's possible to create `aria-describedby` relationships if I write my own JSX, but it's a lot of extra author code if several or more inputs have hint text.

``` jsx
const hintText = (
  <Text
    is="div"
    id="token-hint"
    marginTop={8}
  >
    Tokens are comprised of 3 words separated by hyphens
  </Text>
)

<TextInputField
  label="Token"
  hint={hintPane}
  aria-describedby="token-hint"
/>
```

mxmason:
Closes #1323.
## Overview
When an author provides `description`, `hint` and `validationMessage` values as strings, `evergreen-ui` now generates the right `aria-describedy` values to ensure that these messages are exposed to assistive technology.

⚠️ **Note:** Authors are still required to write their own `aria-describedby` relationships when they provide `description`, `hint` and `validationMessage` as JSX!

## Documentation
I'm happy to mention these new `aria-` associations in the docs, but I'd like some guidance about if and how I should do that, since I don't see a11y spec explicitly mentioned in the docs.

- [x] Updated Typescript types and/or component PropTypes (unneeded in this PR)


src/select/src/SelectField.js
```diff
@@ -55,6 +56,7 @@ const SelectField = memo(function SelectField(props) {
         required={required}
         isInvalid={isInvalid}
         appearance={appearance}
+        aria-describedby={generateAriaDescribedBy(id, { description, hint, validationMessage })}
```
zkuzmic, line 59: This `aria-describedby` isn't reaching the underlying `select` element. Instead, it's being placed on that element's wrapper.

I'd recommend changing the `Select` component to always pass `aria-describedby` down to the inner `select` element.

zkuzmic (changes requested):
@mxmason this is a nice solution, thanks for opening the PR!

Just one change for you. Looks like there's a merge conflict as well.

mxmason:
Thanks for catching my omission, @zkuzmic! I _think_ I fixed it; I will double-check in a few!

zkuzmic (approved):
LGTM 

(sorry for the months-long delay on this one 😅)

PR #1345, opened 2021-10-10
Add ability to render custom autocomplete option when items passed to autocomplete component is array of objects

techsemicolon:
**Overview**

🌲🌲 Hello everyone 🌲🌲

I recently started exploring `evergreen` ui framework and I really like it. 

I found a use case where I wanted to implement `<Autocomplete>` component having items prop which is array of objects. I found a that currently autocomplete supports only array of strings for autocomplete items. So I am making this PR to help others who would like to use this component in similar way. :) 

`<Autocomplete ...../>` component currently uses `itemToString` function to stringify the each item from the `items` array passed through props. 

Now, if you have items like this : 

```js
const fruits = [
  { id: "1", name: "Banana", image: "http://images.com/banana.png"},
  { id: "2", name: "Apple", image: "http://images.com/apple.png"},
  //... 
];
```

Each `AutocompleteItem` of from the autocomplete dropdown receives props which are like this : 

```js
{
  "id": "downshift-x-item-y",
  "children": "[object object]", // This is a string here
  "isHighlighted": false,
  "isSelected": false,
  "key": "[object object]", // This is a string here
  "style": {"position": "absolute", "top": 0, "left": 0, "width": "100%", "height": 32}
}
```

This adds limitation as we do not have access to the item object to render custom option for autocomplete dropdown. Even when Autocomplete component does have `renderItem` prop, it is now limited to items with string elements. 

This PR resolves this by passing the item  in the `renderItem` prop. This enhances flexibility of the `renderItem` to handle custom objects as well as strings.

For example : 

If you have items with structure like this :

```js
const itemsWithdetails = [
    {
        unique_id : 1,
        image_src: "http://images.com/1.png",
        name: "Item 1"
    },
    {
        unique_id : 2,
        image_src: "http://images.com/2.png",
        name: "Item 2"
    },
    //...
];
``` 
Then you can make autocomplete option using available renderItem like below : 

```jsx
<Box padding={40}>
    <Autocomplete
        itemSize={50}
        onChange={handleChange}
        itemToString={item => (item ? item.name : '')}
        items={itemsWithdetails}
        // Custom render dropdown option below with access to original item object, which is then used to get image, name etc 
        // from the item object in the custom renderItem
        renderItem={({ isHighlighted, isSelected, item, ...props }) => (
        <Pane
            display="flex"
            alignItems="center"
            justifyContent="left"
            gap={10}
            key={item.unique_id}
            {...props}
            paddingX={12}
            cursor="pointer"
        >
            <Image src={item.image_src} width={50} />
            <Text color="grey">{item.name}</Text>
        </Pane>
        )}
    >
        {({ getInputProps, getRef, getToggleButtonProps, inputValue, toggleMenu }) => (
        <Box ref={ref => getRef(ref)} display="inline-block">
            <TextInput placeholder="Custom item renderer" value={inputValue} {...getInputProps()} />
            <Button onClick={toggleMenu} {...getToggleButtonProps()}>
            Trigger
            </Button>
        </Box>
        )}
    </Autocomplete>
</Box>
```
I added detailed story for this as well as new component tests. Hope this helps :) 

Screen Shot 2021-10-09 at 6 15 16 PM

**Documentation**
- [x] Added Storybook stories
- [x] Added component tests

🌲🌲 Happy Coding 🌲🌲

### DIFF
--- a/src/autocomplete/stories/index.stories.js
+++ b/src/autocomplete/stories/index.stories.js
@@ -147,7 +147,7 @@ storiesOf('autocomplete', module).add('Autocomplete', () => {
           onChange={handleChange}
           itemToString={item => (item ? item.name : '')}
           items={itemsWithdetails}
-          renderItem={({ item, ...props }) => (
+          renderItem={({ isHighlighted, isSelected, item, ...props }) => (
             <Pane
               display="flex"
               alignItems="center"
@@ -156,6 +156,7 @@ storiesOf('autocomplete', module).add('Autocomplete', () => {
               key={item.unique_id}
               {...props}
               paddingX={12}
+              cursor="pointer"
             >
               <Image src={item.image_src} width={50} />
               <Text color="grey">{item.name}</Text>
--- a/src/autocomplete/src/AutocompleteItem.js
+++ b/src/autocomplete/src/AutocompleteItem.js
@@ -20,7 +20,8 @@ AutocompleteItem.propTypes = {
   children: PropTypes.node,
   style: PropTypes.object,
   isSelected: PropTypes.bool,
-  isHighlighted: PropTypes.bool
+  isHighlighted: PropTypes.bool,
+  item: PropTypes.any
 }
 
 export default AutocompleteItem


src/autocomplete/stories/index.stories.js
```diff
@@ -130,6 +140,42 @@ storiesOf('autocomplete', module).add('Autocomplete', () => {
...
+              <Image src={item.image_src} width={50} />
+              <Text color="grey">{item.name}</Text>
+              <Small color="grey" size={300} flex={1}>
+                (Power: {item.power_score})
+              </Small>
+            </Pane>
+          )}
+        >
+          {({ getInputProps, getRef, getToggleButtonProps, inputValue, toggleMenu }) => (
+            <Box ref={ref => getRef(ref)} display="inline-block">
```
cogwizzle, line 170: What is the value of defining `ref` and passing it directly into the `getRef` function?

Have you considered changing it to be this?
```tsx
<Box ref={getRef} display="inline-block">
```

index.d.ts
```diff
@@ -524,7 +524,11 @@ interface OptionProps extends TableRowOwnProps {
 }
 
 export interface AutocompleteItemProps extends OptionProps {
-  children?: JSX.Element | null
+  children?: JSX.Element | null,
+  /**
+   * The item of the autocomplete items
+   */
+   item: any,
```
cogwizzle, line 531: Have you considered making item a generic to ensure type safety? I think TypeScript is relatively good at inferring the value without explicitly having the pass the generic type.

src/autocomplete/src/Autocomplete.js
```diff
@@ -75,17 +75,18 @@ const AutocompleteItems = ({
             const item = items[index]
             const itemString = itemToString(item)
 
-            return renderItem(
-              getItemProps({
+            return renderItem({
```
cogwizzle, line 78: I'm actually a bit confused about the type here. I noticed that `renderItem` is a property of type any.

Is this meant to be a React component? I realize this is out of scope for this story, but I think it might provide clarity if it is meant to be a React component to render it as a component would be rendered. Have you considered something like this?

```tsx
<RenderItem {...getItemProps({
  item,
  key: itemString,
  index,
  style,
  children: itemString,
  isSelected: itemToString(selectedItem) === itemString,
  isHighlighted: highlightedIndex === index
})} item={item} />
```

Of course this would require you to create an alias from `renderItem: RenderItem` when destructuring the properties on line 43.

index.d.ts
cogwizzle: I see below that we're using propTypes. Not sure if this has any implications, but it could explain this decision.

cogwizzle (commented):
I obviously am not a maintainer, but I figured I would take a look and try and lend a hand while I was reading over the files. Please feel free to reach out if you any questions.

PR #1352, opened 2021-10-29
correctly forward className in code component

### LINKED ISSUE #1351: Code component not forwarding className
I am currently having trouble applying a css property via className to the `<Code>` component. Seems like they aren't being "forwarded" and any css class specified in className does not appear in the rendered html (the css property fullstory-exclude I am applying does not appear). 

image

My workaround so far has been wrapping `<Code>` in a `<div>` See https://codesandbox.io/s/funny-dust-hbts2?file=/index.js

@brandongregoryscott suggested it's most likely due to the ordering of the props being spread after the className prop (comparing to Text where the class name is properly forwarded). 

https://github.com/segmentio/evergreen/blob/master/src/typography/src/Text.js#L27-L30
https://github.com/segmentio/evergreen/blob/master/src/typography/src/Code.js#L15

salolivares:
**Overview**

This PR changes the order the props are spread in `<Code>` so `className` is correctly forwarded. 

Fixes #1351 

Before:
Screen Shot 2021-10-29 at 10 32 33 AM

After:
Screen Shot 2021-10-29 at 10 32 54 AM

**Documentation**

### DIFF
--- a/src/typography/__tests__/Code.test.js
+++ b/src/typography/__tests__/Code.test.js
@@ -44,11 +44,11 @@ describe('Code', () => {
   it('should pass through `className` prop', () => {
     const expected = faker.random.word().toLowerCase()
     const component = (
-      <Code data-testid="text" className={expected}>
+      <Code data-testid="code" className={expected}>
         Testing
       </Code>
     )
     const { getByTestId } = render(component)
-    expect(getByTestId('text')).toHaveClass(expected)
+    expect(getByTestId('code')).toHaveClass(expected)
   })
 })
--- a/src/typography/__tests__/Code.test.js
+++ b/src/typography/__tests__/Code.test.js
@@ -1,0 +1,35 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import faker from 'faker'
+import Code from '../src/Code'
+
+describe('Props', () => {
+  test('<Code /> accepts arbitrary props', () => {
+    const propKey = faker.random
+      .word()
+      .toLowerCase()
+      .replace(/ /g, '')
+    const propValue = faker.random
+      .word()
+      .toLowerCase()
+      .replace(/ /g, '')
+    const component = (
+      <Code data-testid="text" {...{ [propKey]: propValue }}>
+        Testing
+      </Code>
+    )
+    const { getByTestId } = render(component)
+    expect(getByTestId('text')).toHaveAttribute(propKey, propValue)
+  })
+
+  test('<Code /> forwards `className` prop', () => {
+    const expected = faker.random.word().toLowerCase()
+    const component = (
+      <Code data-testid="text" className={expected}>
+        Testing
+      </Code>
+    )
+    const { getByTestId } = render(component)
+    expect(getByTestId('text')).toHaveClass(expected)
+  })
+})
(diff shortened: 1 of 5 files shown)


brandongregoryscott (approved):
Thanks for fixing this up so quickly! If you have some time, would you mind adding a test case for this? We've been trying to add more coverage - this would be a great case to add since we're fixing a bug 😄 

There should be some examples to go off of in `Textarea.test.js` or `TextInput.test.js`

salolivares:
> Thanks for fixing this up so quickly! If you have some time, would you mind adding a test case for this? We've been trying to add more coverage - this would be a great case to add since we're fixing a bug 😄
> 
> There should be some examples to go off of in `Textarea.test.js` or `TextInput.test.js`

Will do 👍

src/typography/__tests__/Code.test.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import faker from 'faker'
+import Code from '../src/Code'
+
+describe('Props', () => {
+  test('<Code /> accepts arbitrary props', () => {
```
akleiner2, line 7: Thanks for the tests! A couple of small nits for consistency's sake: 

- Do you mind changing these to `describe` and `it` blocks? 
- The passthrough of arbitrary props is actually already tested by [ui-box](https://github.com/segmentio/ui-box). Could we keep the test case scoped to `className` forwarding?
- What do you think about just adding a smoke test that the component renders as expected with a "happy path"?

salolivares: @akleiner2 Thanks for the feedback!

* Removed the arbitrary testing of props
* Added "happy path" tests (using snapshots)
* `test` -> `it`

35e227d, 1d60bf9

Let me know if I'm missing anything else 😃

src/typography/__tests__/Code.test.js
```diff
@@ -0,0 +1,54 @@
...
+    ['size 500', 500],
+    ['size 600', 600]
+  ])('<Code /> %s with minimal appearance specified renders as expected', (_, size) => {
+    const component = (
+      <ThemeProvider value={defaultTheme}>
+        <Code size={size} appearance="minimal">{`Text ${size}`}</Code>
+      </ThemeProvider>
+    )
+    const tree = renderer.create(component).toJSON()
+    expect(tree).toMatchSnapshot()
```
brandongregoryscott, line 41: Hmmm... I wish we had a better way to test for this, but I know this is coming from one of the other test suites. I'm fine with leaving it as-is and we can revisit it later

src/typography/__tests__/Text.test.js
```diff
@@ -70,3 +70,16 @@ describe('Sizing', () => {
     `)
   })
 })
+
+describe('Props', () => {
+  it('should forward `className` prop', () => {
+    const MY_CLASS = 'my-custom-class'
+    const component = (
+      <Text data-testid="text" className={`${MY_CLASS}`}>
```
brandongregoryscott, line 78: Nitpick: String interpolation isn't necessary here. Also, it shouldn't _really_ matter, but any thoughts on randomizing the class name similar to the same test for `<Code />`?

brandongregoryscott (approved):
Nothing blocking on my end 😃 thanks for the contribution!

src/typography/__tests__/Text.test.js
salolivares: Ah, missed that. Good point, updated in 2897794

akleiner2 (approved):
Thanks for the revs on this! Nice fix.

Issue #1368, opened 2021-11-22
position prop missing on CornerDialog


martolini:
According to the docs, it should have the `position` prop to place the `CornerDialog`. However, when inspecting the types, it's not there.

Tested on versions 6.6.0 and 6.6.1

PR #1369, opened 2021-11-22
fix: 🐛 Add position prop to the CornerDialogProps type

martolini:
**Overview**
As per the documentation and the source code, the `CornerDialog` component should have position as an optional prop, however, that's not the case in `index.d.ts`.

✅  Closes: #1368

**Screenshots**
Current documentation page
image

**Documentation**


brandongregoryscott (approved):

brandongregoryscott:
Good catch! Thanks 😄

Issue #1370, opened 2021-11-25
Combobox dropdown button doesn't match height of input in Safari


krysia1:
### Problem
In Safari, dropdown button in combobox doesn't match the height of the input (see screenshot or [open docs in Safari](https://evergreen.segment.com/components/combobox))
![Zrzut ekranu 2021-11-25 o 23 36 48](https://user-images.githubusercontent.com/44640941/143505201-7507efe6-6196-481a-9e22-c174a072a858.png)

Note: It does match the height if you define it explicitly, like so
```ts
<Combobox
  openOnFocus
  width="100%"
  height={40}
  items={['Banana', 'Orange', 'Apple', 'Mango']}
  onChange={selected => console.log(selected)}
  placeholder="Fruit"
/>
```

### Expected behavior
The height of the button is the same as the height of the input, just like it is in Chrome (see screenshot)
![Zrzut ekranu 2021-11-25 o 23 42 58](https://user-images.githubusercontent.com/44640941/143505471-756348c4-614d-4c30-8d26-a7fdc604dbf5.png)

### Browser 
Safari Version 14.1.2 (14611.3.10.1.7) also tested on Safari Version 14.0.3 (15610.4.3.1.7)

PR #1371, opened 2021-11-25
fix: remove margin from combobox button for safari

krysia1:
Resolves: #1370 

**Overview**

This explicitly sets combobox top and bottom button margins to zero, so that Safari does not add it's own default margins

**Screenshots**

Before: 

After:

**Documentation**


brandongregoryscott (approved):
Looks good to me in Safari and Chrome. Good catch!

Issue #1375, opened 2021-12-04
Table border styling issues


brandongregoryscott:
There are some minor styling issues with the base `Table` styles, which can be observed on the documentation site (may need to Zoom in to see them): https://evergreen.segment.com/components/table

- Inner content of `Table.Head` is missing `borderTopLeftRadius` and `borderTopRightRadius` style to match its outer container, which causes a sharp edge to render over the rounded border
![image](https://user-images.githubusercontent.com/11774799/144692205-ed36c906-647c-4525-af99-27d3f86b14f2.png)

- Last `Table.Row` is missing `borderBottomLeftRadius` and `borderBottomRightRadius` style to match its outer container, which causes a sharp edge to render over the rounded border
![image](https://user-images.githubusercontent.com/11774799/144692285-1cbf1714-6f89-4bff-9409-51b973b57f98.png)

- Last `Table.Row` has a `borderBottom` style which is overlapping with the bottom border of the outer container causing a thicker border
![image](https://user-images.githubusercontent.com/11774799/144692326-543d8b0f-dc05-4da0-b1a9-ce06dbc9afe2.png)

Gulshanaggarwal:
@brandongregoryscott  can I start working on this issue?

vikasazad:
Hi @brandongregoryscott I would like to contribute to this issue but as i am beginner in opensource contribution. I am not able to start the project in my local machine.
1- I forked it.
2- Run npm install
But as i looked in package.json file the "start" is missing in the scripts and i am not able to resolve this issue.
Please help me out.

brandongregoryscott:
> @brandongregoryscott can I start working on this issue?

@Gulshanaggarwal Of course - feel free to open up a PR!

@vikasazad Please refer to step 2 in the contributing section: https://github.com/segmentio/evergreen/#-step-2-get-storybook-up-and-running

While you can use `npm`, we have set up the project to prefer `yarn`. The command for running a local development server would be `yarn dev` or `npm run dev`.

Gulshanaggarwal:
@vikasazad sorry to say, but I'm already working on the above issue.

Gulshanaggarwal:
@brandongregoryscott  If I apply borderBottomLeftRadius prop inside table.row then it causes the rounded border to all elements, see the  below image-
![Screenshot (241)](https://user-images.githubusercontent.com/58553401/144736748-808641c2-66fd-4c63-a7e2-7ae155e15823.png)


Can you help little bit more?

[...]

PR #1376, opened 2021-12-06
#1375 fixed

Gulshanaggarwal:
I fixed top border styling issues using borderLeft and Right Radius property  and to fixe bottom issues I created a renderLastRow function for last profile item, and set borderBottom to null, so it's border may not overlap with it's outer container.

### DIFF
--- a/src/table/src/TableHead.js
+++ b/src/table/src/TableHead.js
@@ -30,6 +30,8 @@ const TableHead = memo(function TableHead(props) {
       paddingRight={scrollbarWidth}
       className={cx(themedClassName, className)}
       height={height}
+      borderTopLeftRadius="4px"
+      borderTopRightRadius="4px"
       {...boxProps}
       {...rest}
     >
--- a/src/table/src/TableVirtualBody.js
+++ b/src/table/src/TableVirtualBody.js
@@ -183,7 +183,18 @@ const TableVirtualBody = memo(function TableVirtualBody(props) {
   const itemSize = getItemSize(children)
 
   return (
-    <Pane data-evergreen-table-body ref={setPaneRef} height={paneHeight} flex="1" overflow="hidden" {...rest}>
+    <Pane
+      data-evergreen-table-body
+      ref={setPaneRef}
+      borderTopLeftRadius="4px"
+      borderTopRightRadius="4px"
+      borderBottomLeftRadius="4px"
+      borderBottomRightRadius="4px"
+      height={paneHeight}
+      flex="1"
+      overflow="hidden"
+      {...rest}
+    >
       <VirtualList
         height={isIntegerHeight ? paneHeight : calculatedHeight}
         width="100%"
--- a/src/table/stories/AdvancedTable.js
+++ b/src/table/stories/AdvancedTable.js
@@ -229,6 +229,26 @@ export default class AdvancedTable extends React.Component {
     )
   }
 
+  renderLastRow = ({ profile }) => {
+    return (
+      <Table.Row key={profile.id} borderBottom={null}>
+        <Table.Cell display="flex" alignItems="center">
+          <Avatar name={profile.name} />
+          <Text marginLeft={8} size={300} fontWeight={500}>
+            {profile.name}
+          </Text>
+        </Table.Cell>
+        <Table.TextCell>{profile[this.state.column2Show]}</Table.TextCell>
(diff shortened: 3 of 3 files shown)


Gulshanaggarwal:
@brandongregoryscott  I request you to review my PR.

brandongregoryscott (changes requested):
Thanks for taking a look at this, @Gulshanaggarwal. Unfortunately I don't think this solves the root issue.

While your `borderBottom={null}` change fixes that specific story, it does not generically fix it for the last table row in an arbitrary table:

image

I think the proper fix would be adding pseudo selectors `:first-of-type` and `:last-of-type` in the js stylesheets for [`table-head.js`](https://github.com/segmentio/evergreen/blob/master/src/themes/default/components/table-head.js) and [`table-row.js`](https://github.com/segmentio/evergreen/blob/master/src/themes/default/components/table-row.js) respectively. 

These pseudo selectors will also need to be wired up to the component as seen in [`TableRow.js`](https://github.com/segmentio/evergreen/blob/master/src/table/src/TableRow.js#L11-L19).

I may have underestimated the barrier to entry for this issue with the `good-first-issue` tag, sorry! If you'd still like to make these changes, feel free to do so and I will re-review. I am also happy to take on this issue myself to get it resolved.

Gulshanaggarwal:
@brandongregoryscott  I don't know why you are saying this, because if you look at file Advancedtable.js(in bottom side) where I have made changes I have set it to dynamically instead of static.
I have applied my full potential to resolve it, so If you can show me little bit more where the issue is currently now, it will helpful for me.

Thank you!

Gulshanaggarwal:
@brandongregoryscott  can you please explain little bit more what is the new issue is generating because according to me, my solution is perfect.

brandongregoryscott:
@Gulshanaggarwal The issue with the table header is fixed generically, i.e. any random table you create will have correctly rounded top left/top right borders. 

However, the issue with the bottom table row border and border radius is only fixed for the specific example in Storybook that you changed. Try adding another example in [src/table/stories/index.stories.js](https://github.com/segmentio/evergreen/blob/master/src/table/stories/index.stories.js) to see the double bottom border and sharp bottom left/bottom right border radius:

image

This is the table that I created to demonstrate this on your branch:

```diff
storiesOf('table', module)
+  .add('Simple Table', () => (
+    <Box padding={40}>
+      <Table>
+        <Table.Head>
+          <Table.TextHeaderCell>Object</Table.TextHeaderCell>
+          <Table.TextHeaderCell>Action</Table.TextHeaderCell>
+          <Table.TextHeaderCell>Event Name</Table.TextHeaderCell>
+        </Table.Head>
+        <Table.Body>
+          <Table.Row>
+            <Table.TextCell>Product</Table.TextCell>
+            <Table.TextCell>Added</Table.TextCell>
+            <Table.TextCell>Product Added</Table.TextCell>
+          </Table.Row>
+          <Table.Row>
+            <Table.TextCell>Profile</Table.TextCell>
+            <Table.TextCell>Updated</Table.TextCell>
+            <Table.TextCell>Profile Updated</Table.TextCell>
+          </Table.Row>
+        </Table.Body>
+      </Table>
+    </Box>
+  ))
  .add('Advanced Sortable Table', () => (
    <Box padding={24}>
      {(() => {
        document.body.style.margin = '0'
        document.body.style.height = '100vh'
      })()}
      <AdvancedTable />
    </Box>
  ))
```

Since we want to control this behavior by default (not forcing consumers to do `Table.Row` mapping and index calculation for the 'final row' in a table to fix the bottom border issues), it would be better to support this with CSS. Since we're using a CSS-in-JS abstraction, the 'stylesheets' are actually js files broken out by component in the [src/themes/default/components](https://github.com/segmentio/evergreen/tree/master/src/themes/default/components) directory, such as [src/themes/default/components/table-row](https://github.com/segmentio/evergreen/blob/master/src/themes/default/components/table-row.js).

The fix here, as mentioned previously, is to leverage the `:first-of-type` and `:last-of-type` selectors in the `Table.Head` and `Table.Row` styles respectively to add border radius in the appropriate spots, and remove the bottom border for the `Table.Row`. 

Since it was a pretty small change I had to make to demo the fix, I'll be throwing up a PR to get this sorted out. I hope this helps shed some light on how Evergreen works under the hood!

PR #1382, opened 2021-12-10
Generic Theme types, strongly typed DefaultTheme + theme utilities

brandongregoryscott:
**Overview**

Adds generic interfaces & type aliases that allow users to build their own theme objects or easily reference the values set in `defaultTheme`. I added `tsd` as a way to sanity check the type definitions while we're still on JS as well as serve as a playground for quickly inspecting types in VS Code. 

Essentially, `Theme` is the base generic theme which is constrained to a set of `Components` (`'Button'`, `'Input'`, `'Text'`, etc).

The `DefaultTheme` extends `Theme` and overrides some of the generic object types that accept `string` keys, such as `colors`: 

```ts
// generic Theme 
colors: { [color: string]: Color<string | { [group: string]: Color }> }
// DefaultTheme
colors: { [color in DefaultThemeColors]: Color } & {
    border: Color<{ default: string; muted: string }>
    icon: Color<{
      default: string
      muted: string
      disabled: string
      selected: string
    }>
    text: Color<{ danger: string; success: string; info: string }>
  }
```
which is now strongly typed to:
image

A similar strategy is used to expose the various `appearances` and `sizes` for each component. 

While we would like to export a utility for generating interfaces for any generic theme object, we aren't quite there yet. This should act as a holdover for the time being. 

Additionally, I added some other types/utilities:
- `useTheme` is now generically typed, which will default to `DefaultTheme` if not explicitly set.
- Exported a `mergeTheme` function which is really just a wrapper around `_.merge` which should provide consumers an easy way to add/update the default theme as needed, since it does recursive merging unlike `Object.assign` or having to do multiple `...` operations on the whole theme object.
- Exported a `getThemeContext` function which returns a version of the `ThemeContext` which can be typed to a custom theme interface

`mergeTheme` util which returns a union of the `destinationTheme` and `sourceTheme` types:
image

`primary` is being overridden, while `warning` is a new appearance:
image
image

`useTheme` with default type set to `DefaultTheme`:
image

Explicit typing:
image

Factory function for returning a typed context:
image

Since this is such a big change, I'd welcome others to pull down my fork and test out the types locally (assuming you have [`gh`](https://cli.github.com/) and [`yalc`](https://github.com/wclr/yalc) installed)
```
cd ~/your/clone/or/fork/of/evergreen
gh pr checkout 1382
yarn install
yalc publish

cd ~/your/project/using/evergreen
yalc add evergreen-ui
```

**Documentation**


src/theme/src/theme-tools.test.js
```diff
@@ -1,74 +1,149 @@
...
-  it('resolveThemeTokens preserves nested object structures', () => {
-    const theme = {}
-    const props = {
-      baseStyle: {
-        color: 'blue'
-      },
-      appearances: {
-        primary: {
-          backgroundColor: 'pink'
+  describe('mergeTheme', () => {
```
brandongregoryscott, line 28: I typically wouldn't write tests for a function that's just wrapping a 3rd party util, but they were quick to spin up and may prevent a regression if someone were to swap it with a plain object spread/assign

index.d.ts
```diff
@@ -464,8 +464,2262 @@ interface Typography {
 
 export interface Theme {}
 
-export const defaultTheme: Theme
-export const classicTheme: Theme
+export interface DefaultTheme extends Theme {
+  tokens: {
```
brandongregoryscott, line 468: @akleiner2 After digging into the code more, it looks like this `tokens` field is actually deprecated and we're trying to move away from it in favor of bringing all of the other properties (`colors`, `components`, etc) up a level. Wondering what would be the best course of action:

1. Leave it in, since it is currently available and consumers can tap into it
2. Leave it in but add a deprecation warning doc comment
3. Remove it from the type definition since we're trying to remove it from the object entirely

index.d.ts
```diff
@@ -464,8 +464,2262 @@ interface Typography {
...
+      display: string
+      ui: string
+      mono: string
+    }
+    text: {
+      300: {
+        fontSize: string
+        fontWeight: number
+        lineHeight: string
+        letterSpacing: string
```
brandongregoryscott, line 531: I noticed there are some existing types in a `Typography` interface that have some of these values - I didn't do any interface reduction/abstraction, this was a pretty raw conversion from the current object shape. I can go back and try to consolidate some of these common properties into other interfaces if we think it's worth it. What do ya'll think?

index.d.ts
```diff
@@ -464,8 +464,2262 @@ interface Typography {
...
+  lineHeights: string[]
+  zIndices: {
+    focused: number
+    stack: number
+    positioner: number
+    overlay: number
+    toaster: number
+  }
+  components: {
+    Alert: { appearances: {}; sizes: {} }
```
brandongregoryscott, line 761: In my personal project, I opted to leave out the `components` types since I didn't really need them and they added a lot of noise. Since these are more likely to change & will in turn be harder to keep in sync with the actual component styles until we're either ported to TypeScript or have a codegen solution, I am leaning on the side of removing these, but I would love to hear others' thoughts

index.d.ts
```diff
@@ -464,8 +464,2262 @@ interface Typography {
...
+        default: { color: string; backgroundColor: string }
+      }
+      sizes: {}
+    }
+  }
+}
+
+export const defaultTheme: DefaultTheme
+
+export interface ClassicTheme extends Theme {
```
brandongregoryscott, line 1554: I added this just for parity, but I'm not sure how many consumers are leveraging the type definitions + still on the classic theme. I'd imagine there's not many, so I'm happy to remove this too.

anuejn:
Hey, whats the state of this. This would be very nice to have :).

brandongregoryscott:
> Hey, whats the state of this? This would be very nice to have :).

Hey @anuejn! I was hoping to get through this hard-coded version as a holdover, but we wanted to take the time to get the TypeScript types correct to more generically represent the shape of a theme (both the default theme and how it can be extended). I'm working on that in another branch but I don't have an ETA for when it'll be ready. Since evergreen is not natively written in TS, writing the definitions and making sure they are accurate (especially with such a complex object with dynamic props) is clunky and time consuming.

In the meantime, the hard-coded version of `DefaultTheme` can be copied into your local project for reference/overrides. You'll probably have to do some casting, but it was generated from running `JSON.stringify(defaultTheme);` so it should be reasonably accurate.

anuejn:
Okay, thanks for the the quick response. For now I will vendor the code from this branch into said project but I am thrilled to integrate the proper version when it is ready :)

zkuzmic (approved):
@brandongregoryscott this is seriously awesome, great work!

I didn't spend a ton of time checking every type in `index.d.ts`, but am feeling pretty comfortable given all of the tests in `index.test-d.ts`. I also added it to app locally and didn't find any issues. Your draft PR on that repo is looking good as well.

Super excited to never write `useTheme() as any` ever again!

index.d.ts
```diff
@@ -8,6 +8,10 @@ import { TransitionProps, TransitionStatus } from 'react-transition-group/Transi
 
 export { configureSafeHref, BoxProps, BoxOwnProps, BoxComponent, PolymorphicBoxProps, EnhancerProps } from 'ui-box'
 
+// Re-exporting these utility types for testing + consumer usage if needed
+export type Pick<T, Properties extends keyof T> = { [Key in Properties]: T[Key] }
+export type Partial<T> = { [Key in keyof T]?: T[Key] }
+
```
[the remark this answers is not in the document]
brandongregoryscott: These are essentially type maps that you can use to compose a new type from. Both `Pick` and `Partial` are built-in to TS, but I wasn't able to pick them up from the `index.test-d.ts` file for some reason.. so I figured it wouldn't really hurt to re-export them. 

`Pick<T>` lets you pull a subset of properties from another type, i.e.
```ts
interface User {
    firstName: string;
    lastName: string;
    id: string;
    emailAddress: string;
}

// UserIdentity is of the shape { id: string, emailAddress: string }
type UserIdentity = Pick<User, 'id' | 'emailAddress'>
```

`Partial<T>` marks all properties from `T` as optional with the `?` syntax:
```ts
interface User {
    firstName: string;
    lastName: string;
    id: string;
    emailAddress: string;
}

type PartialUser = Partial<User>
// equivalent to this:
interface PartialUser {
    firstName?: string;
    lastName?: string;
    id?: string;
    emailAddress?: string;
}
```

You can also use them in a function signature without declaring a named type alias or interface, if you really want to:
```ts
const pickUserIdentity = (user: User): Pick<User, 'id' | 'emailAddress'> => {
    return { id: user.id, emailAddress: user.emailAddress }
}
```

The TS handbook has a page with some more advanced examples https://www.typescriptlang.org/docs/handbook/2/mapped-types.html and there's a list of the built-in utility types here: https://www.typescriptlang.org/docs/handbook/utility-types.html 😄

brandongregoryscott:
@anuejn This is available in [v6.7.0](https://github.com/segmentio/evergreen/releases/tag/v6.7.0). Try it out and let me know if you run into any issues!

anuejn:
works for me :) (see https://github.com/audapolis/audapolis/pull/227 if you are interested)

anuejn:
Actually one thing would be quite nice for our use case: The ability to put arrays of strings into the colors section of the theme.
This is however nothing that is too important as we can work-around it by creating an object with integer keys for now (for an example use-case, see https://github.com/audapolis/audapolis/pull/227/files#diff-5dd9dde26bd45ecf6de8d7ee6a6a78ba430629ed9c57466fecf746b8d04a2f95R23).

However this already helps a lot and is a very nice solution to the typing problem :)

brandongregoryscott:
Ooh, interesting! We haven't had a use case for a set of color values without specifically named keys yet, but as long as it doesn't muddle with the typing of the string or nested object typing, I think we can probably widen that field's type 👍

PR #1383, opened 2021-12-28
[Fix] Prevent Scrolling gets stuck when other packages/script also change document.body.style.overflow

nicksmider:
**Overview**

When another package/script _OR_ multiple complements use prevent-body-scroll together _previousOverflow_ and _previousPaddingRight_ get stuck.

Scenario observed:

1. React app creates fullscreen overlay(not evergreen), manually sets `document.body.style.overflow = 'hidden'`
2. React app opens Evergreen overlay with `preventBodyScrolling` set overtop the fullscreen. `previousOverflow` is set to `hidden`
3. Evergreen overlay is closed -> `document.body.style.overflow = previousOverflow` keeps it at hidden. Now only the fullscreen overlay(not evergreen) is up
4. The fullscreen module is closed and the React app manually sets `document.body.style.overflow = ''`
5. At this point `document.body.style.overflow` is not set, however, the _next_ Evergreen overlay that does not use `preventBodyScrolling` will default to using the old `previousOverflow` which was never cleared, disabling all scrolling on the overlay, when it was not intended to

The proposed change will create a history of the overlay changes, and once restored, remove it from the history. This will allow multiple overlays to have `preventBodyScrolling` up at a time without having to figure out if another is already up, and it will allow it to play nicely with outside scripts that are also dealing with setting `document.body.style.overflow`

**Documentation** (n/a to this change)

### DIFF
--- a/src/lib/__tests__/prevent-body-scroll.test.js
+++ b/src/lib/__tests__/prevent-body-scroll.test.js
@@ -1,0 +1,62 @@
+import preventBodyScroll from '../prevent-body-scroll'
+
+describe('preventBodyScroll', () => {
+  it('Should change body overflow to hidden on true and reset overflow on false', async () => {
+    expect(document.body.style.overflow).toEqual('')
+    preventBodyScroll(true)
+    expect(document.body.style.overflow).toEqual('hidden')
+    preventBodyScroll(false)
+    expect(document.body.style.overflow).toEqual('')
+  })
+  it('Should be able to handle changes from other scripts changing body overflow', async () => {
+    // Repeat logic of first test
+    expect(document.body.style.overflow).toEqual('')
+    preventBodyScroll(true)
+    expect(document.body.style.overflow).toEqual('hidden')
+    preventBodyScroll(false)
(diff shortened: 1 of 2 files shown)


brandongregoryscott (approved):
Hmm, makes sense to me. Is there an easy way to demo this change in Storybook or the docs or would it be easier to demo in the app?

Additionally, while I know there's no existing unit tests for this utility, it would be awesome to whip up a few at least to cover the new behavior 👍 Food for thought - non-blocking, as this seems like a pretty jarring bug as-is

nicksmider:
> Hmm, makes sense to me. Is there an easy way to demo this change in Storybook or the docs or would it be easier to demo in the app?
> 
> Additionally, while I know there's no existing unit tests for this utility, it would be awesome to whip up a few at least to cover the new behavior 👍 Food for thought - non-blocking, as this seems like a pretty jarring bug as-is

Demoing in the app would definitely be easiest. I can whip up some tests 👍

brandongregoryscott (approved):
Thanks for whipping up those tests!

PR #1384, opened 2021-12-29
Add intent property to Button component

yurimarimo:
**Overview**
There is no indications about the "intent" property in the Button component's document now, so I added it.
https://evergreen.segment.com/components/buttons/props

**Documentation**


brandongregoryscott (approved):
Good catch - thanks for the fix @yurimarimo!

PR #1392, opened 2022-01-18
Update theme documentation

brandongregoryscott:
**Overview**

Updates some of the `/theming` documentation to include references to new TS types, as well as the `mergeTheme` utility function. Fixes some misc. typos along the way, and updates the CodeSandbox link to pull in the same version of evergreen that the doc site is using :)

Rich diff should be visible within the Github PR.

**Documentation**

### DIFF
--- a/docs/documentation/introduction/theming.mdx
+++ b/docs/documentation/introduction/theming.mdx
@@ -4,9 +4,9 @@ Evergreen ships with an all-new, extensible theming architecture that lets end-u
 
 # Classic and Default Theme
 
-For starters, Evergreen ships with two themes. The `defaultTheme` is the current brand (and the one we use internally at Segment!), and the `classicTheme` is the same look and feel from the previous iterations of Evergreen (v4+).
+For starters, Evergreen ships with two themes. The `defaultTheme` is the current brand (and the one we use internally at Segment!), and the `classicTheme` is the same look and feel from the previous iterations of Evergreen (v4 and v5).
 
-> If you are using a previous version of upgrade, you'll want to explicitly set your theme to the classicTheme
+> If you are using a previous version of Evergreen, you'll want to explicitly set your theme to the classicTheme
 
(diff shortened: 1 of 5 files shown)


docs/documentation/introduction/theming.mdx
```diff
@@ -68,11 +66,11 @@ function CustomButtonExample() {
         },
       },
     },
-  }
+  })
 
   return (
     <ThemeProvider value={theme}>
-      <Button>Custom Button</Button>
+      <Button appearance="none">Custom Button</Button>
```
brandongregoryscott, line 73: Looks like the only reason this was working without `appearance="none"` before was because of the multiple object spreads - the entire `Button` object was being replaced (including all appearances), this appeared to work correctly for the example but it would also remove the `primary`, `minimal`, etc. appearances + other sizes.

It would've needed a `...defaultTheme.components.Button` before `baseStyle` on line 53, and which would've shown the need for `appearance="none"`

docs/lib/codesandbox.ts
```diff
@@ -20,9 +21,9 @@ export const getCodeSandboxLink: (source: string) => string = (source) => {
 
   const usedComponents = Array.from(
     new Set(
-      (source.match(/<((\w+))|minorScale|majorScale|(((\w+)Icon))|Position/g) || []).map((component) =>
-        component.replace('<', '')
-      )
+      (
+        source.match(/<((\w+))|minorScale|majorScale|mergeTheme|defaultTheme|(((\w+)Icon))|Position/g) || []
```
brandongregoryscott, line 25: This will include `mergeTheme` and `defaultTheme` in the imports for a codesandbox link if the example needs it, which should prevent confusion for the live code examples in the doc site that "just work" and then have missing imports in CSB.

zkuzmic (approved):
LGTM!

PR #1397, opened 2022-01-28
[Fix] Popover imperative open/close can break due to dependency array.

jon-quin:
I encountered this reproducible bug while using a `SelectMenu`. The imperative `open()` and `close()` can become "broken" until a React re-render is triggered by something else.

Fix:
Using `ref.current` doesn't make sense in dependency arrays. The dependencies here are `open()` and `close()` which are behind `useCallback()` so perf shouldn't take an unnecessary hit.

### DIFF
--- a/src/popover/src/Popover.js
+++ b/src/popover/src/Popover.js
@@ -41,15 +41,6 @@ const Popover = memo(
     const targetRef = useRef()
     const setTargetRef = useMergedRef(targetRef)
 
-    useImperativeHandle(
-      forwardedRef,
-      () => ({
-        open,
-        close
-      }),
-      [popoverNode.current]
-    )
-
     /**
      * Methods borrowed from BlueprintJS
      * https://github.com/palantir/blueprint/blob/release/2.0.0/packages/core/src/components/overlay/overlay.tsx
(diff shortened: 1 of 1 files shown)


brandongregoryscott:
Hey @jon-quin - would you mind spinning up a CodeSandbox example that reproduces this issue? Not sure it's something we've encountered.

jon-quin:
@brandongregoryscott I've set up a CodeSandbox repro here: https://codesandbox.io/s/beautiful-bardeen-vjtr9?file=/src/App.js . LMK if it doesn't work for you.

src/popover/src/Popover.js
```diff
@@ -47,7 +47,7 @@ const Popover = memo(
         open,
         close
       }),
-      [popoverNode.current]
+      [open, close]
```
brandongregoryscott, line 50: Looks good to me (and makes sense - it seems the dependency arrays in that file need some work, but we can tackle those at another time.) Just a heads up, you might have better luck with this controlled behavior by using the `isShown` prop instead of grabbing the ref and calling `open`/`close` manually.

brandongregoryscott (approved):

src/popover/src/Popover.js
brandongregoryscott, line 50: Looks like the tests are failing because `open` and `close` are used before they are defined. Mind rearranging them to resolve this? @jon-quin

brandongregoryscott (changes requested):

src/popover/src/Popover.js
jon-quin: Thanks, I'll take a look at `isShown`!

jon-quin: done, and manually tested working.

brandongregoryscott (approved):

PR #1398, opened 2022-02-01
Add isInvalid field to tagInput

hellooimkat:
**Overview**
We're introducing an `isInvalid` field to the tag input component.

Screen Shot 2022-01-31 at 4 00 59 PM

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2499,6 +2499,7 @@ export interface TagInputOwnProps {
   addOnBlur?: boolean
   className?: string
   disabled?: boolean
+  isInvalid?: boolean
   height?: number
   inputProps?: PolymorphicBoxProps<'input', TextOwnProps>
   inputRef?: React.Ref<HTMLInputElement>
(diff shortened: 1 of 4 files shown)


brandongregoryscott (changes requested):
Instead of pulling in `useTheme` and assigning the borderColor prop here directly, how do you feel about wiring up a pseudo-selector for the invalid state similar to the `TextInput` component? I think we typically try to prefer this approach for the internal styles of the components we provide. 

https://github.com/segmentio/evergreen/blob/2f73d2d8a7102a07bce403a5782fd6218413a870/src/themes/default/components/input.js#L33-L35

https://github.com/segmentio/evergreen/blob/2f73d2d8a7102a07bce403a5782fd6218413a870/src/text-input/src/TextInput.js#L12

☝️ in addition, we should be setting the `aria-invalid` property based on `isInvalid` 👇 

https://github.com/segmentio/evergreen/blob/2f73d2d8a7102a07bce403a5782fd6218413a870/src/text-input/src/TextInput.js#L65

Finally.. I think we'll want to update the typescript definitions to reflect this new prop :) Updating the `PropTypes` provides runtime validation and is used to parse out props for the doc site (I believe), but it should be as simple as adding in `isInvalid?: boolean` to `TagInputOwnProps`: 
https://github.com/segmentio/evergreen/blob/2f73d2d8a7102a07bce403a5782fd6218413a870/index.d.ts#L2498-L2515

PR #1406, opened 2022-02-11
Append a Menu close demo

tiyee:
Close the opened state inside of the Menu.

you can show it like this

```js
/** @format */

import {Popover, Menu, Position, MenuIcon, IconButton} from 'evergreen-ui'
import React from 'react'
import ReactDOM from 'react-dom'

function MenuWithClose() {
    const [open, setOpen] = React.useState(false)
    return (
        <Popover
            isShown={open}
            onOpen={() => {
                setOpen(true)
            }}
            position={Position.BOTTOM_LEFT}
            content={
                <Menu>
                    <Menu.Group>
                        <Menu.Item onSelect={() => setOpen(false)}>Close</Menu.Item>
                    </Menu.Group>
                </Menu>
            }
            shouldCloseOnExternalClick={true}>
            <IconButton icon={MenuIcon} height={40} />
        </Popover>
    )
}

ReactDOM.render(<MenuWithClose />, document.getElementById('root'))

```

### DIFF
--- a/docs/documentation/components/menu.mdx
+++ b/docs/documentation/components/menu.mdx
@@ -42,6 +42,35 @@ It is important to understand the Menu component by itself does not manage a dro
 </Menu>
 ```
 
+## Close the opened state inside of the Menu
+
+you can be able to close the popover inside of the content. It's more flexible than that in [close_from_within_content](https://evergreen.segment.com/components/popover#close_from_within_content)
+
+
+```js
+function MenuWithClose() {
+    const [open, setOpen] = React.useState(false)
+    return (
+        <Popover
+            isShown={open}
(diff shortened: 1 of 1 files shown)


brandongregoryscott:
Hi @tiyee, thanks for the contribution. While this is certainly an acceptable usage of the `Popover` and `Menu` components, it is a more advanced use-case that we aren't necessarily trying to push people to use unless there's a good case for it. Specifically, controlling the `Popover` state yourself has some drawbacks:

The `shouldCloseOnExternalClick` prop has no effect (which you should notice in your example), as well as losing out on the automatic close when the `ESC` key is hit.

PR #1414, opened 2022-02-25
update evergreen-ui and add docs for TagInput autocompleteItems prop

zkuzmic:
**Overview**
Update evergreen-ui package in docs and add autocomplete demo for `TagInput`.

Screen Shot 2022-02-25 at 11 38 43 AM

**Documentation**

### DIFF
--- a/docs/documentation/components/tag-input.mdx
+++ b/docs/documentation/components/tag-input.mdx
@@ -19,7 +19,7 @@ function BasicTagInputExample() {
     <TagInput
       inputProps={{ placeholder: 'Add trees...' }}
       values={values}
-      onChange={values => {
+      onChange={(values) => {
         setValues(values)
       }}
     />
@@ -40,7 +40,7 @@ function SubmitKeyTagInputExample() {
       inputProps={{ placeholder: 'Add trees...' }}
       values={values}
       tagSubmitKey="space"
-      onChange={values => {
+      onChange={(values) => {
         setValues(values)
       }}
     />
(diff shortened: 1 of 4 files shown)


kamebkj:
Looking good! One small question and maybe I'm not using it correctly, but when I try to add a new email value in the "Using onAdd and onRemove (advanced)" section, I get an error of invalid email. 
taginput-email

zkuzmic:
> Looking good! One small question and maybe I'm not using it correctly, but when I try to add a new email value in the "Using onAdd and onRemove (advanced)" section, I get an error of invalid email. taginput-email

Oof, looks like that was an existing bug. I can add a fix.

docs/documentation/components/tag-input.mdx
```diff
@@ -122,20 +122,43 @@ function TagPropCallbackTagInputExample() {
   const [values, setValues] = React.useState(['Kauri', 'Willow'])
   return (
     <TagInput
-      tagProps={value => {
+      tagProps={(value) => {
```
brandongregoryscott, line 125: 🤔 Is there a different `prettier` config for the `docs` site? I've seen some churn with the arrow parens + trailing commas in some other files in the repo, but I'm not quite sure where it's coming from 😅 

Not a big deal, but I'd love to get to the bottom of it 😛

brandongregoryscott (approved):

docs/documentation/components/tag-input.mdx
zkuzmic: Ooo looks like there is a separate config. Didn't realize that.

brandongregoryscott: I wonder if we break out the root config from `package.json` into a `.prettierrc` file - would it use that for both `src` and `docs`? 🤔 ideally they'd match

PR #1430, opened 2022-03-29
[SSCX-517] update positioner to check for dynamic children

neztEx:
**Overview**
Updated the positioner to check for dynamically added children to allow them to be tied to the parent coordinates. Previously, children could render outside the viewport.

JIRA: https://segment.atlassian.net/browse/SSCX-517

https://www.loom.com/share/aff18bb8127d4901aa327a315b172558

**Documentation**

### DIFF
--- a/src/positioner/src/Positioner.js
+++ b/src/positioner/src/Positioner.js
@@ -119,7 +119,7 @@ const Positioner = memo(function Positioner(props) {
   )
 
   // Call `update` whenever the component has "entered" and dimensions change
-  // additionally, when have dynamic children
+  // additionally, when there are dynamic children
   useEffect(() => {
     if (transitionState.current === 'entered') {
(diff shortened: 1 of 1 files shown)


brandongregoryscott (approved):

src/positioner/src/Positioner.js
```diff
@@ -131,7 +132,7 @@ const Positioner = memo(function Positioner(props) {
         cancelAnimationFrame(latestAnimationFrame.current)
       }
     }
-  }, [previousDimensions.height, previousDimensions.width, update])
+  }, [previousDimensions.height, previousDimensions.width, update, children])
```
brandongregoryscott, line 135: Nice - glad you tested out adding it to this side effect vs. adding another one in. 💯

PR #1458, opened 2022-04-22
React 18 support

brandongregoryscott:
**Overview**

- Updates our dev dependencies for react/react-related packages to v18.
- Widens our peer dependency requirements to include React v18

I wanted to move the two `@types` packages that are in our production dependency list, but I'm honestly not sure how to test to see if they are required for our `index.d.ts` yet (i.e. if anything breaks for an end-user if we don't ship with those types). I imagine if those types _are_ needed because we're referencing them in some of our own type definitions, we could specify `@types/react` as a peer dependency instead, but that might be a problem for another day.

I pulled in a local version of the package in `app` and it seems to run fine (no console warning since we're on an older version of React). 

In a React v18 app, it will still produce the warning for now.

```
ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it’s running React 17. Learn more: https://reactjs.org/link/switch-to-createroot
```

**Documentation**

### DIFF
--- a/src/typography/__tests__/Text.test.js
+++ b/src/typography/__tests__/Text.test.js
@@ -59,7 +59,7 @@ describe('Sizing', () => {
 
   test('<Text /> has undefined behavior when trying to set arbitrary sizes', () => {
     render(<Text size={800} />)
-    expect(mockFn.mock.calls.length).toEqual(1)
+    expect(mockFn.mock.calls.length).toBeGreaterThanOrEqual(1)
     expect(mockFn.mock.calls[0][0]).toMatchInlineSnapshot(`
       Extracted Styles:
       box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


brandongregoryscott:
Drafting this - until we update the way we're rendering rendering the Tooltip w/ the new recommended `createRoot` function, React v18 users are going to get the same warning with or without these changes

src/typography/__tests__/Text.test.js
```diff
@@ -59,7 +59,7 @@ describe('Sizing', () => {
 
   test('<Text /> has undefined behavior when trying to set arbitrary sizes', () => {
     render(<Text size={800} />)
-    expect(mockFn.mock.calls.length).toEqual(1)
+    expect(mockFn.mock.calls.length).toBeGreaterThanOrEqual(1)
```
brandongregoryscott, line 62: The warning from `ReactDOM.render` added an additional call here 🤦

krousseau (approved):

PR #1469, opened 2022-05-05
Update deploy-docs script to amend version commits

brandongregoryscott:
**Overview**

Small tweak to the previous change to `yarn deploy-docs` that is run in the `postpublish` step. For one:
1. I separated out the two for clarity (bumping the evergreen version in docs/ from actually deploying the doc site)
1. If the previous commit matches a version number, i.e. `v6.9.10` from running `yarn release`, we'll amend the commit that bumps the doc site `package.json`/`yarn.lock` instead of adding a new one. If for some reason there was another commit in between, it'll just add a new commit.

This should hopefully keep our history a little cleaner while keeping the doc site up-to-date! (I started noticing these doc site version bump commits showing up in each set of release notes and it just adds noise)

Examples of how I tested:
```
git commit --allow-empty -m 'v1.1.1'
git commit --allow-empty -m 'something else'
git show-branch --no-name HEAD | grep -E 'v[0-9]+.[0-9]+.[0-9]+' && echo 'matches release commit' && echo 'should amend commit' || echo 'create new commit'
```

**Documentation**

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,17 +40,18 @@
     "clean": "git clean -Xdf",
     "create-package": "./tools/create-package.js",
     "create-package:components": "./tools/create-package-components.js",
-    "deploy-docs": "cd docs && yarn add evergreen-ui@latest --exact && git add package.json yarn.lock && git commit -m \"Updated doc site evergreen version\" || yarn install && yarn deploy",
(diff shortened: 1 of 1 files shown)


zkuzmic (approved):
LGTM 🎉

PR #1473, opened 2022-05-20
fix esc on tagInput component

### LINKED ISSUE #1472: Pressing `esc` while focused on `TagInput` component throws error
Steps to reproduce:

1. Click or tab into an enabled `TagInput` component, such as the ones on the doc site https://evergreen.segment.com/components/tag-input
1. Press the `esc` key, and you should get the following runtime error:
`TypeError: Cannot read properties of null (reading 'split')`

image

At a first glance, it seems like pressing `esc` is setting the internal `inputValue` (which is expected to be a `string`) to `null`, which breaks the string-splitting behavior for generating the tag list.

pandurijal:
**Overview**
Resolves #1472 

The proposed behavior for pressing `esc`:
- On the TagInput Autocomplete, pressing `esc` will close the dropdown
- On the any other TagInput components, pressing `esc` will do nothing (the same as `TextInput` and `TextArea` component)

**Documentation**

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -138,6 +138,15 @@ const Autocomplete = memo(
           }
         }
 
+        if (Downshift.stateChangeTypes.keyDownEscape) {
+          return {
+            ...changes,
+            ...state,
+            isOpen: false,
+            inputValue: state.selectedItem.label
+          }
+        }
+
         return changes
       },
       [props.items, props.allowOtherValues]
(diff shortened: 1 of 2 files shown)


brandongregoryscott:
Hey @pandurijal! Thanks for the contribution. I was in the middle of typing up a response about some issues I encountered and it looks like your most recent commit resolved them! (and fixed the failing tests - glad it was caught!) Gonna do some additional smoke testing in storybook before I merge it in, but this is looking good now.

pandurijal:
Hi @brandongregoryscott!

Well yeah, I thought the issue was from the Downshift component, turns out your approach could fix the issue without having any additional effect on another component(s). So I think I would go with your solution after testing some more things I will push the changes.

brandongregoryscott (approved):
Looks fixed on the Storybook preview. Thanks for the contribution!

PR #1476, opened 2022-06-02
[SSCX-592] Center EG Autocomplete for TagInput on the input

catligad:
JIRA: https://segment.atlassian.net/browse/SSCX-592

**Overview**
Currently, this component's autocomplete items only appear on top of the input:

https://user-images.githubusercontent.com/34897995/171701929-10599c4c-01c0-43f6-bc21-c85e9103c0a8.mov

We are changing this so that the autocomplete items appear on top of the whole tag input.

No width passed in:
https://user-images.githubusercontent.com/34897995/171701640-8957fe2e-53fa-4f44-bbed-947a69ed0cf3.mov

Width passed in:
https://user-images.githubusercontent.com/34897995/171701659-01f5af8b-ff1c-418b-9104-c49d60c3ca41.mov

**Documentation**
- [ (no need) ] Updated Typescript types and/or component PropTypes
- [ (no need) ] Added / modified component docs
- [ (no need) ] Added / modified Storybook stories

### DIFF
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -116,9 +116,10 @@ const Autocomplete = memo(
     const [targetRef, setTargetRef] = useState()
 
     useEffect(() => {
-      const boundingWidth = targetRef?.getBoundingClientRect().width
-      setTargetWidth(boundingWidth)
-    }, [targetRef, setTargetWidth, props.items.length])
+      if (targetRef) {
+        setTargetWidth(targetRef.getBoundingClientRect().width)
+      }
+    }, [targetRef])
 
     const stateReducer = useCallback(
       (state, changes) => {
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -119,7 +119,7 @@ const Autocomplete = memo(
       if (targetRef) {
         setTargetWidth(targetRef.getBoundingClientRect().width)
       }
-    }, [targetRef])
+    }, [targetRef, props.items.length])
 
     const stateReducer = useCallback(
       (state, changes) => {
--- a/src/autocomplete/src/Autocomplete.js
+++ b/src/autocomplete/src/Autocomplete.js
@@ -116,10 +116,9 @@ const Autocomplete = memo(
     const [targetRef, setTargetRef] = useState()
 
     useEffect(() => {
-      if (targetRef) {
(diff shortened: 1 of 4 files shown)


zkuzmic:
Hmm, it seems like the location of the dropdown is off with these changes:

https://user-images.githubusercontent.com/1144042/171729113-7e0728b3-f7b5-4507-99c5-88564b2d04aa.mov

I'm thinking it should either be the full width of the input or be aligned with the text input like it was before. What do you think?

catligad:
@zkuzmic whoops yes -- let me fix that.

src/tag-input/src/TagInput.js
```diff
@@ -185,15 +187,14 @@ const TagInput = memo(
...
-        {values.map(maybeRenderTag)}
         <Box flexGrow="1" display="inline-block">
           <Autocomplete
             onChange={changedItem => {
               addTags(changedItem)
               setInputValue('')
             }}
             items={hasAutocomplete ? autocompleteItems : []}
-            id={inputId}
+            id={autocompleteId}
```
catligad, line 197: Since we should only have 1 element with 1 ID, I think this is alright; is this ok? Is it ok that we're dynamically changing the ID here?

Context: We have to have a prop that changes in the autocomplete element in order to resize it to the appropriate ref; we could use the `items` prop, but if there are elements that are not in the items that are used in the input, the resize will not work.

I went with this approach instead of passing down a `resizeAutocomplete` or similar function to the child component that would be called on change because I think we can handle this internally.

src/autocomplete/src/Autocomplete.js
```diff
@@ -116,10 +116,9 @@ const Autocomplete = memo(
     const [targetRef, setTargetRef] = useState()
 
     useEffect(() => {
-      if (targetRef) {
-        setTargetWidth(targetRef.getBoundingClientRect().width)
-      }
-    }, [targetRef])
+      const boundingWidth = targetRef?.getBoundingClientRect().width
+      setTargetWidth(boundingWidth)
+    }, [targetRef, setTargetWidth, props.items.length, props.id])
```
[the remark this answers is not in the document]
zkuzmic: Makes sense to me 👍

src/tag-input/src/TagInput.js
zkuzmic: Ah yeah, good catch with the `id`. Not sure why that was being set on the Autocomplete as well.

LGTM!

zkuzmic (approved):
LGTM

Issue #1479, opened 2022-06-13
Interactive badge styles not applied with classnames@2.2.6


mhssmnn:
**Describe the bug**
`isInteractive` for Badge does not work when evergreen is used within a project that has classnames v2.2.6

Other areas may be affected too.

Fix is to bump the version range for classnames from `^2.2.6` to `^2.3.0` – which is the version that is resolved in the `yarn.lock` file, and therefore what is used in the examples (which is why you do not see the bug by default)

**To Reproduce**

1. Use a project with `classnames` at `^2.2.6`
2. Create a `<Badge isInteractive />` component
3. Hover

Reproduced here: https://codesandbox.io/s/delicate-thunder-0sqb3d

**Expected behavior**
Opacity of Badge to drop to 0.8

**Evergreen Version**
v6.9.10

**Environment (please complete the following information):**

- OS: MacOS
- Browser Chrome
- Version 102

PR #1480, opened 2022-06-13
Update classnames version range to ^2.3.0

mhssmnn:
**Overview**
This PR simply updates the classnames version range to align with what is being used in the yarn.lock, and resolves the #1479 bug

**Documentation**
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "@types/react": "^16.9.5",
     "@types/react-transition-group": "^4.4.0",
     "arrify": "^1.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "@types/react": "^16.9.5",
     "@types/react-transition-group": "^4.4.0",
     "arrify": "^1.0.1",
-    "classnames": "^2.2.6",
+    "classnames": "^2.3.0",
```
brandongregoryscott, line 64: Great catch, thanks!

PR #1489, opened 2022-06-29
Unrequired length check in autocomplete

cogwizzle:
**Overview**
I noticed that we were performing a similar check twice in the same file when reviewing the large upcoming typescript change. I tried to update the code to remove the redundancy. A small change but every little bit helps.

**Documentation**


cogwizzle:
If it helps get me workflow approval I did run my test locally and they passed.

PR #1490, opened 2022-06-29
fix(example)Next JS _document error

TIMMLOPK:
**Overview**
If using nextjs `=>12.1.6` , may get `error: "MyDocument.getInitialProps()" should resolve to an object with a "html" prop set with a valid html string`

**Documentation**

### DIFF
--- a/examples/ssr-next/pages/_document.js
+++ b/examples/ssr-next/pages/_document.js
@@ -4,8 +4,8 @@ import { extractStyles } from 'evergreen-ui'
 import Document, { Head, Main, NextScript } from 'next/document'
 
 export default class MyDocument extends Document {
-  static getInitialProps({ renderPage }) {
-    const page = renderPage()
+  static async getInitialProps({ renderPage }) {
+    const page = await renderPage()
     // `css` is a string with css from both glamor and ui-box.
     // No need to get the glamor css manually if you are using it elsewhere in your app.
     //


cogwizzle:
Would this break any existing functionality? I can see that the current version supports the logic that you recommended but I am just curious if it would break existing codebases that haven't updated to 12.x yet.

brandongregoryscott:
@TIMMLOPK Did you mean v12.2+? I don't think NextJS has a v12.6 out yet: https://www.npmjs.com/package/next

In either case, I'm not sure I'm able to reproduce this. Can you provide a CodeSandbox for reproduction?

TIMMLOPK:
> @TIMMLOPK Did you mean v12.2+? I don't think NextJS has a v12.6 out yet: https://www.npmjs.com/package/next
> 
> In either case, I'm not sure I'm able to reproduce this. Can you provide a CodeSandbox for reproduction?

It is 12.1.6,I correct it.I would try to make reproduction

TIMMLOPK:
@brandongregoryscott seem can't make it in codesandbox because it crash.I made a repo for it :https://github.com/TIMMLOPK/error-reproduction

TIMMLOPK:
> Would this break any existing functionality? I can see that the current version supports the logic that you recommended but I am just curious if it would break existing codebases that haven't updated to 12.x yet. I can't find older documentation, but here is some documentation from NEXT supporting @TIMMLOPK's change. https://nextjs.org/docs/advanced-features/custom-document#customizing-renderpage

No , I tried to use 11.x ,it doesn't matter

brandongregoryscott (approved):

PR #1494, opened 2022-07-08
Adds suport to react 18 to toaster component

JesuHrz:
**Overview**
This PR adds support for react 18 within the Toaster component as it is not covered in this component.

With this implementation we will no longer see this log which can be annoying at times
image

In this screenshot, we can see the react version that is being used which is 18 or higher
image

In this screenshot, we can see the react version that is being used which is lower than 18
image


src/toaster/src/Toaster.js
```diff
@@ -15,15 +16,29 @@ export default class Toaster {
     container.setAttribute('data-evergreen-toaster-container', '')
     document.body.appendChild(container)
 
-    ReactDOM.render(
+    const el = () => (
```
brandongregoryscott, line 19: Can we name this `toastManager`? (Additionally, I don't think it needs to be wrapped in a function since both `createRoot(container).render()` and `render()` accept the component instance)

src/toaster/src/Toaster.js
```diff
@@ -15,15 +16,29 @@ export default class Toaster {
...
         bindNotify={this._bindNotify}
         bindRemove={this._bindRemove}
         bindGetToasts={this._bindGetToasts}
         bindCloseAll={this._bindCloseAll}
-      />,
-      container
+      />
     )
+
+    if (getMajorVersion(ReactDOM.version) >= 18) {
```
brandongregoryscott, line 28: This is clever! I hadn't thought to introspect on the package itself to see if we could conditionally choose which API to use for rendering. Have you seen any other libraries taking this same approach to opt-in to the new API depending on the running React version?

Instead of just returning in the `catch` block, how do you feel about falling back to `ReactDOM.render` in the event something throws here? I'm not sure what might cause this block to `throw`, but defensively trying to render using the function that was previously the standard seems better than giving up entirely. We know it should be available to use until at least the next major version of React.

src/lib/getMajorVersion.js
```diff
@@ -0,0 +1,10 @@
+/**
+ *
+ * @param {string} version
+ * @returns {number} The major version of ReactDOM
```
brandongregoryscott, line 4: Since this function might be used with other packages or strings, maybe we drop the bit about 'ReactDOM'? An example might help too:

`@example getMajorVersion('17.0.2') // 17`

brandongregoryscott (changes requested):

src/toaster/src/Toaster.js
JesuHrz: I wrapped it inside a function because this way we don't have to repeat the same code of the component inside the `.render()` function because it will always be the same

Thanks for the feedback @brandongregoryscott

src/toaster/src/Toaster.js
JesuHrz: I've done this approach sometimes on some projects and they have working for now.

Otherwise, yep you are right, with this we can be more confident that if something happens we will render it with the previous version. 

Thanks for the feedback @brandongregoryscott

JesuHrz:
Hey @brandongregoryscott, I just made all changes about your feedback.

Now it is already for your approval

brandongregoryscott:
@JesuHrz It looks like the tests are failing in CI - do they pass locally for you?

tools/test-setup-after-env.js
```diff
@@ -1 +1,3 @@
 import '@testing-library/jest-dom'
+
+jest.mock('scheduler', () => require('scheduler/unstable_mock')) // eslint-disable-line
```
brandongregoryscott, line 3: Can we add a comment on what this is for? I'm guessing it's related to the new `createRoot` call in React 18, but it might be nice to add some context! (Plus the specific rule that's being disabled here):

`// eslint-disable-next-line some-rule -- Scheduler needs to be mocked for React 18`

brandongregoryscott (changes requested):

tools/test-setup-after-env.js
JesuHrz: yep, this makes a lot of sense, thanks for the feedback

JesuHrz: yep, this makes a lot of sense, thanks for the feedback - done! @brandongregoryscott

JesuHrz:
Hey @brandongregoryscott, the last changes are already for your approval

brandongregoryscott (approved):
Pulled it down and tested it - seems to work with both 17 & 18!

PR #1500, opened 2022-07-14
Update documentation for Tooltip.isShown prop

### LINKED ISSUE #1493: Ability to conditionally render Tooltip
There are cases where we want to display a `Tooltip` (or rather, have the _ability_ to display a `Tooltip` on an element) based on a condition. The current `isShown` prop does not support this very well - it controls whether the Tooltip is **always** shown, not just when the child element is hovered. It would be nice to add a new prop to control whether the tooltip _should_ be shown once the child element is hovered.

[This CodeSandbox](https://codesandbox.io/s/conditional-tooltip-example-c8q2ev?file=/src/App.tsx) demos the current behavior with `isShown` (which should remain the same), and how you'd currently need to workaround it to fit the use-case.

brandongregoryscott:
**Overview**

Resolves #1493 

Adds more explicit documentation on what each of the states of the `isShown` prop for the Tooltip (and Popover) do. I added an example to the Tooltip page to demo this functionality. 

image

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1874,28 +1874,88 @@ export declare const Pill: BoxComponent<PillOwnProps, 'strong'>
 export type PopoverStatelessProps = BoxProps<'div'>
 
 export interface PopoverProps {
+  /**
+   * The position the Popover is on. Smart positioning might override this.
+   */
   position?: PositionTypes
+  /**
+   * Controls whether the Popover is shown or not.
+   * - When `true`, the component is manually shown.
+   * - When `false`, the component is manually hidden.
+   * - When `undefined`, the component is uncontrolled and the isShown state is handled internally
+   */
   isShown?: boolean
+  /**
+   * Open the Popover based on click or hover. Default is click.
+   */
   trigger?: 'click' | 'hover'
+  /**
+   * The content of the Popover.
+   */
   content: React.ReactNode | ((object: { close: () => void }) => React.ReactNode)
+  /**
+   * The target button of the Popover.
+   * When a function the following arguments are passed:
+   * ({ toggle: Function -> Void, getRef: Function -> Ref, isShown: Bool })
+   */
   children:
     | ((props: {
         toggle: () => void
         getRef: (ref: React.RefObject<HTMLElement>) => void
-        isShown: NonNullable<PopoverProps['isShown']>
+        isShown: boolean
       }) => React.ReactNode)
     | React.ReactNode
+  /**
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -1874,28 +1874,88 @@ export declare const Pill: BoxComponent<PillOwnProps, 'strong'>
 export type PopoverStatelessProps = BoxProps<'div'>
 
 export interface PopoverProps {
+  /**
+   * The position the Popover is on. Smart positioning might override this.
+   */
   position?: PositionTypes
+  /**
```
park-jemin, line 1881: This is just a personal ask, but could we be a little more explicit about this? I feel like it's a little hard for me to immediately understand at first, so something this would be helpful:

```ts
  /**
   * Controls whether the Popover is shown or not.
   * - When `true`, the component is manually shown.
   * - When `false`, the component is manually hidden.
   * - When `undefined`, the component is uncontrolled and the isShown state 
   *    is handled internally (i.e., retains default behavior and shows on hover)
   */
  isShown?: boolean
```

index.d.ts
```diff
@@ -3005,7 +3065,10 @@ export interface TooltipProps {
    */
   showDelay?: number
   /**
-   * When true, manually show the Tooltip.
+   * Controls whether the Tooltip is shown or not.
```
park-jemin, line 3068: Same here

src/popover/src/Popover.js
```diff
@@ -334,7 +334,10 @@ Popover.propTypes = {
   ]),
 
   /**
-   * When true, the Popover is manually shown.
+   * Controls whether the Popover is shown or not.
```
park-jemin, line 337: Same here

src/tooltip/src/Tooltip.js
```diff
@@ -183,7 +183,10 @@ Tooltip.propTypes = {
   showDelay: PropTypes.number,
 
   /**
-   * When True, manually show the Tooltip.
+   * Controls whether the Tooltip is shown or not.
+   * - When `true`, the component is manually shown.
```
park-jemin, line 187: Same here (idk if I can point to all these with all this a single comment, sorry)

park-jemin (approved):

index.d.ts
brandongregoryscott: Fair enough! How does it sound now?

park-jemin: Much much better, thank you!!

PR #1506, opened 2022-08-03
Set Autocomplete input value based on selected item when allowOtherValues is true

### LINKED ISSUE #1496: Autocomplete: `allowOtherValues` not working?
Hi, this doesn't seem to be correct behavior: I'm using Autocomplete with property `allowOtherValues` enabled. When I select an item from the Autocomplete menu, the input value does not get updated.

Reproduction steps:
Navigate to https://evergreen.segment.com/components/autocomplete#default_example
Click on "Open in CodeSandbox".
Check that the Autocomplete component is working correctly.
Edit sample code to add property `allowOtherValues` to the Autocomplete element.
Refresh the "Browser" frame in CodeSandbox.
Select an item in Autocomplete and verify that it is not updating the value of the input element.

brandongregoryscott:
**Overview**

Resolves #1496

This is a small change to update the `inputValue` state to the selected item when `allowOtherValues` is set. This issue was only  noticeable in uncontrolled instances of the `Autocomplete`, i.e. one that does not set the `selectedItem` prop ([Default Example](https://evergreen.segment.com/components/autocomplete#default_example) on our doc site). The correct value (i.e. the selected item) was being set internally, but the input field was not reflecting it. 

I added some tests to cover this behavior as well. React testing library was updated to get rid of some warnings about the v18 `createRoot` API.

<table>
<tr>
	<td>Previous
	<td>Current
<tr>
	<td>Previous Behavior - Item is selected but not updated as input value</img>
	<td>Current Behavior - Item is selected and input value reflects selection</img>
</table>

**Documentation**


rutvij2811:
Hey @brandongregoryscott ,
I also thought of a similar solution to this issue but I seem to fail to build the package.
Did you face any similar issues? I am very new to open source contribution and could not find any FAQ of sorts to check if it is an existing issue.
Could you guide me or point out some existing resources if any? Thanks in advance.

brandongregoryscott:
@rutvij2811 Please open up a new issue if you're having trouble running the project. Any output from the build, and version numbers of the tools in your environment would help. Thanks!

package.json
```diff
@@ -107,7 +107,7 @@
     "@storybook/react": "^6.0.21",
     "@storybook/theming": "^6.0.21",
     "@testing-library/jest-dom": "^5.11.6",
-    "@testing-library/react": "^11.2.2",
+    "@testing-library/react": "13.3.0",
```
dlasky, line 110: generally speaking would prefer if this was a separate PR for rollback purposes, but i wont block on it.

dlasky (approved):

package.json
brandongregoryscott: ah, true! I figured it was lower risk since it was a dev dependency/testing package, but I can revert and follow it up. I think the test output is just a bit spammy without the upgrade, but it should still pass

dlasky: not a blocker, just something to consider for future work!

PR #1517, opened 2022-08-19
fix: fixed nextjs/react hydration error when using portals

oxala:
**Overview**
Currently, while using nextjs, if the page will have an evergreen component which uses portal that is visible (isShown={true}), during initial render (page load) react will complain about failed hydration:

```
Unhandled Runtime Error
Error: Hydration failed because the initial UI does not match what was rendered on the server.
See more info here: https://nextjs.org/docs/messages/react-hydration-error
```

This PR is my attempt to change the smallest amount code and following these resources:
- https://github.com/vercel/next.js/blob/canary/examples/with-portals/components/ClientOnlyPortal.js
- https://github.com/tajo/react-portal/issues/217

_Side note:_
This doesn't actually render portal content on the server-side. Portal initiation happens on the component mounts on client.
As far as I googled, there is no solution to have portals properly rendering on server-side. Some modules showcasing the idea, like [this module](https://github.com/MichalZalecki/react-portal-universal). But they are not production ready, as they can [leak portals across concurrent requests](https://github.com/MichalZalecki/react-portal-universal/issues/5).

**Documentation**


brandongregoryscott:
Hey @oxala, thanks for taking a look into this. I pulled this down and tried running it with your example repo, but I'm still seeing the same hydration error:

image

Is that expected, or should it be rendering properly on the client?

brandongregoryscott (approved):

brandongregoryscott:
Pulled it down w/ your example repo and it seems to solve the issue. Going to do a little more smoke testing with this change before merging it in, but it makes sense based on those discussions linked. Thanks for fixing this up!

Issue #1522, opened 2022-09-23
RadioGroup doesn't support name prop


richardscarrott:
`RadioGroup` generates a [random `name` prop](https://github.com/segmentio/evergreen/blob/master/src/radio/src/RadioGroup.js#L25), making it unusual for JS handling and unusable for HTML form submissions.

i.e. A native form action (server side) has no way of knowing what the name will be.

https://codesandbox.io/s/empty-https-ej6nsc

brandongregoryscott:
Seems like an oversight, but should be simple enough to fix!

brandongregoryscott:
This should be resolved in https://github.com/segmentio/evergreen/releases/tag/v6.10.4

PR #1524, opened 2022-09-28
Adds HTML name attribute support for RadioGroup

fs-nathan:
**Overview**
Currently the `RadioGroup` component automatically generates a string prefix 'RadioGroup' for the HTML `name` attribute on input type radio. This PR is about to add a name prop support for the component so the developer can set the attribute on their own and have more controls of form radio buttons on both client side and server side if they need it.

A recent issue is #1522 

***Before this change***
image

***After this change***
image

**Documentation**

**Testing**

Added new unit test and run test on local

image

### DIFF
--- a/src/radio/__tests__/RadioGroup.test.js
+++ b/src/radio/__tests__/RadioGroup.test.js
@@ -56,5 +56,12 @@ describe('Radio Group', () => {
     })
   })
 
+  it('should render the radio buttons with auto-generated HTML name attributes when name property is empty', () => {
+    render(<RadioGroup options={options} label="Permissions" />)
+    screen.getAllByRole('radio').forEach(element => {
+      expect(element.getAttribute('name')).toContain('RadioGroup')
+    })
+  })
+
   // untested props: defaultValue, size
 })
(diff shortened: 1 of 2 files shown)


fs-nathan:
#1522

fs-nathan:
Hi @brandongregoryscott Would you mind reviewing this PR? Thank you!

src/radio/__tests__/RadioGroup.test.js
```diff
@@ -49,5 +49,12 @@ describe('Radio Group', () => {
     expect(onChangeMock).toBeCalledTimes(1)
   })
 
+  it('should render the radio buttons with HTML name attributes specified', () => {
```
brandongregoryscott, line 52: Can we add a test to make sure each radio input still has a name if not passed in? (I realize there's no tests for this currently, but always good to backfill coverage when we're updating stuff anyway!)

fs-nathan: Sure. I will add a test for the case when the prop is not passed in

fs-nathan: I added new test for the case `name` property is empty @brandongregoryscott

brandongregoryscott (approved):

PR #1526, opened 2022-10-04
Omit children, items, and onChange from Combobox.autocompleteProps

### LINKED ISSUE #1525: Combobox `autocompleteProps` should be `Omit<AutocompleteProps, 'children' | 'items' | 'onChange'>`
When providing `autocompleteProps` to the `Combobox`, it requires `items`, `onChange`, and `children`, which don't really make sense - the `Combobox` component manages these internally. It looks like the `Combobox` spreads these props after the `items`/`onChange` from its _own_ props, so it won't "break" anything by allowing these thru, but it's not really the intended API of the component.

See this CodeSandbox for an example of passing `items` and `onChange` through `autocompleteProps`: https://codesandbox.io/s/combobox-autocompleteprops-override-dc4rle?file=/index.js

brandongregoryscott:
Fixes #1525

`items`, `children` and `onChange` are required properties in `AutocompleteProps`, but they are passed in as top-level props for the `Combobox`. You'd need to cast any object you pass in if you wanted to override specific Autocomplete behavior. This should be closer to the intended usage of the component and ease the need to cast any props you're sending thru it, if you really need to.

**Overview**

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1054,7 +1054,7 @@ export interface ComboboxOwnProps {
   /**
    * Properties forwarded to the autocomplete component. Use with caution.
    */
-  autocompleteProps?: AutocompleteProps
+  autocompleteProps?: Omit<AutocompleteProps, 'children' | 'items' | 'onChange'>
   /**
    * When true, open the autocomplete on focus.
    */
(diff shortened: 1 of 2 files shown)


neztEx (approved):
Looks good!

PR #1528, opened 2022-10-07
Fix: allow OptionsList to not focus on the search bar

chenxzhang:
Original Jira ticket: https://segment.atlassian.net/browse/PROT-4257
Loom: https://www.loom.com/share/79fd27a5ac2a415c93aac856c50c427a

**Overview**
There's a bug in Protocols tracking plan, where the focus of the input changes to the focus of the filter search bar. This happens because the tracking plan editor pages refreshes every few seconds and the focus is autofocused to the search bar  component. This PR adds a prop `shouldAutoFocus` to `OptionsList` so that we can disable the autofocus as needed. The change will not affect existing behaviors. 

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2388,7 +2388,8 @@ export interface SelectMenuProps extends Omit<PopoverProps, 'position' | 'conten
    */
   hasFilter?: boolean
   /**
-   * @default true. When true, auto focuses on the search/filter bar. 
+   * When true, auto focuses on the search/filter bar. 
+   * @default true
    */
    shouldAutoFocus?: boolean
   /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -2388,6 +2388,10 @@ export interface SelectMenuProps extends Omit<PopoverProps, 'position' | 'conten
    */
   hasFilter?: boolean
   /**
+   * @default true. When true, auto focuses on the search/filter bar. 
+   */
+   shouldAutoFocus?: boolean
+  /**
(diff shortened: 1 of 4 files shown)


src/select-menu/src/SelectMenu.js
```diff
@@ -186,7 +188,12 @@ SelectMenu.propTypes = {
   /**
    * The height of the items in the select menu list
    */
-  itemHeight: PropTypes.number
+  itemHeight: PropTypes.number,
+
+  /**
+   * When true, menu auto focuses on the search/filter bar.
+   */
+  shouldAutoFocus: PropTypes.bool
```
brandongregoryscott, line 196: Can we add this to the `SelectMenuProps` interface in `index.d.ts` as well? (Listing `@default true` might be helpful as well)

index.d.ts
```diff
@@ -2387,6 +2387,10 @@ export interface SelectMenuProps extends Omit<PopoverProps, 'position' | 'conten
    * When true, show the filter.
    */
   hasFilter?: boolean
+  /**
+   * @default true. When true, auto focuses on the search/filter bar.
```
brandongregoryscott, line 2391: Can we move the description on a separate line from the `@default` JSDoc attribute? I'm thinking that shows up weird in your IDE:

image

vs the expected:

image

chenxzhang: i see, thanks for the note!

brandongregoryscott (approved):

PR #1529, opened 2022-10-08
feat: insert test spinner and coverage command

DevJoaoLopes:
**Overview**
- creation of unit test for component Spinner
- command insertion for generating reports ``` yarn coverage ```

- *When executing the yarn coverage command, a folder called coverage containing the file below is generated in the project root.*
- Command: ``` yarn coverage ```
- Path where the generated file is located:

image

- Open file **index.html**:
image

**Documentation**

### DIFF
--- a/src/spinner/__tests__/Spinner.test.js
+++ b/src/spinner/__tests__/Spinner.test.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import Spinner from '../src/Spinner'
+
+describe('Spinner', () => {
+  jest.useFakeTimers()
+  jest.spyOn(global, 'setTimeout')
+  it('should not crash when rendering', () => {
+    expect(() => {
+      render(<Spinner />)
+    }).not.toThrowError()
+  })
+
+  it('should render', () => {
+    expect(() => {
+      render(<Spinner />)
+    }).not.toBeNull()
+  })
+
+  it('should render Spinner with delay time', () => {
+    const spinnerDelay = render(<Spinner delay={300} />)
+
+    jest.advanceTimersByTime(300)
+
+    expect(setTimeout).toHaveBeenCalledTimes(1)
+    expect(setTimeout).toHaveBeenLastCalledWith(expect.any(Function), 300)
+    expect(spinnerDelay).not.toBeNull()
+  })
+})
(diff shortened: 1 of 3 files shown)


src/spinner/__tests__/Spinner.test.js
```diff
@@ -0,0 +1,29 @@
...
+  it('should not crash when rendering', () => {
+    expect(() => {
+      render(<Spinner />)
+    }).not.toThrowError()
+  })
+
+  it('should render', () => {
+    expect(() => {
+      render(<Spinner />)
+    }).not.toBeNull()
```
brandongregoryscott, line 17: Commenting out the `if` condition in this block `if (!isVisible) {  return null }` in the component does not fail this test. Similar to the below test, I think you'd want something like this to test for its presence:

```tsx
render(<Spinner data-testid="Spinner" />)

expect(screen.getByTestId('Spinner')).not.toBeNull()
```

src/spinner/__tests__/Spinner.test.js
```diff
@@ -0,0 +1,29 @@
...
+  })
+
+  it('should render Spinner with delay time', () => {
+    const spinnerDelay = render(<Spinner delay={300} />)
+
+    jest.advanceTimersByTime(300)
+
+    expect(setTimeout).toHaveBeenCalledTimes(1)
+    expect(setTimeout).toHaveBeenLastCalledWith(expect.any(Function), 300)
+    expect(spinnerDelay).not.toBeNull()
```
brandongregoryscott, line 27: This assertion will always pass - `render` returns an object wrapping a bunch of React Testing Library functions, whether the component renders anything or not. 

I think you'd probably want to assert that the outermost `<Box />` is found in the DOM after time has progressed. 

```tsx
  it('should render after delay time', () => {
    render(<Spinner delay={300} data-testid="Spinner" />)

    act(() => {
      jest.advanceTimersByTime(300)
    })

    expect(setTimeout).toHaveBeenCalledTimes(1)
    expect(setTimeout).toHaveBeenLastCalledWith(expect.any(Function), 300)
    expect(screen.getByTestId('Spinner')).not.toBeNull()
  })
```

Additionally, the _opposite_ of this test case might be useful - the `Spinner` should not render before the delay:
```tsx
  it('should not render before delay time', () => {
    const delay = 300
    render(<Spinner delay={delay} data-testid="Spinner" />)

    act(() => {
      jest.advanceTimersByTime(delay / 2)
    })

    expect(() => screen.getByTestId('Spinner')).toThrow()
  })
```

The `act` is required otherwise RTL will return console warnings for state updates that might not tracked the same way the browser would see them.

PR #1531, opened 2022-10-11
Add props for FileUploader i18n support

### LINKED ISSUE #1459: Extract localizable strings so they can be customized
As an example, in the file picker control, the following strings are hard coded:

```
<Text color={disabled ? colors.gray500 : colors.blue400}>Browse </Text>
const orDragCopy = `or drag ${maxFiles === 1 ? 'a file' : 'files'} here`
```

It would be good if both were extracted to props:
- Browse would be a single string prop.
- The data for `onDragCopy` would be a function that would take the count and return a string.

The result would look something like this:
```
<FileUploader
  browseText={t('browse')}
  dragText={(count) => t('fileCount', { count })}
/>
```

The same should apply to other strings.

brandongregoryscott:
**Overview**

Resolves #1459 

This PR introduces two new props to the `FileUploader` to allow the inner/CTA text as well as the validation message on drag to be overridden. The default behavior/copy should remain the same and these props are optional. 

(From Storybook)

Custom browseOrDragText function that returns a string
---

FileUploader - browseOrDragText

Custom dragMaxFilesMessage function
---

FileUploader - dragMaxFilesMessage

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1435,6 +1435,10 @@ export interface FileUploaderOwnProps extends FormFieldOwnProps {
    */
   acceptedMimeTypes?: MimeType[]
   /**
+   * Function to return a string or component for the 'Browse or drag' text
+   */
+  browseOrDragText?: (maxFiles: number) => React.ReactNode
+  /**
    * When true, displays a disabled state where drops don't fire and the native browser picker doesn't open
    */
   disabled?: boolean
--- a/index.d.ts
+++ b/index.d.ts
@@ -2397,6 +2397,11 @@ export interface SelectMenuProps extends Omit<PopoverProps, 'position' | 'conten
    */
   hasFilter?: boolean
   /**
+   * When true, auto focuses on the search/filter bar. 
+   * @default true
+   */
+   shouldAutoFocus?: boolean
+  /**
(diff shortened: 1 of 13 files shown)


index.d.ts
```diff
@@ -1434,10 +1434,19 @@ export interface FileUploaderOwnProps extends FormFieldOwnProps {
    * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
    */
   acceptedMimeTypes?: MimeType[]
+  /**
+   * Function to return a string or component for the 'Browse or drag' text
+   */
+  browseOrDragText?: (maxFiles: number) => React.ReactNode
```
neztEx, line 1440: What is the difference between this call and the `dragMaxFilesMessage`? In this call we pass in the `maxFiles` and then return a string or component vs the other it only returns a `string` when the `maxFiles` size has been hit on drag.

brandongregoryscott: `browseOrDragText` is the text inside the dropzone, while `dragMaxFilesMessage` is the validation message that appears underneath the dropzone on drag:

image

neztEx (approved):
Looks good!

PR #1535, opened 2022-10-18
Add shouldAutoFocus prop to Overlay, SideSheet, and Dialog components

### LINKED ISSUE #1404: Ability to disable autofocus logic in Overlay component
On more than one occasion, the autofocus behavior in `<Overlay />` has been more obstructive than helpful when trying to properly set focus in a `Dialog` or `SideSheet` which are built on-top of it.

This is the autofocus logic as it stands today:
https://github.com/segmentio/evergreen/blob/0ba8a87b70c029febac59b57277336a8f45fe5cd/src/overlay/src/Overlay.js#L157-L168

[CodeSandbox example of the current default behavior (focuses the `Submit` button despite other input/interactive elements appearing prior)](https://codesandbox.io/s/evergreen-issues-1404-default-sidesheet-focus-behavior-y75uhn?file=/src/App.tsx)

For a specific example, if you have a SideSheet that has multiple input elements that should naturally take focus on open and there's an element that has a `tabIndex` manually set for accessibility purposes - it will prioritize focusing that element first, which probably doesn't make sense. [CodeSandbox](https://codesandbox.io/s/evergreen-issues-1404-default-sidesheet-focus-behavior-w-tabindex-ke57ox?file=/src/App.tsx)

One other thing to note is that the current behavior may not even be correct without a manually set `tabIndex`. For example, if you have a `SideSheet` that is rendering our `Combobox` component, it will focus the dropdown button instead of the input element that comes before it. [CodeSandbox](https://codesandbox.io/s/evergreen-issues-1404-default-sidesheet-focus-behavior-w-combobox-2ndb0e?file=/src/App.tsx)

This forces consumers to either accept incorrect focusing behavior, or wire up some unfortunate workarounds such as an invisible `TextInput` that has `autoFocus={true}`. 

While a longer-term solution might be updating the autofocus priority to be a little smarter (or controlled by the consumer if desired), I think a good middle-ground is retaining the current behavior but adding a prop that can disable it entirely.

brandongregoryscott:
Resolves #1404 

**Overview**

This PR introduces a prop to `Overlay`, `SideSheet` and `Dialog` that allows a consumer to disable the default autofocus behavior that lives in `Overlay.js`. As noted in that task, there are certain cases where the behavior can be incorrect or undesirable but there's no obvious way to opt out of it. This PR addresses that - retaining the current behavior, but opening up a way to disable it if needed.

Behavior can be demoed in the respective Storybook areas:

https://deploy-preview-1535--evergreen-storybook.netlify.app/?path=/story/overlay--autofocus-disabled

https://deploy-preview-1535--evergreen-storybook.netlify.app/?path=/story/side-sheet--autofocus-disabled

(Scroll to bottom and click button 'Show Dialog with autofocus disabled')
https://deploy-preview-1535--evergreen-storybook.netlify.app/?path=/story/dialog--dialog

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2460,18 +2460,43 @@ export interface SelectMenuProps extends Omit<PopoverProps, 'position' | 'conten
 
 export declare const SelectMenu: React.FC<SelectMenuProps>
 
-export interface SideSheetProps {
+export interface SideSheetProps
+  extends Pick<
+    OverlayProps,
+    'isShown' | 'onBeforeClose' | 'preventBodyScrolling' | 'shouldAutoFocus' | 'shouldCloseOnEscapePress'
+  > {
   children: React.ReactNode | (() => React.ReactNode)
-  isShown?: boolean
+
+  /**
+   * Function that will be called when the exit transition is complete.
+   */
   onCloseComplete?: () => void
+
+  /**
+   * Function that will be called when the enter transition is complete.
+   */
   onOpenComplete?: () => void
-  onBeforeClose?: () => boolean
+
+  /**
+   * Boolean indicating if clicking the overlay should close the overlay.
+   * @default true
+   */
   shouldCloseOnOverlayClick?: boolean
-  shouldCloseOnEscapePress?: boolean
+
+  /**
+   * Width of the SideSheet.
+   */
   width?: string | number
+
+  /**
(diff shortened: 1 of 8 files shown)


neztEx (approved):
Looks good!

PR #1540, opened 2022-10-27
Changed image storybook url. Previous link doesn't work.

quentinneal:
**Overview**

I noticed that the image component src URL doesn't work and does not display on the storybook image section. I replaced the URL.

Files affected:

index.stories.js at evergreen\src\image\stories

**Documentation**


brandongregoryscott (approved):
Good catch!

PR #1541, opened 2022-10-27
feat(FilePicker): props for i18n support

KillerCodeMonkey:
**Overview**

- Added props to translate all hard coded labels and texts in the FilePicker Component
  - inputText - a function called with File[] and returning the value string
  - browseOrReplaceText - a function called with the file count (number) and returning a string for the button tex
- Added tests + fixed and activated the "onBlur" test
- Fixed styling (missing marginLeft -1 for button)
- removed the unnecessary dependency to `react-test-renderer` only to make snapshots -> use testing library for it

***no file***
Screenshot_20221027_134837

***one file***
Screenshot_20221027_134858

***multiple files***
Screenshot_20221027_134914

**Documentation**

### DIFF
--- a/src/file-picker/__tests__/FilePicker.test.js
+++ b/src/file-picker/__tests__/FilePicker.test.js
@@ -144,46 +144,69 @@ describe('<FilePicker />', () => {
     expect(input.placeholder).toEqual(placeholder)
   })
 
-  it('sets browseText', async () => {
-    const text = 'select file'
-    const { container } = render(<FilePicker browseText={text} />)
+  it('sets browseOrReplaceText', async () => {
+    const browseText = 'Select'
+    const replaceText = 'Replace'
+    const replaceMultipleText = 'Replace all'
 
-    expect(await getButton(container)).toHaveTextContent(text)
-  })
+    const { container } = render(
+      <FilePicker
+        multiple
+        browseOrReplaceText={fileCount => {
+          if (!fileCount) return browseText
+          if (fileCount === 1) return replaceText
+          return replaceMultipleText
+        }}
+      />
+    )
 
-  it('sets replaceText', async () => {
-    const text = 'replace files'
-    const { container } = render(<FilePicker replaceText={text} />)
+    const button = await getButton(container)
+
+    expect(button).toHaveTextContent(browseText)
 
     const fileInput = await getFileInput(container)
     userEvent.upload(fileInput, [new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' })])
 
-    expect(await getButton(container)).toHaveTextContent(text)
-  })
+    expect(button).toHaveTextContent(replaceText)
 
-  it('sets replaceMultipleText', async () => {
-    const text = 'replace files'
-    const { container } = render(<FilePicker multiple replaceMultipleText={text} />)
-
-    const fileInput = await getFileInput(container)
     userEvent.upload(fileInput, [
       new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' }),
       new File(['(□_□¬)'], 'brucelee.png', { type: 'image/png' })
     ])
 
-    expect(await getButton(container)).toHaveTextContent(text)
+    expect(button).toHaveTextContent(replaceMultipleText)
   })
 
   it('sets multipleText', async () => {
-    const text = 'files selected'
-    const { container } = render(<FilePicker multiple multipleText={text} />)
+    const noFiles = ':('
+    const oneFile = 'File: chucknorris.png'
+    const multipleFiles = 'Files: 2'
+
+    const { container } = render(
+      <FilePicker
+        multiple
+        inputText={files => {
+          if (!files.length) return noFiles
+          if (files.length === 1) return `File: ${files[0].name}`
+          return `Files: ${files.length}`
+        }}
+      />
+    )
+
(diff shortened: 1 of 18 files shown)


src/file-picker/src/FilePicker.js
```diff
@@ -184,7 +185,27 @@ FilePicker.propTypes = {
    * Class name passed to the FilePicker.
    * Only use this if you know what you are doing.
    */
-  className: PropTypes.string
+  className: PropTypes.string,
+
+  /**
+   * Text of the button when no file is selected
+   */
+  browseText: PropTypes.string,
```
brandongregoryscott, line 193: I think these props can probably be reduced based on the parts of the component we're rendering copy for - the button and the input value. 

How about an API that passes the current file count (for the button) and file list (for the input value)? This would require the consumer to set two props instead of four:

```tsx
interface FilePickerProps {
  // ...
  
  /**
   * Function that returns the call-to-action button text for selecting files
   */
  browseOrReplaceText?: (fileCount: number) => string;

  /**
   * Function that returns the text in the input field describing what file or files have been added
   */
  inputText?: (files: File[]) => string;

  // ...
}
```

The current logic in the component can be extracted into utility functions that are used when props aren't provided:

```tsx
const getBrowseOrReplaceText = (fileCount) => {
  const action = fileCount === 0 ? 'Select' : 'Replace'
  const fileLabel = fileCount > 1 ? 'files' : 'file'

  return `${action} ${fileLabel}`
}

const getInputValue = (files) => {
  if (files.length === 1) {
    return files[0].name
  }

  if (files.length > 1) {
    return `${files.length} files`
  }

  return ''
}
```

This also gives the consumer the flexibility to provide a custom input label based on the file name(s), if they want - which currently isn't possible.

brandongregoryscott (changes requested):

brandongregoryscott:
>one question: i changed the docs, but when i follow the steps of the readme (even after i did yarn build) the example is not recognizing my new properties.

The doc site references a published version of the package, so it won't pull in local changes without a little bit of work. I think we're probably OK to leave this change out of the doc site since we don't really call out internationalization for any other component docs.

src/file-picker/src/FilePicker.js
```diff
@@ -184,7 +188,20 @@ FilePicker.propTypes = {
...
+  /**
+   * Function that returns the call-to-action button text for selecting files.
+   * Gets the file count as number as parameter.
+   */
+  browseOrReplaceText: PropTypes.func,
+
+  /**
+   * Function that returns the text in the input field,
+   * Describing what file or files have been added.
+   * Gets the selected files as File[] as parameter.
```
brandongregoryscott, line 202: One last nitpick - can we update this to use the `@type` JSDoc annotation vs. describing it?

`* @type {(files: File[]) => string}`

src/file-picker/src/FilePicker.js
```diff
@@ -184,7 +188,20 @@ FilePicker.propTypes = {
    * Class name passed to the FilePicker.
    * Only use this if you know what you are doing.
    */
-  className: PropTypes.string
+  className: PropTypes.string,
+
+  /**
+   * Function that returns the call-to-action button text for selecting files.
+   * Gets the file count as number as parameter.
```
brandongregoryscott, line 195:
```suggestion
   `* @type {(fileCount: number) => string}`
```

src/file-picker/src/FilePicker.js
```diff
@@ -184,7 +188,20 @@ FilePicker.propTypes = {
    * Class name passed to the FilePicker.
    * Only use this if you know what you are doing.
    */
-  className: PropTypes.string
+  className: PropTypes.string,
+
+  /**
+   * Function that returns the call-to-action button text for selecting files.
```
brandongregoryscott, line 194: Additionally, can we add these to the `index.d.ts` file in the root of the repo under `FilePickerProps`? (or `FilePickerOwnProps` if that's where the inner props are defined)

src/file-picker/__tests__/FilePicker.test.js
```diff
@@ -99,32 +124,89 @@ describe('<FilePicker />', () => {
...
+
+    userEvent.upload(fileInput, [
+      new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' }),
+      new File(['(□_□¬)'], 'brucelee.png', { type: 'image/png' })
+    ])
+
+    expect(button).toHaveTextContent(replaceMultipleText)
+  })
+
+  it('sets multipleText', async () => {
```
brandongregoryscott, line 180: This should be `inputText`

brandongregoryscott (changes requested):

src/file-picker/src/FilePicker.js
KillerCodeMonkey: sorry did not notice that there are ts typings

src/file-picker/src/FilePicker.js
KillerCodeMonkey: sry overlooked the ts typings

src/file-picker/__tests__/FilePicker.test.js
KillerCodeMonkey: good catch

brandongregoryscott (approved):

brandongregoryscott:
Thanks for the contribution, @KillerCodeMonkey!

PR #1561, opened 2022-12-01
Tooltip: User defined id for aria-describedBy

### LINKED ISSUE #842: NextJS Warning: Prop `aria-describedby` did not match.
**Describe the bug**
When using next.js and adding Tooltip component, it displays this error: 

> Warning: Prop `aria-describedby` did not match. Server: "evergreen-tooltip-192" Client: "evergreen-tooltip-1"

**To Reproduce**
1. Open new NextJS project
2. Install evergreen-ui
3. Use `<Tooltip />` component.

**Evergreen Version**
4.27.4

**Environment (please complete the following information):**
 - OS: windows
 - Browser chrome

harish-prakash:
**Overview**
Using the tooltip in SSR (server-side rendering) applications cause an ugly error due to render comparison.
```
Warning: Prop `aria-describedby` did not match. Server: "evergreen-tooltip-40" Client: "evergreen-tooltip-2"
```

* Resolves #842

**Documentation**
- [X] Updated Typescript types and/or component PropTypes
- [X] Added / modified component docs

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3114,7 +3114,7 @@ export interface TooltipProps {
   /**
    * The id of the tooltip.
    */
-  id: string,
+  id?: string
   /**
    * The position the Tooltip is on.
    */
--- a/index.d.ts
+++ b/index.d.ts
@@ -3112,6 +3112,10 @@ export interface TooltipProps {
    */
   appearance?: TooltipAppearance
   /**
+   * The id of the tooltip.
+   */
+  id: string,
+  /**
    * The position the Tooltip is on.
    */
   position?: PositionTypes
--- a/src/tooltip/src/Tooltip.js
+++ b/src/tooltip/src/Tooltip.js
@@ -22,7 +22,7 @@ const Tooltip = memo(function Tooltip(props) {
     statelessProps = emptyProps
   } = props
 
-  const id = useId('evergreen-tooltip')
+  const id = useId('evergreen-tooltip', props.id)
   const [isShown, setIsShown] = useState(propIsShown || false)
   const [isShownByTarget, setIsShownByTarget] = useState(false)
   const closeTimer = useRef(undefined)
@@ -152,6 +152,11 @@ Tooltip.propTypes = {
    * The appearance of the tooltip.
    */
   appearance: PropTypes.oneOf(['default', 'card']),
+
+  /**
+   * The id of the tooltip.
+   */
+  id: PropTypes.string,
 
   /**
    * The position the Popover is on.


src/tooltip/src/Tooltip.js
```diff
@@ -22,7 +22,7 @@ const Tooltip = memo(function Tooltip(props) {
     statelessProps = emptyProps
   } = props
 
-  const id = useId('evergreen-tooltip')
+  const id = useId('evergreen-tooltip', props.id)
```
brandongregoryscott, line 25: I pulled this down and it seems to solve the warning in the original issue (when providing an explicit id). However, I think this is somewhat of a workaround - ideally, the `useId` hook should be reworked to return stable ids between the client <> server so an explicit id isn't required to suppress this warning during SSR.

I'm cool w/ adding in the ability to set an explicit id for the Tooltip either way, but I think we'll need to add it to `TooltipProps` in the `index.d.ts` file for TypeScript users:

https://github.com/segmentio/evergreen/blob/0c36c459c6964303d9877fc67b2a0c918c011170/index.d.ts#L3109-L3142

brandongregoryscott (changes requested):

src/tooltip/src/Tooltip.js
brandongregoryscott: `index.d.ts` _"should"_ be generated, but we're still on JS 🥲

>Without a unique reference for this component between the server and client, how would you create a stable id?

I'm not entirely sure! I'm not super familiar with SSR myself, but I did some quick searching and it looks like `useId` is available as a hook directly from React in v18 which solves this: https://github.com/reactwg/react-18/discussions/111

Mantine implements a [stable id hook](https://github.com/mantinedev/mantine/blob/master/src/mantine-hooks/src/use-id/use-id.ts) by setting the state in an effect: 

```tsx
import React, { useState } from 'react';
import { useIsomorphicEffect } from '../use-isomorphic-effect/use-isomorphic-effect';


const randomId = () => `mantine-${Math.random().toString(36).slice(2, 11)}`;


const useReactId: () => string | undefined =
  (React as any)['useId'.toString()] || (() => undefined);


function useClientId() {
  const [uuid, setUuid] = useState('');


  useIsomorphicEffect(() => {
    setUuid(randomId());
  }, []);


  return uuid;
}


function getReactId() {
  const id = useReactId();
  return id ? `mantine-${id.replace(/:/g, '')}` : '';
}


export function useId(staticId?: string) {
  return typeof staticId === 'string' ? staticId : getReactId() || useClientId();
}
```

index.d.ts
```diff
@@ -3111,6 +3111,10 @@ export interface TooltipProps {
    * The appearance of the Tooltip.
    */
   appearance?: TooltipAppearance
+  /**
+   * The id of the tooltip.
+   */
+  id: string,
```
brandongregoryscott, line 3117: `id` should be optional
```suggestion
  id?: string
```

brandongregoryscott (approved):

harish-prakash:
Thank you for handling this 🙇‍♂️

PR #1575, opened 2022-12-11
added 'tab' accessibility to modal with focus lock to prevent from fo…

simona1245:
as of the documentation and thorough internet searching about "tab focusing" it is clear that creating tab focus from scratch is not applicable. (see stackoverflow)

Library to add: react-focus-trap

### DIFF
--- a/docs/components/homepage/Resources.tsx
+++ b/docs/components/homepage/Resources.tsx
@@ -20,7 +20,7 @@ const Resources = () => {
           </Paragraph>
         </Pane>
         <Pane display="flex">
-          <ResourceCard title="Evergreen Figma Library" logo={<FigmaIcon />} url="https://www.figma.com/@segment" />
+          <ResourceCard title="Evergreen Figma Library" logo={<FigmaIcon />} url="https://www.figma.com/@twilio" />
           <ResourceCard title="Segment Brand Hub" logo={<SegmentIcon />} url="https://brand.segment.com/" />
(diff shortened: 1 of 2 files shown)


simona1245:
I think adding a new library and pushing the code made deploy preview fail. 

help?

brandongregoryscott:
Thanks for taking a look at this @simona1245! 

I pulled it down and gave it a run myself, and it does seem to trap focus as expected. One potential issue I noticed is that it no longer closes the dialog when you click outside of it. I'm not entirely sure whether we _should_ have this behavior, but it will be a breaking change if we keep it as-is. I'm going to dig into web accessibility standards a bit and see what I can find.

brandongregoryscott:
Hmm, after taking a look at what Paste does, they actually split out two variants of what looks like our `Dialog` component that are for different use cases:

https://paste.twilio.design/components/modal

This seems to be more or less our Dialog component, i.e. it closes on an outside click (but it is accessible in that it traps focus inside, handles escape to close, etc)

https://paste.twilio.design/components/alert-dialog

I think this is closer to what the proposed implementation in this PR is - same visual styling as their Modal component, but closing the AlertDialog has to be explicitly done with the Cancel/Confirmation action (escape does not close it either).

We're in somewhat of a middle state between these two variants. In the interest of backwards compatibility, can we figure out how to maintain the outside-click-to-close behavior?

PR #1577, opened 2022-12-14
Add codemods for ejecting/replacing classic/deprecatedDefaultTheme objects

brandongregoryscott:
**Overview**

This PR adds a few codemods for ejecting and replacing imports of the `classicTheme` or `deprecatedDefaultTheme` objects. I hadn't originally planned on adding such functionality, but it looks like we still have _lots_ (200+) of references to the `deprecatedDefaultTheme` in our own app, so this seemed like the easiest way to migrate for now.

Note: The theme objects have been exported as-is (basically a flattened JSON object from the latest version on Evergreen) - that means the selectors have not been updated for the new ui-box prop. These are being provided as-is, and should be migrated off of sooner rather than later. I was very surprised and disappointed to see so many references to the `deprecatedDefaultTheme` still lingering around in app 😢 I'm working with our designers to remap those colors to our modern color palette, so hopefully we can swap/remove it sooner rather than later.

**Documentation**


brandongregoryscott:
Found an edge case that might be worth handling (although a simple find + replace in VS Code would also fix this)
image

dlasky (approved):

PR #1587, opened 2023-01-05
Fix incorrect default border color for TextInput

### LINKED ISSUE #1581: FileUploader shadows TextInput's borders
When both `<FileUploader>` and `<TextInput>` present, TextInput's border would disappear.

Reproduction link: https://codesandbox.io/s/evergreen-ui-shadow-input-border-vveww1

brandongregoryscott:
**Overview**

Resolves #1581

For some reason we were applying a `border: 1px solid transparent` style in the base `Input` theme object, and then overriding it individually in the `default` appearance object with a `borderColor` prop. I don't think this was intentional, so I've made the default border color gray400 and explicitly set the transparent border in the `none` appearance object, and added a few tests. Aside from this, I believe there's a deeper issue in `ui-box` where shorthand props + individual props can conflict depending on when they are added to the stylesheet/used, but that's an issue to tackle another day.

FileUploader story
image

TextInput story (no change)
image

**Documentation**


src/text-input/__tests__/TextInput.test.js
```diff
@@ -51,60 +47,56 @@ describe('TextInput', () => {
...
+    render(makeTextInputFixture({ disabled: true }))
+    const input = screen.getByTestId('input')
     userEvent.type(input, 'Testing')
 
-    expect(() => getByDisplayValue('Testing')).toThrowError()
-    expect(getByDisplayValue('')).toEqual(input)
-  })
-})
-
-describe('TextInputField', () => {
```
brandongregoryscott, line 73: Most of these tests were already in `TextInputField.test.js`, but must not have been removed when the files were split up.

dlasky (approved):
lgtm!

PR #1590, opened 2023-01-09
Port over minimal TS configs, deps and port Pane

brandongregoryscott:
**Overview**

This was bit of a stretch goal of mine for V7 - bringing TypeScript into the project and kicking off the port! Instead of doing a massive one-pass refactor from JS -> TS like #1467, I pulled out the parts of the migration utils I wrote and reworked them to act on one component at a time, and (for the most part) be re-runnable without side effects if the component has already been ported.

In the interest of keeping this PR smaller and not littering the repo with temporary codemod scripts that aren't necessarily _battle-tested_ yet, I kept them out of this PR. I may break out another repo or just keep a branch alive that has the utilities to use when I'm porting over more components.

Summary of changes:
- Added a rollup plugin for bundling our UMD build to TS
- Added the babel TS preset for bundling our ESM build
- CommonJS build now uses tsc
- Switched to @typescript-eslint instead of babel-eslint for parsing
- Prettier was updated to support newer TS syntax (i.e. type import/exports)
- Generated types from tsc will be re-exported through the manually written index.d.ts until we can re-point our `package.json` to the generated `index.d.ts`
- PropTypes will stay in for now - we are parsing them for our doc site, so until we can parse prop interfaces instead, we have to maintain both 😞 

**Documentation**


brandongregoryscott:
Pulled this into app and there's definitely something jacked up with the Pane types - going to do some further investigation!

brandongregoryscott:
After updating the type cast of the memoized Pane and re-adding the `react-transition-group` types, I'm not seeing any more ts errors in app!

package.json
```diff
@@ -45,9 +47,9 @@
     "deploy-docs": "cd docs && yarn install && yarn deploy",
     "dev": "start-storybook -p 6006",
     "generate-icons": "./tools/generate-icons.js",
-    "lint": "eslint \"src/**/*.js\" --cache",
+    "lint": "eslint \"src/**/*.{js,ts,tsx}\" --cache",
     "postpublish": "branch=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD); if [[ $branch == \"master\" ]]; then yarn update-docs || yarn deploy-docs; else echo \"not on main branch, skipping docs deploy\"; fi;",
-    "prepublishOnly": "rm -rf esm commonjs umd && yarn build",
+    "prepublishOnly": "rm -rf esm commonjs umd types && yarn build",
     "release": "np",
     "size": "size-limit",
```
dlasky, line 54: may want to add a tsc typecheck command and add it as an initial step in the build

src/layers/src/Pane.tsx
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * Elevation of the Pane on click. Might get deprecated.
+   */
+  activeElevation?: Elevation
+}
+
+export type PaneProps<T extends React.ElementType = 'div'> = PolymorphicBoxProps<T, PaneOwnProps>
+
+// Temporary type alias to satisfy TS, may be extracted to a shared types directory later
+type Elevation = 0 | 1 | 2 | 3 | 4
```
dlasky, line 57: maybe useful here if we upgrade: https://catchts.com/range-numbers

src/layers/src/Pane.tsx
```diff
@@ -0,0 +1,162 @@
...
+  )
+
+  // @ts-expect-error TS(2322): Type '{ selectors: SelectorMap | undefined; style:... Remove this comment to see the full error message
+  return <Box ref={ref} className={className} {...themedProps} {...restProps} />
+}
+
+const Pane = memo(forwardRef(_Pane)) as <T extends React.ElementType = 'div'>(props: PaneProps<T>) => JSX.Element
+
+// @ts-expect-error TS(2339): Property 'propTypes' does not exist on type '<E ex... Remove this comment to see the full error message
+Pane.propTypes = {
```
dlasky, line 108: curious why we are leaving this?

dlasky (approved):

src/layers/src/Pane.tsx
brandongregoryscott: We actually parse the propTypes from the component src files for the `/:component/props` page on our doc site (i.e. https://evergreen.segment.com/components/file-uploader/props), so I'm keeping them in until more time can be dedicated to a solution that reads the ts definitions instead

I ran into this during my massive 1-pass-port and tried swapping out `react-docgen` with something like https://github.com/styleguidist/react-docgen-typescript, but I think that came with its own issues/upgrades that were needed first.

PR #1598, opened 2023-01-22
Cancel animation frame on component unmount to avoid memory leaks

Arooba-git:
**Overview**
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in renowned SPA applications. 
While running the tool and analyzing the code of Evergreen, we saw that it does a very good job of ensuring that all async operations are cancelled when the component unmounts. However, as per Memlab execution results, we found that the cancellation of requestAnimationFrames is forgotten at certain places (TableVirtualBody being one of them) and was causing the memory to leak (screenshots below).

**Screenshots**
[before]
Screen Shot 2023-01-22 at 6 44 55 PM

Screen Shot 2023-01-22 at 6 45 22 PM
<br /><br />

Hence we added the fix by cancelling the frameRequest and you can see the heap size and # of leaks reducing noticeably:

Screen Shot 2023-01-22 at 6 50 05 PM

Screen Shot 2023-01-22 at 6 50 19 PM

You can test this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering the maximum # of use cases.
Following is a sample of the scenario file we used: 
[evergreen-scenario-memlab.txt](https://github.com/segmentio/evergreen/files/10474132/evergreen-scenario-memlab.txt)

Note that some other reported leaks originate from the Storybook library, hence ignored.

### DIFF
--- a/src/table/src/TableVirtualBody.js
+++ b/src/table/src/TableVirtualBody.js
@@ -1,4 +1,4 @@
-import React, { memo, useState, useEffect } from 'react'
+import React, { memo, useState, useEffect, useRef } from 'react'
 import VirtualList from '@segment/react-tiny-virtual-list'
 import debounce from 'lodash.debounce'
 import PropTypes from 'prop-types'
@@ -29,6 +29,7 @@ const TableVirtualBody = memo(function TableVirtualBody(props) {
   const [paneRef, setPaneRef] = useState()
   const [isIntegerHeight, setIsIntegerHeight] = useState(false)
   const [calculatedHeight, setCalculatedHeight] = useState(0)
+  const requestId = useRef(null)
 
   const updateOnResize = () => {
     autoHeights = []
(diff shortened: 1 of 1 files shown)


src/table/src/TableVirtualBody.js
```diff
@@ -52,7 +53,7 @@ const TableVirtualBody = memo(function TableVirtualBody(props) {
     }
 
     // When height is still 0 (or paneRef is not valid) try recursively until success.
-    requestAnimationFrame(() => {
+    requestId.current = requestAnimationFrame(() => {
```
brandongregoryscott, line 56: Good catch! Should we also capture the animation frame request id for the call on line 136? I'm thinking that also has potential for a memory leak.

Arooba-git: Yes, as you might know, frameRequests should always be cancelled once the component is destroyed/unmounted. The scenario file I executed might have missed covering the call of the enclosing function of line 136, hence was not pointed out in Memlab leaks result, but best to clear that too :)

brandongregoryscott (approved):

PR #1604, opened 2023-02-01
Fix vertical text alignment in Alert/Toast

kaelig:
**Overview**

Fixes an issue where the removable variant of the Alert component introduces vertical spacing below the text, making it look misaligned.

Screenshot 2023-02-01 at 3 59 42 PM

### Before (detailed box model)
Screenshot 2023-02-01 at 3 52 15 PM

### After (detailed box model)
Screenshot 2023-02-01 at 3 52 48 PM

**Documentation**

### DIFF
--- a/src/alert/src/Alert.js
+++ b/src/alert/src/Alert.js
@@ -66,7 +66,7 @@ const Alert = memo(
           )}
         </Pane>
         {isRemoveable && (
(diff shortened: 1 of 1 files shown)


brandongregoryscott (approved):
Good catch! I think we should probably use flex to center align all of these elements, but since it was already positioned this way, this is fine for now. Thanks!

Issue #1602, opened 2023-01-31
Combobox button border is darker than the input next to it


brandongregoryscott:
Split from #1508 

![Screen Shot 2022-08-05 at 7 41 24 PM](https://user-images.githubusercontent.com/28040841/183101143-0f1237c5-ea6f-47d5-b13f-e7934a198bd7.png)
![Screen Shot 2022-08-05 at 7 42 06 PM](https://user-images.githubusercontent.com/28040841/183101160-745ea27a-1b65-498f-8feb-a17aeed9891e.png) ![Screen Shot 2022-08-05 at 7 42 25 PM](https://user-images.githubusercontent.com/28040841/183101166-0ef3e89c-8a57-4000-9a03-f359b2758395.png)

JeevantheDev:
Bug is fixed, PR has been raised. https://github.com/segmentio/evergreen/pull/1614

PR #1614, opened 2023-02-05
Bug 1602 combobox button

JeevantheDev:
**Overview**
- Created a props named as intent (**type as string**) for **Combobox** component and set default value 'input'.
- For Button component styles set another if condition inside **borderColorForIntent** function which will check for intent is equal to 'input'. If satisfied it will add the borderColor same as TextArea border.
- For Button disabled same condition updated.
#1602 Reference Bug
evergreen-combobox

**Documentation**

### DIFF
--- a/src/buttons/src/IconButton.js
+++ b/src/buttons/src/IconButton.js
@@ -23,6 +23,7 @@ const IconButton = memo(
         ref={ref}
         paddingLeft={0}
         paddingRight={0}
+        intent={intent}
         flex="none"
         height={height}
         width={height}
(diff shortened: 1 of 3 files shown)


brandongregoryscott:
Hey @JeevantheDev, thanks for the PR. I don't think adding an `input` intent really fits in with the current API (other intents are `danger`, `warning`, `success` and `none`, which is the default). Additionally, this does not make the make the border color match the TextInput when focused (still not sure if it should, since our standard Buttons don't, but this component is a form element composed of both.. 🤔 )
 
image
image

I wish I had a better suggestion for how to properly implement this right now, but it requires partially overriding the selectors that are setup in the `Button` component which isn't very obvious, even to me. Since the Combobox has been like this for so long, I don't think this is a huge priority to fix right now.

PR #1615, opened 2023-02-05
Cleanup doc site warnings, add types, etc.

brandongregoryscott:
**Overview**

The doc site has had a huge number of warnings for unexpected `any` and non-null assertions for a while. Most of these were on the `name` property for the generic `Item` in our information architecture object, which is always present. I've updated the type and removed a bunch of assertions/casts. There are some other type changes in #1611 that should almost completely resolve the build warnings.

Nothing should functionally change here, just some code cleanup/logic consolidation.

**Documentation**


docs/components/IconSearch.tsx
```diff
@@ -39,7 +36,7 @@ const Item: React.FC<{ name: string }> = ({ name }) => {
         >
           {/* eslint-disable-next-line */}
           {/*  @ts-ignore */}
-          {React.createElement(evergreen[name] as any, {
+          {React.createElement(evergreen[name] as React.ReactElement, {
```
MaybeRex, line 39: do you need the @ts-ignore now that you fixed the any?

MaybeRex (approved):

docs/components/IconSearch.tsx
brandongregoryscott: Good catch - probably not!

PR #1616, opened 2023-02-09
fix: `toaster` types

filippofilip95:
Hello ! :) 

**Overview**
This PR fixes the Typescript type of the `toaster` instance in the following way:
1. Retype `title` argument in `toaster`. The `title` was typed as only a `string` but instead, it can support `ReactNode`. Type `ReactNode` is for `title` also specified in [interface Toast](https://github.com/segmentio/evergreen/blob/b5efd40b271b386eb30e2cc053fa0cb7d733b11e/index.d.ts#L2794).
2. Add function `remove` to the `toaster`. Function `remove` is exposed in the toaster however it doesn't exist in types. Function `remove` is also already used in the storybook [here](https://github.com/segmentio/evergreen/blob/da12d7d7619da0a8797d137c2105957dfedcafd9/src/toaster/stories/index.stories.js#L124).

Documentation doesn't need a change since it's already considering that type `title` is a `node`.

Screenshot 2023-02-09 at 10 40 03

**Documentation**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2843,6 +2843,10 @@ export const toaster: {
    * Returns all visible Toasts.
    */
   getToasts: () => Toast[]
+  /**
+   * Removes toast with specific id
+   */
+  remove: (id: string) => void
 }
 
 export interface OverlayProps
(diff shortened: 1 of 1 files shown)


brandongregoryscott (approved):

brandongregoryscott:
Thanks for the contribution, @filippofilip95!

PR #1620, opened 2023-03-08
Fix SelectMenu width & height docs

ibrahimpg:
**Overview**
The width/height portion of the SelectMenu docs was simply a duplicate of the example above it about removing the title/filter. some small uniformity changes as well.

Before/after
fix

**Documentation**


brandongregoryscott (approved):
Nice catch @ibrahimpg!

Issue #1625, opened 2023-03-23
Property `selected` should include `PropTypes.number` to allow integer values


ADTC:
https://github.com/segmentio/evergreen/blob/e13b9b4f2609ed3ecb5f7dc218dfe4fe1a1da048/src/select-menu/src/SelectedPropType.js#L8

Please allow `PropTypes.number` here. Sometimes the value is numeric -  quite common when display label is a string, but internal value is an int.

Although the application works, this throws error in the console.

```
Warning: Failed prop type: Invalid prop `selected` supplied to `SelectMenu`, expected one of type [string].
SelectMenu@webpack-internal:///./node_modules/evergreen-ui/esm/select-menu/src/SelectMenu.js:39:15
▽ React 6
        printWarning
        error
        checkPropTypes
        validatePropTypes
        jsxWithValidation
        jsxWithValidationDynamic
```

Right now I'm overriding the propTypes to fix the error:

```js
SelectMenu.propTypes = {
  ...SelectMenu.propTypes,
  selected: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.arrayOf(PropTypes.string),
    PropTypes.number,
  ]),
};
```

PR #1627, opened 2023-04-06
Added type number to proptypes

AbhiJain2196:
**Overview**
Instead of making changes in the SelectMenu.js file, we can directly change the SelectedPropType.js file and add PropTypes.number to PropTypes.oneOfType() function

**Documentation**

### DIFF
--- a/src/select-menu/src/SelectedPropType.js
+++ b/src/select-menu/src/SelectedPropType.js
@@ -2,9 +2,10 @@ import PropTypes from 'prop-types'
 
 /**
  * Selected can either be a string (single values)
+ * or can be a number ( single value)
  * or an array of string (multiple values)
  * NOTE: multiple values are not supported atm
  */
-const SelectedPropType = PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)])
+const SelectedPropType = PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string), PropTypes.number])
 
 export default SelectedPropType


ADTC:
Thanks. Fixes #1625

ADTC:
It's not linking it for auto-closure in the sidebar. I think you'll have to edit the top description and [add the `Fixes` phrase there yourself](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

Thanks for the work. Appreciate it.

ADTC:
Shouldn't we also support an array of numbers? I'm not sure why it allows an array of strings though, is that for multi-select?

AbhiJain2196:
In case of single or multi select, I think this code block will cover both the scenarios
`PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.string, PropTypes.number]))`

ADTC:
> In case of single or multi select, I think this code block will cover both the scenarios
> ```js
> PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.string, PropTypes.number]))
> ````

That won't work probably for single select, as it's not an array. This would: 
```js
PropTypes.oneOfType([
  PropTypes.string,
  PropTypes.number,
  PropTypes.arrayOf(PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.number
  ]))
])
```
But that may not be proper, since it allows mixed arrays (both strings and numbers in the same array).

This won't allow mixed arrays. The arrays have to be pure:
```js
PropTypes.oneOfType([
  PropTypes.string,
  PropTypes.number,
  PropTypes.arrayOf(PropTypes.string),
  PropTypes.arrayOf(PropTypes.number)
])
```
I would go with this instead.

brandongregoryscott:
Hey folks - I think we'll need to update the TS definitions with whatever we use for PropTypes, as well. If you're using TS, you currently can't pass a number to `selected` either: https://codesandbox.io/s/pr-1627-selectmenu-number-typing-27seme?file=/src/App.tsx

PR #1674, opened 2024-02-13
fix windows build

maxrchung:
**Overview**
I would like to run this project on Windows but encountered some issues trying to run some of the scripts. In this PR, I adjusted a few things to make `yarn build` and `yarn dev` work:
- Change single quotes to escaped double quotes. Based on this [concurrently issue](https://github.com/open-cli-tools/concurrently/issues/167) it seems Windows treats quotes differently.
- Use [`cross-env`](https://www.npmjs.com/package/cross-env) to set environment variables in script.
- Use [`rimraf`](https://www.npmjs.com/package/rimraf) to remove files.

I verified on my Windows and Mac machines that `yarn build` and `yarn dev` were successful after the changes. One thing separately I also noticed was that Node 18 did not seem to be supported, and I had to downgrade to Node 16 in order to run the project.

N/A

**Documentation**


package.json
maxrchung, line 1: Note: I didn't update all of the scripts. I left `update-docs` as is.

